PR #46, opened 2021-06-15
Support @Nested test classes.

### LINKED ISSUE #45: `@Nested` test classes are not supported.
First of all: Big kudos for a project.

I was eager to try this out on https://github.com/neo4j-contrib/cypher-dsl. We do have a lot of `@Nested` classes, however.

Changing the `App2Test` class in your example code like this

```java
package org.graalvm.buildtools.example;

import org.junit.jupiter.api.Nested;
import org.junit.jupiter.api.Test;
import static org.junit.jupiter.api.Assertions.assertNotNull;

class App2Test {
    @Test void appHasAGreeting() {
        App classUnderTest = new App();
        assertNotNull(classUnderTest.getGreeting(), "app should have a greeting");
    }

    @Nested
    class ANestedTest {

        @Test
        void apphasNestedGreetingToo() {
            App classUnderTest = new App();
            assertNotNull(classUnderTest.getGreeting(), "app should have a greeting");
        }
    }
}
```

will fail in Maven `mvn -Pnative test` with 

```
junit-platform-native] Running in 'test listener' mode.
Fatal error:java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: Class org.graalvm.buildtools.example.App2Test$ANestedTest cannot be instantiated reflectively . It does not have a nullary constructor.
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
```

The same result will happen with the Gradle project, even when the agent is used (with a workaround specified in this issue https://github.com/graalvm/native-build-tools/issues/44) 

I have attached the generated native image files.
[test.zip](https://github.com/graalvm/native-build-tools/files/6649728/test.zip)

Class is registered for reflection, but something still is odd.

gradinac:
Closes #45

### DIFF
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
@@ -156,20 +156,23 @@ public final class JUnitPlatformFeature implements Feature {
                 .map(Optional::get)
                 .filter(ClassSource.class::isInstance)
                 .map(ClassSource.class::cast)
-                .forEach((ClassSource classSource) -> {
-                    Class<?> clazz = classSource.getJavaClass();
-                    if (debug) {
-                        System.out.println("[Debug] Found test class: " + clazz);
-                    }
-                    RuntimeReflection.registerForReflectiveInstantiation(clazz);
-                    for (Field field : clazz.getDeclaredFields()) {
-                        RuntimeReflection.register(field);
-                    }
-                    for (Method method : clazz.getDeclaredMethods()) {
-                        RuntimeReflection.register(method);
-                    }
-                });
+                .map(ClassSource::getJavaClass)
+                .forEach(this::registerTestClassForReflection);
 
         return testPlan;
+    }
+
+    private void registerTestClassForReflection(Class<?> clazz) {
+        if (debug) {
+            System.out.println("[Debug] Registering test class for reflection: " + clazz.getName());
+        }
+        RuntimeReflection.register(clazz.getDeclaredConstructors());
+
+        for (Field field : clazz.getDeclaredFields()) {
+            RuntimeReflection.register(field);
+        }
+        for (Method method : clazz.getDeclaredMethods()) {
+            RuntimeReflection.register(method);
+        }
     }
 }
--- a/common/junit-platform-native/src/test/java/com/example/project/CalculatorTests.java
+++ b/common/junit-platform-native/src/test/java/com/example/project/CalculatorTests.java
@@ -43,11 +43,17 @@ package com.example.project;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
 
 class CalculatorTests {
+
+    CalculatorTests(TestInfo info) {
+        System.out.println("Running test: " + info.getDisplayName());
+    }
 
     @Test
     @DisplayName("1 + 1 = 2")
@@ -56,7 +62,7 @@ class CalculatorTests {
         assertEquals(2, calculator.add(1, 1), "1 + 1 should equal 2");
     }
 
-   @Test
+    @Test
     @DisplayName("1 + 2 = 3")
     void addsTwoNumbers2() {
         Calculator calculator = new Calculator();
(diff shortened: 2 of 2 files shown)


common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -156,20 +157,31 @@ private TestPlan registerTestPlan(Launcher launcher, List<Path> classpath) {
...
         return testPlan;
     }
+
+    private void registerTestClassForReflection(ClassSource classSource) {
+        Class<?> clazz = classSource.getJavaClass();
+        if (debug) {
+            System.out.println("[Debug] Found test class: " + clazz);
+        }
+
+        Class<?> declaringClass = clazz.getDeclaringClass();
```
gradinac, line 171: We could also do `clazz.getEnclosingClass` here - the only difference between `getDeclaringClass` and `getEnclosingClass` is that `getEnclosingClass` also works for anonymous classes. However, I don't think that's relevant in this context, but I may be wrong

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -156,20 +157,31 @@ private TestPlan registerTestPlan(Launcher launcher, List<Path> classpath) {
...
+        if (declaringClass == null || Modifier.isStatic(clazz.getModifiers())) {
+            /* A top level or a static nested class must have a no-arg constructor. */
+            RuntimeReflection.registerForReflectiveInstantiation(clazz);
+        } else {
+            /* A non-static nested class cannot have a no-arg constructor as it needs a reference to an enclosing parent's object. */
+            RuntimeReflection.register(clazz.getDeclaredConstructors());
```
gradinac, line 177: We don't need to explicitly register the declaring class for reflection here as it will be one of the processed test classes.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
sbrannen: JUnit 4 and JUnit Jupiter do not support anonymous classes as test classes, but that doesn't mean that some other `TestEngine` in the world doesn't.

I think it might be unlikely that there's such a `TestEngine`, but I can't 100% rule it out.

So maybe going on the safe side (using `getEnclosingClass()`) is the better choice, especially if it has no adverse side effects.

gradinac: Ah, this is a good point - I'll change it to `getEnclosingClass`

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -156,20 +157,31 @@ private TestPlan registerTestPlan(Launcher launcher, List<Path> classpath) {
...
+
+    private void registerTestClassForReflection(ClassSource classSource) {
+        Class<?> clazz = classSource.getJavaClass();
+        if (debug) {
+            System.out.println("[Debug] Found test class: " + clazz);
+        }
+
+        Class<?> declaringClass = clazz.getEnclosingClass();
+        if (declaringClass == null || Modifier.isStatic(clazz.getModifiers())) {
+            /* A top level or a static nested class must have a no-arg constructor. */
```
sbrannen, line 173: That's not entirely true.

It is up to the `TestEngine` implementation to determine how the test class is instantiated.

For example, JUnit Jupiter only requires that a test class have a single constructor. Whether or not that single constructor accepts arguments is up to the test author. See [Dependency Injection for Constructors and Methods](https://junit.org/junit5/docs/current/user-guide/#writing-tests-dependency-injection) for details.

Other `TestEngine` implementations may support multiple constructors as they see fit.

Long story, short: we should register all declared constructors for reflection.

gradinac: Ah, that's a good point, it makes sense that JUnit Jupiter would support dependency injection and that other engines could place their own requirements on the test classes. I agree, we should register all declared constructors for reflection.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -156,20 +156,23 @@ private TestPlan registerTestPlan(Launcher launcher, List<Path> classpath) {
                 .map(Optional::get)
                 .filter(ClassSource.class::isInstance)
                 .map(ClassSource.class::cast)
```
sbrannen, line 158:
```suggestion
                .map(ClassSource.class::cast)
                .map(ClassSource::getJavaClass)
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -156,20 +156,23 @@ private TestPlan registerTestPlan(Launcher launcher, List<Path> classpath) {
...
+                .forEach(this::registerTestClassForReflection);
 
         return testPlan;
     }
+
+    private void registerTestClassForReflection(ClassSource classSource) {
+        Class<?> clazz = classSource.getJavaClass();
+        if (debug) {
+            System.out.println("[Debug] Found test class: " + clazz);
+        }
```
sbrannen, line 168:
```suggestion
    private void registerTestClassForReflection(Class<?> clazz) {
        if (debug) {
            System.out.println("[Debug] Registering test class for reflection: " + clazz.getName());
        }
```

sbrannen (changes requested):
I've requested some minor polishing changes.

Please also introduce constructors for `CalculatorTests` and `NestedNestedTest` that accept a `TestInfo` argument just to ensure that the `Feature` does not break this _feature_ of JUnit Jupiter.

sbrannen (approved):

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
sbrannen: Out of curiosity, why did you decide against `clazz.getName()`?

gradinac:
I agree completely both with the changes and with adding a constructor with the `TestInfo` parameter - it's good to have as many features tested as possible.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
gradinac: Oh, snap! I was sure it was `clazz.getName()` before, my bad! Updating it now - it makes more sense to have `class.getName()`

Issue #53, opened 2021-06-15
Support for Compatibility with JUnit 4?


dzou:
Many thanks for this great project.

I wanted to ask about `junit-platform-native`: Is there future plans to make this compatible with JUnit4 running tests via `junit-vintage-engine`?

sbrannen:
The initial goal is support for the JUnit Platform in general, with specific support for JUnit Jupiter.

However, some preliminary work has been done to provide built-in support for JUnit Vintage:

https://github.com/graalvm/native-build-tools/blob/acf14743521e01260155825ab2e2e5cc55905ea5/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java#L94-L95

If you are experiencing specific issues, please provide a sample application demonstrating what you are encountering or provide stack traces, debug output, etc. that could help the team assess the problem.

As I mentioned above, the current "support" for JUnit Vintage is preliminary which means you likely have to add some manual configuration for yourself -- for example, reflection config for JUnit Vintage internals. Though, in theory it should be possible to get JUnit 4 tests running via the JUnit Vintage `TestEngine`.

dzou:
I see, thanks for the clarification. I'll give running the JUnit4 tests a try and followup in this thread what configurations I used if I'm able to get it working.

dzou:
I played around with it and was able to get JUnit 4 working with the following additional classes marked for build-time initialization:

```
--initialize-at-build-time=org.junit.runner.Description,org.junit.vintage.engine.descriptor.RunnerTestDescriptor,org.junit.runners.JUnit4,org.junit.runners.BlockJUnit4ClassRunner
```

Link to example: https://github.com/dzou/junit4-experiment

I'll try this out on a bigger set of tests and see if it still works.

sbrannen:
> I played around with it and was able to get JUnit 4 working with the following additional classes marked for build-time initialization:

Great!

Thanks for sharing, and please do keep us posted on your progress.

See also: https://github.com/graalvm/native-build-tools/pull/59#discussion_r658849595

gradinac:
Hey @dzou! We've added really basic support for the JUnit Vintage engine as part of #59 - could you try out the latest changes on master and see if they work for your samples?

While it's at a really basic level right now, it should hopefully grow over time and encompass more JUnit features out of the box

dzou:
Great! I'll give it a try and will followup.

dzou:
[...]

PR #59, opened 2021-06-18
Introduce and migrate to a plugin configuration mechanism.

### LINKED ISSUE #50: `@ParameterizedTest` with `@EnumSource` are not supported.
A parameterized tests with an `@EnumSource` like this

```java
package org.graalvm.buildtools.example;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.EnumSource;

class App2Test {

	enum Things {
		A, B, C
	}

	@ParameterizedTest
	@EnumSource(Things.class)
	void functionNameSanityCheck(Things thing) {

		Assertions.assertNotNull(thing);
	}

}
```

fails with

```
Failures (1):
  JUnit Jupiter:App2Test:functionNameSanityCheck(Things)
    MethodSource [className = 'org.graalvm.buildtools.example.App2Test', methodName = 'functionNameSanityCheck', methodParameterTypes = 'org.graalvm.buildtools.example.App2Test$Things']
    => org.junit.platform.commons.JUnitException: Failed to find a no-argument constructor for ArgumentsProvider [org.junit.jupiter.params.provider.EnumArgumentsProvider]. Please ensure that a no-argument constructor exists and that the class is either a top-level class or a static nested class
       org.junit.jupiter.params.ParameterizedTestExtension.instantiateArgumentsProvider(ParameterizedTestExtension.java:110)
       java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
       java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
       java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
       java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
       [...]
       Suppressed: org.junit.platform.commons.PreconditionViolationException: Configuration error: You must configure at least one set of arguments for this @ParameterizedTest
         org.junit.platform.commons.util.Preconditions.condition(Preconditions.java:281)
         org.junit.jupiter.params.ParameterizedTestExtension.lambda$provideTestTemplateInvocationContexts$6(ParameterizedTestExtension.java:94)
         java.util.stream.AbstractPipeline.close(AbstractPipeline.java:323)
         java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271)
         [...]
     Caused by: java.lang.NoSuchMethodException: org.junit.jupiter.params.provider.EnumArgumentsProvider.<init>()
       java.lang.Class.getConstructor0(DynamicHub.java:3349)
       java.lang.Class.getDeclaredConstructor(DynamicHub.java:2553)
       org.junit.platform.commons.util.ReflectionUtils.newInstance(ReflectionUtils.java:488)
       org.junit.jupiter.params.ParameterizedTestExtension.instantiateArgumentsProvider(ParameterizedTestExtension.java:102)
       [...]
```

Applies to 0.9.0 and 0.9.1-SNAPSHOT, tested with GraalVM 21.1.0 JDK 11 in the Maven examples project.

### LINKED ISSUE #51: `@ParameterizedTest` with `@MethodSource` are not supported.
A parameterized tests with an `@MethodSource` like this

```java
import java.util.stream.Stream;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

class App2Test {

	enum Things {
		A, B, C
	}

	@ParameterizedTest
	@MethodSource("aMethod")
	void functionNameSanityCheck(Things thing) {

		Assertions.assertNotNull(thing);
	}

	private static Stream<Arguments> aMethod() {
		return Stream.of(
			Arguments.of(Things.A),
			Arguments.of(Things.B),
			Arguments.of(Things.C)
		);
	}
}
```

fails with

```
Failures (1):
  JUnit Jupiter:App2Test:functionNameSanityCheck(Things)
    MethodSource [className = 'org.graalvm.buildtools.example.App2Test', methodName = 'functionNameSanityCheck', methodParameterTypes = 'org.graalvm.buildtools.example.App2Test$Things']
    => org.junit.platform.commons.JUnitException: Failed to find a no-argument constructor for ArgumentsProvider [org.junit.jupiter.params.provider.MethodArgumentsProvider]. Please ensure that a no-argument constructor exists and that the class is either a top-level class or a static nested class
       org.junit.jupiter.params.ParameterizedTestExtension.instantiateArgumentsProvider(ParameterizedTestExtension.java:110)
       java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
       java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
       java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
       java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
       [...]
       Suppressed: org.junit.platform.commons.PreconditionViolationException: Configuration error: You must configure at least one set of arguments for this @ParameterizedTest
         org.junit.platform.commons.util.Preconditions.condition(Preconditions.java:281)
         org.junit.jupiter.params.ParameterizedTestExtension.lambda$provideTestTemplateInvocationContexts$6(ParameterizedTestExtension.java:94)
         java.util.stream.AbstractPipeline.close(AbstractPipeline.java:323)
         java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271)
         [...]
     Caused by: java.lang.NoSuchMethodException: org.junit.jupiter.params.provider.MethodArgumentsProvider.<init>()
       java.lang.Class.getConstructor0(DynamicHub.java:3349)
       java.lang.Class.getDeclaredConstructor(DynamicHub.java:2553)
       org.junit.platform.commons.util.ReflectionUtils.newInstance(ReflectionUtils.java:488)
       org.junit.jupiter.params.ParameterizedTestExtension.instantiateArgumentsProvider(ParameterizedTestExtension.java:102)
       [...]
```

Applies to 0.9.0 and 0.9.1-SNAPSHOT, tested with GraalVM 21.1.0 JDK 11 in the Maven examples project.

### LINKED ISSUE #54: Configure types referenced via `@ExtendWith`, `@ArgumentsSource`, etc. for reflection
## Overview

As discussed in https://github.com/graalvm/native-build-tools/issues/50#issuecomment-862186948, the `JUnitPlatformFeature` currently has hard-coded reflection config for JUnit Jupiter's `CsvArgumentsProvider`, but as we see from #50 and #51 the `JUnitPlatformFeature` would at the very least need to configure *all* `junit-jupiter-params` `ArgumentProvider` implementations for reflective instantiation.

That is possible, but it would not help with `ArgumentProvider` implementations from users or third party libraries that are loaded by JUnit Jupiter via `@ArgumentSource` declarations (such as the `@ArgumentsSource(MethodArgumentsProvider.class)` declaration on `@MethodSource`).

## Details

The broader topic that needs to be addressed is how to automatically register test-related types for reflection.

Within JUnit Jupiter, the following annotations are used to refer to types or methods that JUnit accesses via reflection: `@ExtendWith`, `@ArgumentsSource`, `@TestMethodOrder`, `@DisplayNameGeneration`, `@IndicativeSentencesGeneration`, `@ConvertWith`, `@AggregateWith`, `@EnabledIf`, `@DisabledIf`, `@MethodSource`.

These fall into two categories.

#### Type References

`@ExtendWith`, `@ArgumentsSource`, `@TestMethodOrder`, `@DisplayNameGeneration`, `@IndicativeSentencesGeneration`, `@ConvertWith`, and `@AggregateWith` are used to reference a type that will be instantiated via reflection.

#### Method References

`@MethodSource`, `@EnabledIf`, and `@DisabledIf` are used to reference methods that will be invoked via reflection.

If the method that is referenced is local to the test class -- for example, `@MethodSource("localMethod")` -- there should not be any need for additional reflection configuration, since all methods of the test class should have already been registered for reflection. However, there may potentially be an issue if the referenced method is declared in a super type.

On the contrary, if one of these three annotations is used to reference a `static` method in a different class (i.e., not within the test class itself) -- for example, `@MethodSource("org.example.MyTestUtils#externalMethod")` -- the declared methods in the external class will need to be opened up for reflection.

## Deliverables

- [ ] Determine how to automatically configure types and methods referenced via `@ExtendWith`, `@ArgumentsSource`, `@TestMethodOrder`, `@DisplayNameGeneration`, `@IndicativeSentencesGeneration`, `@ConvertWith`, `@AggregateWith`, `@EnabledIf`, `@DisabledIf`, `@MethodSource` for reflection.
- [ ] If appropriate, remove all manual reflection configuration from the `JUnitPlatformFeature` for types that will be picked up automatically by changes made in conjunction with this issue.

gradinac:
This PR introduces a plugin configuration mechanism to separate plugin configuration from the JUnit feature. The core of this mechanism is the `PluginConfigProvider` service. The service currently implements 2 callbacks:
 - onLoad() - triggered when the `JUnitPlatformFeature` is loaded
 - onTestClassRegistered() - triggered when a test class is registered for inclusion in the image.

The provider receives a `NativeImageConfiguration` object that allows registering classes for reflection and build time initialization, removing the dependency to Native Image classes.

Currently, three such providers are implemented:
 - JUnit Platform
 - JUnit Jupiter
 - JUnit Vintage

All plugin-specific configuration has been moved from the feature to their respective provider.

Closes #54
Closes #51 
Closes #50

### DIFF
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/UniqueIdTrackingTestExecutionListener.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/UniqueIdTrackingTestExecutionListener.java
@@ -58,7 +58,7 @@ import java.util.List;
  * that were executed and generates a file (currently hard coded to
  * {@code ./build/test_ids.txt} with Gradle or {@code ./target/test_ids.txt} with Maven)
  * that contains test IDs which can be passed to custom launcher to select exactly those test classes.
- *
+ * <p>
  * This file should be replaced with org.junit.platform.launcher.listeners.UniqueIdTrackingListener,
  * once junit-platform-launcher version 1.8 gets released.
  *
@@ -72,7 +72,16 @@ public class UniqueIdTrackingTestExecutionListener implements TestExecutionListe
     private final List<String> uniqueIds = new ArrayList<>();
 
     @Override
+    public void executionSkipped(TestIdentifier testIdentifier, String reason) {
+        addTest(testIdentifier);
+    }
+
+    @Override
     public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
+        addTest(testIdentifier);
+    }
+
+    private void addTest(TestIdentifier testIdentifier) {
         if (testIdentifier.isTest()) {
             this.uniqueIds.add(testIdentifier.getUniqueId());
         }
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2021 Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The Universal Permissive License (UPL), Version 1.0
@@ -38,48 +38,32 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package org.graalvm.junit.platform;
+package org.graalvm.junit.platform.config.core;
 
-import org.junit.platform.engine.TestExecutionResult;
-import org.junit.platform.launcher.TestExecutionListener;
-import org.junit.platform.launcher.TestIdentifier;
-import org.junit.platform.launcher.TestPlan;
-import org.junit.platform.reporting.legacy.LegacyReportingUtils;
+import org.graalvm.junit.platform.JUnitPlatformFeature;
 
-import java.io.PrintWriter;
+import java.lang.reflect.Executable;
+import java.lang.reflect.Field;
 
-@SuppressWarnings("unused")
-public class PrintTestExecutionListener implements TestExecutionListener {
+public interface NativeImageConfiguration {
 
-    TestPlan testPlan;
-    final PrintWriter out;
+    void registerForReflection(Class<?>... classes);
 
-    public PrintTestExecutionListener() {
-        out = new PrintWriter(System.out);
-    }
+    void registerForReflection(Executable... methods);
 
-    public PrintTestExecutionListener(PrintWriter out) {
-        this.out = out;
-    }
+    void registerForReflection(Field... fields);
 
-    @Override
-    public void testPlanExecutionStarted(TestPlan testPlan) {
-        this.testPlan = testPlan;
-    }
-
-    @Override
-    public void executionSkipped(TestIdentifier testIdentifier, String reason) {
-        printTest(testIdentifier, "SKIPPED: " + reason);
-    }
-
-    @Override
-    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
-        printTest(testIdentifier, testExecutionResult.getStatus().name());
-    }
-
-    private void printTest(TestIdentifier testIdentifier, String status) {
-        if (testIdentifier.getParentId().isPresent() && !testIdentifier.isContainer()) {
-            out.println(LegacyReportingUtils.getClassName(testPlan, testIdentifier) + " > " + testIdentifier.getDisplayName() + " " + status + "\n");
+    default void registerAllClassMembersForReflection(Class<?>... classes) {
+        for (Class<?> clazz : classes) {
+            JUnitPlatformFeature.debug("[Native Image Configuration] Registering for reflection: %s", clazz.getName());
+            registerForReflection(clazz);
+            registerForReflection(clazz.getDeclaredConstructors());
+            registerForReflection(clazz.getDeclaredMethods());
+            registerForReflection(clazz.getDeclaredFields());
         }
     }
+
+    void initializeAtBuildTime(String... classNames);
+
+    void initializeAtBuildTime(Class<?>... classes);
 }
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/PluginConfigProvider.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/PluginConfigProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2021 Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * The Universal Permissive License (UPL), Version 1.0
@@ -38,48 +38,12 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package org.graalvm.junit.platform;
+package org.graalvm.junit.platform.config.core;
 
-import org.junit.platform.engine.TestExecutionResult;
-import org.junit.platform.launcher.TestExecutionListener;
-import org.junit.platform.launcher.TestIdentifier;
-import org.junit.platform.launcher.TestPlan;
-import org.junit.platform.reporting.legacy.LegacyReportingUtils;
+public interface PluginConfigProvider {
 
-import java.io.PrintWriter;
+    void onLoad(NativeImageConfiguration config);
 
-@SuppressWarnings("unused")
-public class PrintTestExecutionListener implements TestExecutionListener {
+    void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration registry);
 
-    TestPlan testPlan;
-    final PrintWriter out;
-
-    public PrintTestExecutionListener() {
-        out = new PrintWriter(System.out);
-    }
-
-    public PrintTestExecutionListener(PrintWriter out) {
-        this.out = out;
-    }
-
-    @Override
-    public void testPlanExecutionStarted(TestPlan testPlan) {
-        this.testPlan = testPlan;
-    }
-
-    @Override
-    public void executionSkipped(TestIdentifier testIdentifier, String reason) {
-        printTest(testIdentifier, "SKIPPED: " + reason);
-    }
-
-    @Override
-    public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
-        printTest(testIdentifier, testExecutionResult.getStatus().name());
-    }
-
-    private void printTest(TestIdentifier testIdentifier, String status) {
-        if (testIdentifier.getParentId().isPresent() && !testIdentifier.isContainer()) {
-            out.println(LegacyReportingUtils.getClassName(testPlan, testIdentifier) + " > " + testIdentifier.getDisplayName() + " " + status + "\n");
-        }
-    }
 }
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2021, 2021 Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
+ * The above copyright notice and either this complete permission notice or at a
+ * minimum a reference to the UPL must be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.junit.platform.config.jupiter;
+
+import org.graalvm.junit.platform.config.core.NativeImageConfiguration;
+import org.graalvm.junit.platform.config.core.PluginConfigProvider;
+import org.graalvm.junit.platform.config.util.AnnotationUtils;
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.IndicativeSentencesGeneration;
+import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.jupiter.api.condition.DisabledIf;
+import org.junit.jupiter.api.condition.EnabledIf;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.aggregator.AggregateWith;
+import org.junit.jupiter.params.converter.ConvertWith;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.platform.commons.support.AnnotationSupport;
+
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+import static org.graalvm.junit.platform.JUnitPlatformFeature.debug;
+
+public class JupiterConfigProvider implements PluginConfigProvider {
+
+    @Override
+    public void onLoad(NativeImageConfiguration config) {
+        String[] buildTimeInitializedClasses = new String[]{
+                "org.junit.jupiter.engine.config.EnumConfigurationParameterConverter",
+                "org.junit.jupiter.engine.descriptor.ClassTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.TestFactoryTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.JupiterTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.JupiterTestDescriptor$1",
+                "org.junit.jupiter.engine.descriptor.MethodBasedTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor",
+                "org.junit.jupiter.engine.descriptor.TestTemplateTestDescriptor",
+                "org.junit.jupiter.engine.execution.ConditionEvaluator",
+                "org.junit.jupiter.engine.execution.ExecutableInvoker"
+        };
+        for (String className : buildTimeInitializedClasses) {
+            config.initializeAtBuildTime(className);
+        }
+    }
+
+    @Override
+    public void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration registry) {
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, ArgumentsSource.class, ArgumentsSource::value);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, ExtendWith.class, ExtendWith::value);
+        AnnotationUtils.forEachAnnotatedMethod(testClass, EnumSource.class, (m, annotation) -> handleEnumSource(m, annotation, registry));
+        handleTestOrderer(testClass, registry);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, DisplayNameGeneration.class, DisplayNameGeneration::value);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, IndicativeSentencesGeneration.class, IndicativeSentencesGeneration::generator);
+        AnnotationUtils.forEachAnnotatedMethodParameter(testClass, ConvertWith.class, annotation -> registry.registerAllClassMembersForReflection(annotation.value()));
+        AnnotationUtils.forEachAnnotatedMethodParameter(testClass, AggregateWith.class, annotation -> registry.registerAllClassMembersForReflection(annotation.value()));
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, MethodSource.class, JupiterConfigProvider::handleMethodSource);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, EnabledIf.class, JupiterConfigProvider::handleEnabledIf);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, DisabledIf.class, JupiterConfigProvider::handleDisabledIf);
+    }
+
+    private static Class<?>[] handleMethodSource(MethodSource annotation) {
+        return handleMethodReference(annotation.value());
+    }
+
+    private static Class<?>[] handleEnabledIf(EnabledIf annotation) {
+        return handleMethodReference(annotation.value());
+    }
+
+    private static Class<?>[] handleDisabledIf(DisabledIf annotation) {
+        return handleMethodReference(annotation.value());
+    }
+
+    private static Class<?>[] handleMethodReference(String... methodNames) {
+        List<Class<?>> classList = new ArrayList<>();
+        for (String methodName : methodNames) {
+            String[] parts = methodName.split("#");
+            /*
+             * If the method used as an argument source resides in a different class than the test class, it must be specified
+             * by the fully qualified class name, followed by a # and the method name
+             */
+            debug("Found method reference: %s", methodName);
+            if (parts.length == 2) {
+                String className = parts[0];
+                debug("Processing method reference from another class: %s", className);
+                try {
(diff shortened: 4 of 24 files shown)


common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java
```diff
@@ -0,0 +1,76 @@
...
+                "org.junit.jupiter.engine.execution.ExecutableInvoker"
+        };
+        for (String className : buildTimeInitializedClasses) {
+            registry.initializeAtBuildTime(className);
+        }
+    }
+
+    @Override
+    public void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration registry) {
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, ArgumentsSource.class, source -> new Class[]{source.value()});
```
gradinac, line 72: For now, I've only migrated existing features to the new configuration mechanism. I suspect we may now support `EnumSource`s as well, however - I would mark that as "fixed" once we add tests for it in a separate PR.

lazar-mitrovic (approved):

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/UniqueIdTrackingTestExecutionListener.java
```diff
@@ -71,8 +71,17 @@
 
     private final List<String> uniqueIds = new ArrayList<>();
 
+    @Override
+    public void executionSkipped(TestIdentifier testIdentifier, String reason) {
+        addTest(testIdentifier);
```
gradinac, line 76: We also need to include tests that were skipped on the JVM in the native-image used for testing. If not, we would not be able to run tests disabled through `DisableIf` and `EnableIf` on the native-image, even if the condition that disabled them changed in the meantime.

sbrannen: That's an interesting topic. On the one hand, I can see that people might want to have each JUnit Jupiter `ExecutionCondition` reevaluated within the native image. On the other hand, people might want literally the exact same tests executed without having to have the conditions reevaluated (maybe b/c the conditions don't work the same within the native image).

Though perhaps it's best to include them by default now. We could later add an option to make that behavior configurable.

BUT... if we make this change here, I should probably make the same change in JUnit 5.8's `UniqueIdTrackingListener`.

gradinac: Both requirements seem valid - in this case, it may be best to go with the current approach and include an option later that allows the users to skip condition evaluation on the native-image. Note that this is a bit tricky: we are still evaluating the conditions, but we're only doing it for the tests whose condition allowed their execution on the JVM - this means that the set of tests we could execute on native before this change was either the same or a subset of the tests that had ran on the JVM

gradinac: I agree with changing the `UniqueIdTracingListener` - initially, it could also include both executed and skipped tests. If we move forward with adding an option to skip tests that were skipped on the JVM, we might need to update the output file format to also provide test execution information for each tests (by information, I mean either executed or skipped, not succeeded or failed)

sbrannen: > If we move forward with adding an option to skip tests that were skipped on the JVM, we might need to update the output file format to also provide test execution information for each tests (by information, I mean either executed or skipped, not succeeded or failed)

Why would we need to differentiate between those?

If a test method was disabled during the JVM run, it may or may not be disabled within the native image run. In either case, we would still select it by it's unique ID, and the JUnit Jupiter `ExecutionCondition` would evaluate it within the native image.

gradinac: Ah - that was related to the above proposal for an option that allows people to skip tests that were skipped on the JVM. That way, we could differentiate whether the test was skipped or executed from inside the feature. However, I agree with you - it probably isn't needed for now

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/PluginConfigProvider.java
```diff
@@ -38,12 +38,12 @@
...
-package com.example.project;
+package org.graalvm.junit.platform.config.core;
 
-public class Calculator {
+public interface PluginConfigProvider {
 
-    public int add(int a, int b) {
-        return a + b;
-    }
+    void onLoad(NativeImageConfiguration registry);
```
sbrannen, line 45:
```suggestion
    void onLoad(NativeImageConfiguration config);
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/PluginConfigProvider.java
```diff
@@ -38,12 +38,12 @@
...
+
+    void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration registry);
```
sbrannen, line 47:
```suggestion
    void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration config);
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/platform/PlatformConfigProvider.java
```diff
@@ -0,0 +1,77 @@
...
+                "org.junit.platform.commons.logging.LoggerFactory",
+                "org.junit.platform.engine.UniqueIdFormat",
+                "org.junit.platform.commons.util.ReflectionUtils"
+        };
+        for (String className : buildTimeInitializedClasses) {
+            registry.initializeAtBuildTime(className);
+        }
+
+        try {
+            registry.registerForReflection(Class.forName("org.junit.platform.commons.annotation.Testable").getMethods());
```
sbrannen, line 66: `@Testable` does not declare any attributes (i.e. methods).

So why is `getMethods()` used here?

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java
```diff
@@ -0,0 +1,153 @@
...
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, ExtendWith.class, ExtendWith::value);
+        AnnotationUtils.forEachAnnotatedMethod(testClass, EnumSource.class, (m, annotation) -> handleEnumSource(m, annotation, registry));
+        handleTestOrderer(testClass, registry);
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, DisplayNameGeneration.class, source -> new Class[]{source.value()});
+        AnnotationUtils.registerClassesFromAnnotationForReflection(testClass, registry, IndicativeSentencesGeneration.class, source -> new Class[]{source.generator()});
+        AnnotationUtils.forEachAnnotatedMethodParameter(testClass, ConvertWith.class, annotation -> registry.registerAllClassMembersForReflection(annotation.value()));
+        AnnotationUtils.forEachAnnotatedMethodParameter(testClass, AggregateWith.class, annotation -> registry.registerAllClassMembersForReflection(annotation.value()));
```
sbrannen, line 97: For things like `@ConvertWith` and `@AggregateWith`, why do we need to register all class members for reflection?

Don't we just need to register the class and its default constructor for reflection?

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/util/AnnotationUtils.java
```diff
@@ -0,0 +1,121 @@
...
+                    consumer.accept(annotation);
+                }
+            }
+        });
+    }
+
+    private static void forEachMethod(Class<?> clazz, Consumer<Method> consumer) {
+        Set<Method> methods = new HashSet<>();
+        methods.addAll(Arrays.asList(clazz.getDeclaredMethods()));
+        methods.addAll(Arrays.asList(clazz.getMethods()));
```
sbrannen, line 97: Please see related comment about `getDeclaredMethods()` and `getMethods()` in class hierarchies.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/vintage/VintageConfigProvider.java
```diff
@@ -0,0 +1,62 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.junit.platform.config.vintage;
+
+import org.graalvm.junit.platform.config.core.NativeImageConfiguration;
+import org.graalvm.junit.platform.config.core.PluginConfigProvider;
+
+public class VintageConfigProvider implements PluginConfigProvider {
```
sbrannen, line 46: It might be a good idea to include some of the feedback from @dzou (https://github.com/graalvm/native-build-tools/issues/53#issuecomment-867100493) in the Vintage support and add some JUnit 4 based tests.

Otherwise, that can be added in conjunction #53 to reduce the scope of this PR.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/vintage/VintageConfigProvider.java
```diff
@@ -0,0 +1,62 @@
...
+                "org.junit.vintage.engine.support.UniqueIdReader",
+                "org.junit.vintage.engine.support.UniqueIdStringifier"
+        };
+        for (String className : buildTimeInitializedClasses) {
+            registry.initializeAtBuildTime(className);
+        }
+    }
+
+    @Override
+    public void onTestClassRegistered(Class<?> testClass, NativeImageConfiguration registry) {
```
sbrannen, line 60: That looks suspiciously empty, but I suppose we'll learn more once we add dedicated JUnit 4 tests.

common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter/TestMethodOrderTests.java
```diff
@@ -0,0 +1,111 @@
...
+package org.graalvm.junit.jupiter;
+
+import org.graalvm.junit.util.Calculator;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.MethodOrderer;
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+
+public class TestMethodOrderTests {
```
sbrannen, line 50: Where are you verifying the expected method execution order?

common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter/TestMethodOrderTests.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+    public static class ArgumentOrderer {
```
sbrannen, line 53: Since these are static nested classes, they are treated like top-level classes in JUnit Jupiter.

Thus, I'm wondering if these test classes are actually executed in the build, since they do not follow the pattern `*Test(s)`.

In other words, have you verified that they are executed as part of the test suite in the build?

common/junit-platform-native/src/test/java/org/graalvm/junit/util/Calculator.java
```diff
@@ -0,0 +1,64 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.junit.util;
+
+import org.junit.jupiter.api.Assertions;
+
+public class Calculator {
```
sbrannen, line 45: For our tests that "test the tests", we don't really need a subject under test (SUT) like the `Calculator`.

It's of course good to use something like the `Calculator` in demos, but in our "testing the tests" tests the test execution is the SUT. So I'd probably just get rid of the `Calculator` for these types of tests.

If you want something to fail, just invoke `fail(<optional message>);`. If you want a test method to pass, just make it an empty test method unless there is something you can assert within the test method that guarantees its method execution order, display name, etc.

sbrannen (changes requested):
This PR is certainly improving a lot!

I've added several questions and suggestions.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/NativeImageConfigurationImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public void initializeAtBuildTime(String... classNames) {
+        for (String className : classNames) {
+            Class<?> clazz;
+            try {
+                clazz = Class.forName(className);
+                initializeAtBuildTime(clazz);
+            } catch (ClassNotFoundException e) {
+                JUnitPlatformFeature.debug(" [Native Image Configuration] Failed to register class for build time initialization: %s Reason: %s", className, e);
```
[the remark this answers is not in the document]
gradinac: It was intended - it seemed a bit nicer to separate this particular log with two spaces:
```
[Debug] Some other log
[Debug]  [Native Image Configuration] A message
```
However, I can remove it

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+public abstract class NativeImageConfiguration {
+
+    public abstract void registerForReflection(Class<?>... classes);
+
+    public abstract void registerForReflection(Executable... methods);
+
+    public abstract void registerForReflection(Field... fields);
+
+    public void registerAllClassMembersForReflection(Class<?>... classes) {
```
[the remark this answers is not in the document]
gradinac: It is a utility method that is extensively used in the PR - hence the `abstract` class instead of an interface. It may be cleaner to move it to a utility class, but we would end up with `UtilityClass.registerAllClassMembersForReflection(config, clazz)` instead of `config.registerAllClassmembersForReflection(clazz)`. 
Making this a default method and converting the class to an interface could work as well - I am up for both approaches

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/NativeImageConfigurationImpl.java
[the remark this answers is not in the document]
sbrannen: Yes, please remove it.

Let's not focus on special formatting here.

Ideally we should be using a logging framework that allows the user to specify formatting of log messages.

So let's just pretend we are already using a logging framework, since that will allow us to migrate to a logging framework more easily later.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+ * SOFTWARE.
+ */
+package org.graalvm.junit.platform.config.core;
+
+import org.graalvm.junit.platform.JUnitPlatformFeature;
+
+import java.lang.reflect.Executable;
+import java.lang.reflect.Field;
+
+public abstract class NativeImageConfiguration {
```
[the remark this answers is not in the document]
gradinac: See https://github.com/graalvm/native-build-tools/pull/59#discussion_r658866359

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/NativeImageConfigurationImpl.java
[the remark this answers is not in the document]
gradinac: I will remove the space.

I've also given a logging framework a thought while implementing this, but unfortunately, due to the context in which we are running, we can't safely do that: any code that runs in the feature implicitly initializes all used classes at image build time. If we were to use a logging framework, we would force the initialization of all used logging framework classes - this would affect user applications

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public abstract void registerForReflection(Field... fields);
+
+    public void registerAllClassMembersForReflection(Class<?>... classes) {
+        for (Class<?> clazz : classes) {
+            JUnitPlatformFeature.debug(" [Native Image Configuration] Registering for reflection: %s", clazz.getName());
+            registerForReflection(clazz);
+            registerForReflection(clazz.getDeclaredConstructors());
+            registerForReflection(clazz.getDeclaredMethods());
+            registerForReflection(clazz.getMethods());
```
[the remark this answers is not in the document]
gradinac: That is a good point - however, wouldn't we see the parent classes as a separate test class and traverse it while processing the test plan?
One note: this method is not only used for test classes but also for other JUnit features that require reflection

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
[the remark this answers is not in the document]
sbrannen: OK. Let's give the latter a shot.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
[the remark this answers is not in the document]
sbrannen: > That is a good point - however, wouldn't we see the parent classes as a separate test class and traverse it while processing the test plan?

Not necessarily. The superclass may be `abstract` which means it's not a "test class".

Similarly, the superclass may be a test class that is somehow excluded from the test suite (via tags, disabled, etc.).

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/NativeImageConfigurationImpl.java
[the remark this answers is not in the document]
sbrannen: Hmmm..... what is the recommend approach for logging from a `Feature`?

/cc @vjovanov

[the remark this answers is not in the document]
gradinac: I've seen logging be one of the most common pain points with build time initialization - let's say you're using SLF4J and create a logger in a static field in a class that's initialized at build time: depending on what the user plugs in as the underlying logging framework, you will initialize different classes at build time. This also means async logging/writing output to a file become unusable as the thread for logging would get started at image build time (in the builder JVM) and the file would be created on the machine building the image.

One thing I've just checked is that we initialize the `java.util` JDK package at build time - this would include the JDK logging framework under `java.util.logging`. In theory, this should mean we could safely use it for now, hopefully without affecting third party applications (this depends on how much static state from the image build remains in these classes). Another alternative would be a mini-homegrown logging framework, but I would avoid that for now

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/core/NativeImageConfiguration.java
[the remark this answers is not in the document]
gradinac: Ah, that makes sense - I fully agree in this case, I will create another utility method that registers the whole class hierarchy for reflection and use that instead when handling test classes.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/platform/PlatformConfigProvider.java
gradinac: That's a good question, I've moved this from earlier code but it does not seem necessary - we could only do a `Class.forName` here to see if the class is indeed present on the classpath.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java
gradinac: For `ConvertWith` and `AggregateWith`, we shouldn't need to register anything apart from the class and the default constructor. However, I've noticed that for some other annotations, we may need to register additional methods. For example, for any annotation consumer, we may also need:
```
org.junit.jupiter.params.support.AnnotationConsumerInitializer#isAnnotationConsumerAcceptMethod
```
To be on the safe side, I've went with registering everything. We can in the future do registration on a more fine-grained level

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/util/AnnotationUtils.java
gradinac: I agree, we should also only use `getDeclared<member>` methods here and traverse the class hierarchy.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/vintage/VintageConfigProvider.java
gradinac: I agree completely, however, I would leave that for a separate PR

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/vintage/VintageConfigProvider.java
gradinac: I agree, most likely we will require special configuration here as well

common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter/TestMethodOrderTests.java
gradinac: These tests currently only verify that including method orderers doesn't break the build.

We currently don't test the actual method order, however this is definitely something that we should do. Method ordering is special: it seems to be an integral part of the test plan. As we create the test plan during the image build, this order will be fixed for all future test executions. We could then inspect the test plan and verify if the ordering is correct.

An interesting side effect of this is that method orderers are used during the image build. As a consequence, we don't register them for reflection, but rather for build time initialization.

common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter/TestMethodOrderTests.java
gradinac: They are executed as they will be seen by the JUnit test listener: 
```

org.graalvm.junit.jupiter.NestedTests$NestedTest > 3 + 2 = 5, from a nested test class SUCCESSFUL

org.graalvm.junit.jupiter.NestedTests$NestedTest$NestedNestedTest > 3 + 2 = 5, from a nested nested test class SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$ArgumentOrderer > testOne() SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$ArgumentOrderer > testTwo() SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$DisplayNameOrderer > A - I must be first SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$DisplayNameOrderer > B - I must be second SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$MethodNameOrderer > testA() SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$MethodNameOrderer > testB() SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$RandomOrderer > testTwo() SUCCESSFUL

org.graalvm.junit.jupiter.TestMethodOrderTests$RandomOrderer > testOne() SUCCESSFUL
```

common/junit-platform-native/src/test/java/org/graalvm/junit/util/Calculator.java
gradinac: I agree - I will remove the calculator and instead rework the tests in the way you've proposed here.

sbrannen (approved):

PR #83, opened 2021-06-30
Add support for resources config inference

melix:
This pull request builds on top of #78 ,which explains the number of commits on this PR. Please take a look at commit 65c7a83 for details.

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/model/resources/ClassPathDirectoryAnalyzer.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/model/resources/ClassPathDirectoryAnalyzer.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2021, 2021 Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
+ * The above copyright notice and either this complete permission notice or at a
+ * minimum a reference to the UPL must be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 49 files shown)


melix:
I have verified that I can replace the adhoc GraalVM Micronaut Gradle plugin with this plugin if this feature is added :tada: 

I have _not_ added the same feature to the Maven plugin yet.

melix:
So this is now implemented both for the Gradle and Maven plugins. The idea is the following:

- there is a task which generates a `resources-config.json` file
- this task can be fed with explicit configuration, for example
```groovy
nativeBuild {
    resources {
        includedPatterns.add("META-INF/my-directory/.*")
    }
}
```
- or it can be fed with inferred resources, for example:
```groovy
nativeBuild {
    resource {
        inference {
            enabled = true // this is **false** by default
            restrictToProjectDependencies = false
            inferenceExclusionPatterns.add(".*[.]html")
       }
    }
}
```
- resource inference is disabled by default
- resources are typically anything which is not a `.class` file
- resource inference excludes a number of patterns by default, and will automatically detect libraries which provide a `native-image/resource-config.json` file
- resource inference is limited to project (module in Maven) dependencies by default, but can be expanded to external dependencies ("jars")

As such, this feature is completely opt-in and simplifies configuration of native image building greatly. The Micronaut plugin can be greatly simplified if this gets merged and has been tested successfully with this change.

@vjovanov wdyt?

vjovanov:
Looks good. Few questions: 
1) How are we going to extend this to predicated config, i.e., express: When type `"x.y.Z"` reachable then include pattern?
2) I don't completely understand inference: Why do we need `enabled`? Is inference more like include all files from JARs with a few exceptions?
3) `restrictToProjectDependencies = false` I think this should be opposite, i.e., `extendToProjectDependencies`. The community should anyhow provide solutions for libraries and the user should configure only their code, right? This should make it easier for them.

melix:
> 1. How are we going to extend this to predicated config, i.e., express: When type "x.y.Z" reachable then include pattern?

So there are the two aspects: for _explicit configuration_, then it's just a matter of implementing a DSL which supports this. But you're probably talking about _inference_, in which case I'm not sure if we can do inference, unless the code itself contains annotations which allow us to infer this.

> 2. Why do we need enabled?

We need it because by default, we do no inference at all, and just generate a resource config file based on the explicit configuration. If you enable it, then we start introspecting the classpath for resources, which could lead to unnecessary resources being included in the image.

> 3. he community should anyhow provide solutions for libraries and the user should configure only their code, right?

So I think it relates to your question 2. also, regarding how inference works. By defaut, we do nothing. If it is enabled, there are 2 categories of "things" we're going to introspect:

- the "local artifacts", that is to say the current project resources, and any project transitive dependency. In Gradle terminology, those are _project dependencies_: `implementation (project(":lib"))` is a _project dependency_, while `implementation("com.acme:external:1.0")` is an external one. By default inference is limited to _project dependencies_, which means that we would generate a resource config file which includes resources from the current project and the local project dependencies.
- the "external artifacts": everything else. If you have a dependency on an external artifact, or a project dependency with an external dependency, then _by default_ we wouldn't "open" the jars to analyze them and infer the resources to include.

That's why I think it's better to do "restrictToProjectDependencies" by default, because we want to encourage library authors (external dependencies, in a nutshell), to _embed_ the native configuration. This is also why engine will exclude from inference any library which provides a `META-INF/native-image/.../resource-config.json` file (I understood those were automagically interpreted by `native-image`).

vjovanov:
OK, this is great! It will also make it easier for the end-users to start quickly. Two more things that come to mind: 

>If you enable it, then we start introspecting the classpath for resources, which could lead to unnecessary resources being included in the image.

Doesn't the `inference` block itself convey the information that we want to do inference?

Also, `inference` makes me think there is a sophisticated algorithm behind it, but here it leans towards `includeAll` with a list of excluded patterns? Would maybe `includeAll` be more accurate here--also I feel the inner `excludePatterns` fits better to `includeAll`.

melix:
> Doesn't the inference block itself convey the information that we want to do inference?

Yes, but the `inference` block just configures the patterns for exclusion. You still have to provide a way to enable it (we don't do this _just_ by calling an empty `inference { }` block, for example.

> Also, inference makes me think there is a sophisticated algorithm behind it, but here it leans towards includeAll with a list of excluded patterns? Would maybe includeAll be more accurate here--also I feel the inner excludePatterns fits better to includeAll.

There is a bit of "intelligence" indeed: a default list of exclusions, that you'd have to configure explicitly otherwise, and the automatic exclusion of jars containing native-image information. If "inference" feels too smart, I'd prefer something along the lines of `autodetect`, wdyt?

vjovanov:
To me `autodetect` sounds better. Also, I think we don't need `enabled` then as `autodetect` says what we are doing. 

Could it be 
```
resources { 
  autodetect = true
}
```
or 
```
resources {
  autodetect
}
```
and only when we have special configuration use the scope?

melix:
> could be
```
resources { 
  autodetect = true
}
```

but you still need to encapsulate the parameters of auto-detection. Alternatively you need to flatten all parameters, including detection parameters, to the top-level resources configuration, but this doesn't sound right to me. Note that I know no other Gradle plugin which would automatically enable a feature just by calling its configuration block: this is IMHO an antipattern.

melix:
We can, however, introduce a `autodetect()` method, without parameters, to set the `enabled` flag automatically, this is explicit configuration and I'm fine with it.

vjovanov:
> We can, however, introduce a autodetect() method, without parameters, to set the enabled flag automatically, this is an explicit configuration and I'm fine with it.

That way one could use either `autodetect()` or `autodetect{...}` for specific cases. I think this makes sense as most people will simply use `autodetect()` and will be done.

PR #87, opened 2021-07-01
Consolidate build

### LINKED ISSUE #82: Reorganize this repository as a Gradle project with multiple subprojects
At the moment this repository contains three different projects which are interconnected using bash scripts and `mavenLocal` repository. This (obviously) isn't optimal, and results in duplication of configuration files and complex building logic.

Fixing this issue would provide a top level Gradle project that would host all of the common logic. This would also simplify the CI and streamline the release process.

melix:
This pull request consolidates the build in order to avoid tedious build stages. In particular:

- it introduces a top-level Gradle build which coordinates the builds of all subprojects
- every subproject can be built independently, or imported separatetely in the IDE if needed to focus on a specific project
- it centralizes all dependency versions in a root `gradle/libs.versions.toml` file
- it builds the Maven plugin using Gradle instead of Maven

work in progress

Fixes #82


build-logic/src/main/kotlin/org.graalvm.build.java.gradle.kts
```diff
@@ -0,0 +1,130 @@
...
+        languageVersion.set(JavaLanguageVersion.of(11))
+    })
+    (options as StandardJavadocDocletOptions).addBooleanOption("html5", true)
+}
+
+tasks.withType<Test>().configureEach {
+    testLogging {
+        events.addAll(listOf(TestLogEvent.STANDARD_OUT, TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.FAILED))
+    }
+}
```
lazar-mitrovic, line 130: Newline

build-logic/src/main/kotlin/org/graalvm/build/MavenExtension.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package org.graalvm.build
+
+import org.gradle.api.provider.Property
+
+abstract class MavenExtension {
+    abstract val name: Property<String>
+    abstract val description: Property<String>
+}
```
lazar-mitrovic, line 49: Newline

common/junit-platform-native/README.md
```diff
@@ -7,7 +7,7 @@ There are two main modes of operation using this feature:
 1. Out of the box (for tests which do not use reflection internally).
 2. With the agent run (for tests with internal reflection).
 
-Either way, artifact produced by building this repository should be included on classpath (either by using build tool plugins - as was done in [examples subdirectory](../../examples), or by adding manual configuration).
+Either way, artifact produced by building this repository should be included on classpath (either by using build tool plugins - as was done in [examples subdirectory](../../samples), or by adding manual configuration).
```
lazar-mitrovic, line 10: samples

common/junit-platform-native/build.gradle
```diff
@@ -1,85 +1,81 @@
+/*
```
lazar-mitrovic, line 1: If the entire project is now using Kotlin, should we move away from Groovy here as well?
(just a question - we can do it in another PR in order to reduce scope)

common/scripts/buildAll.sh
```diff
@@ -3,14 +3,9 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
lazar-mitrovic, line 3: I believe that don't need this script anymore, as the user can just run `./gradlew publishToMavenLocal` in the root directory.

common/scripts/testAll.sh
```diff
@@ -3,11 +3,13 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
lazar-mitrovic, line 3: This file is IMO also redundant (user should just run `./gradlew test` in the root directory)

common/scripts/updateVersions.sh
```diff
@@ -1,49 +0,0 @@
-#!/bin/bash -e
```
lazar-mitrovic, line 1: Do we have something similar in place that can replace versions in all of the samples?

As far as I saw this is critical for maven configuration in said samples, since bumping the version would be done by hand in each test.

common/utils/buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+plugins {
+    `java-gradle-plugin`
```
lazar-mitrovic, line 43: Shouldn't we replace this with `kotlin-dsl` also?

native-maven-plugin/README.md
```diff
@@ -53,7 +53,7 @@ The plugin figures out which JAR files it needs to pass to the native image and
 what the executable main class should be. If the heuristics fails with the `no main manifest attribute, in target/<name>.jar` error, the main class should be
 specified in the `<configuration>` node of the plugin. When `mvn -Pnative package` completes, an executable is ready for use, generated in the _target_ directory of the project.
 
-*You can also take a look at example project [here](../examples/maven).*
+*You can also take a look at example project [here](../samples/maven).*
```
lazar-mitrovic, line 56: This directory doesn't exist.

native-maven-plugin/build.gradle.kts
```diff
@@ -0,0 +1,194 @@
...
+
+plugins {
+    `java-library`
+    groovy
+    checkstyle
+    `java-test-fixtures`
+    id("org.graalvm.build.java")
+    id("org.graalvm.build.publishing")
+    id("org.graalvm.build.maven-functional-testing")
+    id("com.bmuschko.docker-remote-api") version "7.1.0"
```
lazar-mitrovic, line 54: I believe including docker here is a overkill. Our CI already has Maven and Gradle preinstalled, and we include a script for downloading and configuring GraalVM and Native Image (that should be replaced once we start providing toolchain fetching).

If nothing, debug log is filled with Docker logs and dependency resolutions (first image build starts at line 13k).
Also this is adding a new dependency (with potential compliance issues).

If fetching maven is the issue here, we can create a `mvnw` which is now upstreamed and official way to fetch maven.  
If you believe that we should Dockerize our builds, I'm for that, but I believe that it should be at a repository level, and that we should run root Gradle project inside of said Docker environment.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/GraalVMContainerController.groovy
```diff
@@ -0,0 +1,106 @@
...
+
+import groovy.transform.CompileStatic
+import groovy.transform.PackageScope
+import org.spockframework.util.NotThreadSafe
+import org.testcontainers.Testcontainers
+import org.testcontainers.containers.BindMode
+
+@CompileStatic
+@NotThreadSafe
+class GraalVMContainerController {
```
lazar-mitrovic, line 52: This entire logic would be unnecessary if we drop Docker from the equation.

lazar-mitrovic (commented):
Overall looks good, I'm glad we are getting rid of hacky shell scripts :)

build-logic/build.gradle.kts
```diff
@@ -38,11 +38,18 @@
...
  */
-package org.graalvm.buildtools;
 
-// Do not edit this file directly!
-// Edit 'version.properties' in root of the repository and then use 'common/scripts/updateVersions.sh' to keep everything in sync.
-public class VersionInfo {
-    public static final String JUNIT_PLATFORM_NATIVE_VERSION = "0.9.2-SNAPSHOT";
+plugins {
+    `kotlin-dsl`
+    id("com.github.joschi.licenser") version "0.6.1"
```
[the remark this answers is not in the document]
melix: Yes, it verifies license headers _and_ updates them. The same can be done with checkstyle + IDE, yes, although I have found that my IDE doesn't consistently create license headers. It's also not granted that anyone would use an IDE to develop. That said:

- the checkstyle rules only verifies existing source sets (not build files, not samples, ...)
- this plugin is a build time dependency, not a runtime one, with AL2 in any case. It's not even used for _user code_, I can barely imagine how this could be a problem.

build-logic/src/main/kotlin/org.graalvm.build.common.settings.gradle.kts
```diff
@@ -0,0 +1,107 @@
...
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
```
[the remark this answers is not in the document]
melix: Indeed it's an oversight :)

common/scripts/updateVersions.sh
melix: No, we don't. We _could_, but it's not strictly required: the tests pass properties which override whatever is used in the POM file. So the important bit is that the properties are defined, their actual value isn't much important. I can write a task to do the update in any case.

common/utils/buildSrc/build.gradle.kts
melix: No, this `buildSrc` project only uses Java code.

native-maven-plugin/build.gradle.kts
melix: Did you look at the [commit message introducing this](https://github.com/graalvm/native-build-tools/pull/87/commits/d96237b8242f1c1054c76ec4ba925590c873e940) ? The problem is not downloading Maven, it's having reproducible builds. With Maven it's barely impossible to have reliable builds because it insists on copying everything into the `.m2` repository. This means that:

- builds are order dependent: depending on what order you execute the dependent projects, dependencies might be cached, including snapshot dependencies, ignoring local changes unless you run systematically with `-U`
- we cannot cache the external dependencies from locally built dependencies in isolation, meaning that as soon as we build one project (say the junit native extension), dependencies are put in .m2 _but also_ the project themselves, making it impossible to have integration tests which start from a clean state _without having to redownload everything everytime_

So what this does is basically introducing docker in order to be able to properly isolate test executions, and not having tests accidentally passing because another test passed. To reduce build time, we create a container which _predownloads_ the external dependencies we need, and each test uses that image. While it introduces a layer of complexity, this is *much more reliable* than other mechanisms: any other way would require a way to clean up the local .m2 repository from "pollution" of the tests execution (capture state before and after).

As for the compliance thing, again I can't imagine how this could be an issue.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/GraalVMContainerController.groovy
melix: Blame Maven :)

common/junit-platform-native/build.gradle
melix: I don't think we have to. It's perfectly fine to have a mix of Groovy and Kotlin scripts. I used Kotlin for the new ones to get IDE support.

common/scripts/buildAll.sh
melix: I reintroduced it _if_ we want users to be able to execute the samples themselves. We may want to provide something different, though.

common/scripts/testAll.sh
melix: Just FTR, `./gradlew test` runs _unit tests_. `./gradlew functionalTest` runs functional tests. `./gradlew fullFunctionalTest` runs functional tests, for all supported Gradle versions, for Gradle. And `./gradlew check` runs all verifications (tests, checkstyle, ...).

native-maven-plugin/build.gradle.kts
graemerocher: I don’t see how this impacts compliance it is just a build time dependency and ensures we have better Maven tests

common/scripts/updateVersions.sh
lazar-mitrovic: Having such task would be nice since I wouldn't like having `samples` that point to a older release in the master. I know that it would confuse me as a end user (since I would just copy a sample and build upon that).

build-logic/build.gradle.kts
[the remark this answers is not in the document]
lazar-mitrovic: Thing is that there is no optimal solution to this issue. 

Checkstyle currently serves as a guard against merging unlicensed (or wrong licensed files) as it forces the user to update headers manually. Yes this is tedious, however automatic tool might create issues like [this](https://github.com/graalvm/native-build-tools/pull/87/files/b6ca784221ca9c183e8f02e8a592eb1cefa30df9#r663263153) which could possibly fall through (due to all of the tests passing). I believe that build files and samples could be included in the checkstyle definition as well (for the build files there is [file filter](https://github.com/melix/native-build-tools/blob/051df2f64bf0b0a29f012ba032a85d518a763fb3/config/checkstyle.xml#L244) change as well as forcing checkstyle to include files which are not in the sourcesets).

As you said correctly, user could work without IDE, but if I'm not mistaken, user can also not run the check task and push unlicensed code to the PR. At the end this is a trade-off between developer convenience and reviewer convenience, where in the main GraalVM repository we focus on later (since developer is responsible for making sure that his code is properly formatted).

As Graeme said that build time dependencies are OK, I'm fine with merging this change, even though I would personally avoid this path.

common/scripts/buildAll.sh
lazar-mitrovic: Well, users could:
```
gradle publishToMavenLocal
cd samples/<sampleName>
gradle nativeBuild
```
right?

common/scripts/testAll.sh
lazar-mitrovic: In that case this script wasn't even working correctly (`testAll` is a wrong name since it runs only a single gradle and maven test).

native-maven-plugin/build.gradle.kts
lazar-mitrovic: I still believe that all of that could be achieved without the dependency on Docker. 
For example you can:
* use the `mvnw` to download separate distribution of maven
* [change the `.m2` location ](https://stackoverflow.com/a/57223696) to the empty directory
* run the bootstrapping project and inject the fake repository
* save copy of the resulting repository in order to re-inject it after subsequent test execution

Current state seems to me as if we are shooting a fly with a cannon. However, I'm fine with merging this as is and fixing it in a follow-up PR in order to move faster. One thing that should be changed prior to merging this is the removal of [GraalVM downloading step in the CI](https://github.com/melix/native-build-tools/blob/051df2f64bf0b0a29f012ba032a85d518a763fb3/.github/workflows/native-maven-plugin.yml#L26-L31) since we are already downloading GraalVM through a Docker image.

melix: This is still significantly slower, in particular for local builds, since for every change you'd have to re-download every single dependency Maven needs. The docker images can be layered which makes their build much faster, especially for local builds (for CI it doesn't matter yet since we don't use caching, but we could).

Also even if you change the .m2 location, you'd have to use a _separate_ m2 location for every single test. And your local settings.xml (in your user home) would influence the build. It's barely impossible to properly isolate Maven builds, unfortunately.

melix: > One thing that should be changed prior to merging this is the removal of GraalVM downloading step in the CI since we are already downloading GraalVM through a Docker image.

Good catch!

common/scripts/updateVersions.sh
melix: I'd be keen on doing this in a separate PR later, since it will require some more rework which would make this PR even longer, wdyt?

lazar-mitrovic: Sounds good to me.

lazar-mitrovic (approved):
Maven build fails due to CI runner using Java 8 as default.
This can be fixed by adding
```yaml
- uses: actions/setup-java@v1
        with:
          java-version: '11'
```
to the CI config file. You can then merge this PR when ready.

common/scripts/updateVersions.sh
melix: I actually ended up doing it in this PR, since it involved less work than I initially thought.

Issue #84, opened 2021-06-30
Test listener overwrites `test_ids` file when executed in parallel. 


lazar-mitrovic:
When invoking `test` task with large amounts of tests, multiple JVM processes are being spawned in parallel. When test listener is triggered, all of those processes overwrite the same file (instead of appending to it). This results in `test_ids` file which contains only the tests that were executed in the last JVM process (which is undeterministic and might result in a empty file). 

This behavior can be triggered by invoking the test task with 500+ tests on Gradle plugin.

Reproducer:
```bash
git clone git@github.com:rsocket/rsocket-java.git
curl https://gist.githubusercontent.com/lazar-mitrovic/b075423362aef58c2d7d1b4132e6899e/raw/cb4ef77a4989b1985c2d1f999136037e40acba63/rsocket-repro.diff -O
cd rsocket-java
git apply ../rsocket-repro.diff
./gradlew clean rsocket-core:test rsocket-core:nativeTest
```
Issue dissapears when working with a smaller number of tests - uncomment the test filter from the end of `rsocket-core/build.gradle` and it should work as intended.

AFAIK this issue is also present in listener from the current [JUnit5 master](https://github.com/junit-team/junit5/blob/main/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java).

Possible fix might be to separate listener output files, or to introduce some sort of locking mechanism that would ensure that file is accessed correctly.

Potential workaround for the end-users ATM might be to use some Gradle magic to force all tasks to be executed sequentially and in the same JVM, but this would come with a extreme performance penalty (and I'm unsure if such configuration option is even present).

cc: @sbrannen @gradinac @melix

melix:
That's an interesting discovery indeed. This could potentially be fixed by implementing this "dry run" mode (#76), granted that discovery is done in a single process.

Alternatively, the test listener could, instead of a file, be passed an output directory. Then:

- the build tool would be responsible for passing a "clean" directory when the tests are started
- the listener would be responsible for generating a unique file in that directory (say `test_ids.txt` exist, create `test_ids2.txt`, ...)
- the build tool would merge the test ids files into a single one

In #71 I already added the ability to pass in the path to the generated file, but we could use a _directory_ instead. That would probably be cheaper/more reliable than the dry run for now.

sbrannen:
[...]

PR #86, opened 2021-07-01
Skip Maven native testing when there is no test

sdeleuze:
Please wait for @aclement validation before merging this PR.

Validated with the `maven` example and the repro provided in #81.
I did not implement logging because this is what Surefire does for regular tests, and it kind of make sense with Maven regular output where the name of the project and of the plugin are already displayed.

Closes #81

PR #89, opened 2021-07-02
Support concurrent use cases in the UniqueIdTrackingListener

sbrannen:
Please wait for approval from @aclement before merging this PR.

This set of commits serves as a fix for #84.


sbrannen:
Please note that there is currently no mechanism in place to clean up the old test UID files (from a previous run of the build) at the start of the build.

This still needs to be implemented for the Gradle and Maven plugins.

sbrannen:
Analogous to https://github.com/graalvm/native-build-tools/pull/86#issuecomment-872303476, I am submitting this PR as a VMware employee.

@gilles-duboscq, can you please help ensure that the bot allows me to submit PRs as well?

gradinac (commented):
Thank you very much for fixing this! Overall, it looks good - I've left a single minor comment. I didn't mark this as approved so that we don't merge it yet by accident.

gradinac:
> Please note that there is currently no mechanism in place to clean up the old test UID files (from a previous run of the build) at the start of the build.
> 
> This still needs to be implemented for the Gradle and Maven plugins.

I think we should prepare a PR for this and merge it right after these changes - I assume that without it, we could hit errors if the test set changes in between runs (for example, an existing test gets deleted)

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -85,37 +84,49 @@ public void duringSetup(DuringSetupAccess access) {
...
-        } catch (FileNotFoundException e) {
-            listenerStream = getClass().getClassLoader().getResourceAsStream(UniqueIdTrackingTestExecutionListener.FILE_NAME);
-        }
-        try (BufferedReader br = new BufferedReader(new InputStreamReader(listenerStream))) {
-            List<String> classes = br.lines().collect(Collectors.toList());
-            System.out.println("[junit-platform-native] Running in 'test listener' mode.");
-            return classes.stream().map(DiscoverySelectors::selectUniqueId).collect(Collectors.toList());
-        } catch (IOException | NullPointerException e) {
-            System.out.println("[junit-platform-native] Running in 'test discovery' mode. Note that this is a fallback mode.");
+            Path outputDir = new UniqueIdTrackingListener().getOutputDir();
```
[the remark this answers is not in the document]
sbrannen: That's only true in this variant of the listener. The official listener actually uses `ConfigurationParameters` available in the `TestPlan` beginning with JUnit Platform 1.8.

https://github.com/junit-team/junit5/blob/9a1db7309f908f474b6042c9eb68c1e596e58453/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java#L197

Thus, we will require a different approach once we switch to the official listener.

My goal here was to get something in place that works now.

Once we switch to the official listener, I think the Gradle and Maven plugins should always specify an explicit output directory (instead of the way the current implementation of the Maven plugin relies on the default output directory selection based on convention). Then we would no longer need to rely on the behavior in `getOutputDir()`.

Does that make sense?

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/UniqueIdTrackingListener.java
```diff
@@ -0,0 +1,261 @@
...
+        try {
+            outputFile = getOutputFile();
+        } catch (IOException ex) {
+            logger.error(ex, () -> "Failed to create output file");
+            // Abort since we cannot generate the file.
+            return;
+        }
+
+        logger.debug(() -> "Writing unique IDs to output file " + outputFile.toAbsolutePath());
+        try (PrintWriter writer = new PrintWriter(Files.newBufferedWriter(outputFile, StandardCharsets.UTF_8))) {
```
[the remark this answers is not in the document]
sbrannen: Please keep in mind that this listener will be superseded by the official one in JUnit Platform 1.8.

Thus, if you want additional logging included, please add a note to https://github.com/junit-team/junit5/issues/2619 to have it reopened before JUnit 5.8 M2 is released or create a new issue against the JUnit 5 issue tracker.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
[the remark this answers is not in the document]
gradinac: Ah, in that case, it makes sense to leave it as is - I agree, the build tool should explicitly specify this directory

sbrannen:
> I think we should prepare a PR for this and merge it right after these changes - I assume that without it, we could hit errors if the test set changes in between runs (for example, an existing test gets deleted)

Yes, if we don't clean up the old files we will definitely run into issues that like (phantom test UIDs).

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/UniqueIdTrackingListener.java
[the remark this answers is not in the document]
gradinac: That's a good point - I forgot about that while initially looking into this PR, my bad :) If we end up needing it, I'll post a comment on the issue

Looking into the logging, I've noticed that the official listener uses `org.junit.platform.commons.logging.Logger` - however, this one seems to be a facade for java.util.logging so we shouldn't hit class initialization related problems (as JUL is already initialized at build time).

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -226,7 +232,7 @@ private static NativeImageOptions createMainExtension(Project project) {
         return buildExtension;
     }
 
-    private static NativeImageOptions createTestExtension(Project project, NativeImageOptions mainExtension) {
+    private static NativeImageOptions createTestExtension(Project project, NativeImageOptions mainExtension, File testResultsDir) {
```
melix, line 235: :x: This shouldn't use a `File` or the configuration is order dependent. I will look into fixing this properly.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -144,19 +144,25 @@ private void configureJavaProject(Project project, Provider<NativeImageService>
...
         TaskCollection<Test> testTask = findTestTask(project);
         Provider<Boolean> testAgent = agentPropertyOverride(project, testExtension);
         TaskProvider<ProcessGeneratedGraalResourceFiles> processAgentTestFiles = registerProcessAgentFilesTask(project, PROCESS_AGENT_TEST_RESOURCES_TASK_NAME);
 
         testTask.configureEach(test -> {
             testListDirectory.set(new File(testResultsDir, test.getName() + "/testlist"));
             test.getOutputs().dir(testResultsDir);
-            test.systemProperty("graalvm.testids.outputdir", testListDirectory.getAsFile().get());
+            // Set system property read by the UniqueIdTrackingListener.
+            test.systemProperty("junit.platform.listeners.uid.tracking.output.dir", testListDirectory.getAsFile().get());
```
melix, line 165: This is also where we should register an action to cleanup the previously executed code. Something like:

```
test.doFirst("cleanup test ids", task -> {
                File dir = testListDirectory.getAsFile().get();
                if (dir.exists()) {
                    GFileUtils.deleteDirectory(dir);
                }
            });
```

melix (changes requested):
The logic is good. It misses however:

- cleanup of the previous test ids execution directory
- proper wiring of the output directory with Gradle

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
sbrannen: Thanks

common/junit-platform-native/build.gradle
```diff
@@ -4,6 +4,7 @@ plugins {
     id 'maven-publish'
     id 'signing'
     id 'org.graalvm.build.interop.publishing'
+    id 'eclipse'
```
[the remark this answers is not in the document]
sbrannen: I tend to shy away from Buildship, because I have always run into issues with it.

If it's a problem for the team, feel free to remove it, and I'll deal with it some other way.

common/junit-platform-native/build.gradle
```diff
@@ -33,7 +34,9 @@ clean.doLast {
 }
 
 test {
```
lazar-mitrovic, line 36: We should definitely add a task that cleans the `test_ids` directory here as well @melix

melix: This shouldn't be necessary. Having to add something to `clean` means that something is not declared as an output. In this case the directory where we write test ids is registered, it will be cleaned before every execution so I think we're fine. In any case, users (including us) shouldn't have to care about creating a task to do the cleanup.

lazar-mitrovic: I'm not sure if I understood you correctly, we do declare that `clean` deletes entire `build` directory here:
https://github.com/graalvm/native-build-tools/blob/046f65081fdbe2473f5edc56c8ca679a24024a58/common/junit-platform-native/build.gradle#L32-L34

 but I don't see where we add `test_ids` to outputs of the `test` task?

melix: The whole point of `clean` _is_ to delete the `build` directory. You should never have to do this. Ever. If you have to something is wrong.

lazar-mitrovic: OK so we should remove those lines, but where do we define that `test_ids` directory should be emptied prior to running `test` task (or where do we define that `test` has `test_ids` folder as it's output?

melix: This is done in https://github.com/sbrannen/native-build-tools/pull/1

lazar-mitrovic: Тhanks I missed that one :)

lazar-mitrovic: That should fix Gradle plugin, but it doesn't fix tests for `junit-platform-feature` (which are separate to the plugin).

melix: I didn't look into those tests. But since each test should be isolated, it should be each test's responsibility to do the cleanup. I'll take a look.

aclement (approved):
Looks good to me, approved.

melix:
Merged, thank you all!

PR #96, opened 2021-07-07
Introduce a command line argument provider for building native image options

melix:
This commit introduces a native image command line argument provider,
which basically converts the native image options into the list of
CLI arguments which needs to be passed to the native image tool.

This was previously found in the native build task itself, but it
prevented other tools or plugins to reuse the logic for their own
needs. In particular, Micronaut has its own native image build task
which builds _within a Docker image_, and therefore would have had
to copy the whole logic.

Using this, the logic can be shared by other Gradle plugins/builds.


melix:
retargetted to 0.9.2

Issue #77, opened 2021-06-25
Add support for multiple test images


melix:
## Proposal

At the moment, the Gradle plugin is wired so that we generate a test image for unit tests. However, Gradle supports multiple test tasks and multiple test source sets. For example it's pretty common to have a functional test source set, an integration test source set, ...

Not all those would require a native image build, but we should provide a configuration mechanism so that native images for all those can be generated. This probably means reworking a bit the integration in order to be "test task independent".

vjovanov:
"Not all those would require a native image build" - I think all of them need an image build.

vjovanov:
To understand better, multiple test tasks, can you give an example of how it is done for java. Then we can see how to do for native.

melix:
> To understand better, multiple test tasks, can you give an example of how it is done for java. Then we can see how to do for native.

I don't think there's anything to do for native. This is really simple. Take this Gradle plugin for example: it has unit tests, which basically verify very quickly the algorithms, and it has functional tests, which are full fledged tests which spawn a Gradle daemon, execute a build and verify the results. In this case, most of the functional tests don't make sense to execute in native mode (forget about the fact they use Spock so are no executable natively yet):

- either the code of the tests themselves is executed in native, but then they have no value because the only thing they will do is execute assertions and setup
- or what you'd want to run is the _code under test_, which is Gradle itself, which, in this case, isn't executed in native either, it would always run in JVM mode

But in lots of cases, it would actually be the other way around: take a Micronaut app. In this case, you will also find a unit test directory, and a functional test one. Again, the unit tests will verify the basics, do a lot of mocking, etc... They are pretty fast to execute, but their goal is not to verify the actual behavior of the application wrt to clients. That's the role of _functional tests_, which live in a separate source set. Then what Micronaut does is actually building a native version of the _application_, and run tests in JVM mode _against the native image_. This makes a lot of sense, since, again, what you want to test in this case is the behavior of the application, _not the tests themselves_.

[...]

PR #98, opened 2021-07-09
Introduce a top-level `graalvmNative` DSL container

melix:
This is a follow-up to the idiomatic Gradle conversion. Before this commit,
there were 2 extensions configured at the top level, with the same type,
to configure the main native image and the test native image. Those
extensions were respectively named `nativeBuild` and `nativeTest`. In
addition, there were 2 tasks with similar names, which had slightly
different semantics: `nativeBuild` would build the main image, but
`nativeTest` would _run_ the tests. The task which would build the native
test image was called `nativeTestBuild`.

There were therefore a few issues to fix:

- the confusion between the extension names and the task names, making it
unclear whether you were configuring one or the other
- the fact that multiple extensions of the same type were created, which is
considered an anti-pattern in Gradle and makes it complicated to get the
expected extensions (See https://github.com/gradle/gradle/issues/17622)
- it was complicated to add support for more test images, as hinted by
the follow up issue #77

So this commit changes the configuration to use an idiomatic container,
named `graalvmNative` which can contain _many_ images. For now it contains only
two:

```
graalvmNative {
    binaries {
        main {
            // .. the main native image
        }
        test {
            // .. the test image
        }
    }
}
```

In addition, it renames the confusing task names to something closer to the
actual semantics of other tasks in Gradle:

- `nativeBuild` -> `nativeCompile` (`build` does much more)
- `nativeTestBuild` -> `nativeTestCompile`

### DIFF
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -43,19 +43,19 @@ pluginManagement {
 ```
 _(this step will be redundant once this plugin is published to the Gradle Plugin Portal)._
 
-After that, we can configure the image build by using a `nativeBuild` configuration block:
+After that, we can configure the image build by using a `graal` configuration block:
 ```groovy
-nativeBuild {
-  imageName = "my-app"
-  mainClass = "org.test.Main"
-  verbose = true
-  fallback = false
+graal {
+    main {
+        imageName = "my-app"
+        mainClass = "org.test.Main"
+        verbose = true
+        fallback = false
+    }
 }
 ```
 
-The plugin then adds `nativeBuild` and `nativeRun` tasks that build and run the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
-
-> _To help ease the community transition, at the moment we support a subset of the configuration syntax and aliased task names from the most popular unofficial_ GraalVM  Gradle _plugins (_`_io.micronaut.application_` _and_ `_com.palantir.graal_`_). Note that this behavior might eventually be deprecated._
+The plugin then adds `nativeAssemble` and `nativeRun` tasks that respectively creates a native executable and runs the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
 
 #### Testing in _Gradle_
 
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -45,8 +45,8 @@ _(this step will be redundant once this plugin is published to the Gradle Plugin
 
 After that, we can configure the image build by using a `graal` configuration block:
 ```groovy
-graal {
-    nativeImages {
+javaNative {
+    images {
         main {
             imageName = "my-app"
             mainClass = "org.test.Main"
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -46,11 +46,13 @@ _(this step will be redundant once this plugin is published to the Gradle Plugin
 After that, we can configure the image build by using a `graal` configuration block:
 ```groovy
 graal {
-    main {
-        imageName = "my-app"
-        mainClass = "org.test.Main"
-        verbose = true
-        fallback = false
+    nativeImages {
+        main {
+            imageName = "my-app"
+            mainClass = "org.test.Main"
+            verbose = true
+            fallback = false
+        }
     }
 }
 ```
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -43,7 +43,8 @@ pluginManagement {
 ```
 _(this step will be redundant once this plugin is published to the Gradle Plugin Portal)._
 
-After that, we can configure the image build by using a `graal` configuration block:
+After that, we can configure the image build by using a `javaNative` configuration block:
+
 ```groovy
 javaNative {
     images {
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -43,10 +43,10 @@ pluginManagement {
 ```
 _(this step will be redundant once this plugin is published to the Gradle Plugin Portal)._
 
-After that, we can configure the image build by using a `javaNative` configuration block:
+After that, we can configure the image build by using a `jvmNative` configuration block:
 
 ```groovy
-javaNative {
+jvmNative {
     images {
         main {
             imageName = "my-app"
@@ -58,7 +58,7 @@ javaNative {
 }
 ```
 
-The plugin then adds `nativeAssemble` and `nativeRun` tasks that respectively creates a native executable and runs the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
+The plugin then adds `jvmNativeCompile` and `nativeRun` tasks that respectively creates a native executable and runs the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
 
 #### Testing in _Gradle_
 
--- a/common/docs/ANNOUNCEMENT.md
+++ b/common/docs/ANNOUNCEMENT.md
@@ -43,11 +43,11 @@ pluginManagement {
 ```
 _(this step will be redundant once this plugin is published to the Gradle Plugin Portal)._
 
-After that, we can configure the image build by using a `jvmNative` configuration block:
+After that, we can configure the image build by using a `graalvmNative` configuration block:
 
 ```groovy
-jvmNative {
-    images {
+graalvmNative {
+    binaries {
         main {
             imageName = "my-app"
             mainClass = "org.test.Main"
@@ -58,7 +58,7 @@ jvmNative {
 }
 ```
 
-The plugin then adds `jvmNativeCompile` and `nativeRun` tasks that respectively creates a native executable and runs the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
+The plugin then adds `nativeCompile` and `nativeRun` tasks that respectively creates a native executable and runs the main class (as one might expect ☺). If the reflection configuration is necessary for the Native Image building, this plugin also provides a simple option that activates the `[native-image-agent](https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/#assisted-configuration-of-native-image-builds)` without any additional user setup. More information (and _Kotlin_ configuration syntax) is available in the [documentation](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md).
 
 #### Testing in _Gradle_
 
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -129,12 +129,14 @@ This plugin works with the `application` plugin and will register a number of ta
 
 The main tasks that you will want to execute are:
 
-- `nativeBuild`, which will trigger the generation of a native executable of your application
+- `nativeAssemble`, which will trigger the generation of a native executable of your application
 - `nativeRun`, which executes the generated native executable
-- `nativeTestBuild`, which will build a native image with tests found in the `test` source set
+- `nativeTestAssemble`, which will build a native image with tests found in the `test` source set
 - `nativeTest`, which will <<testing,execute tests>> found in the `test` source set in native mode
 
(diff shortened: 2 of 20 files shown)


melix:
The only thing I'm unsure is if we should add another nesting level. Basically currently the extension _is_ the `NativeImageOptions` container. Which is fine for now, but we might want to provide more options in the future, which are _not_ specific to native image.

So something like:

```
graal {
    images {
       // this is where we configure NativeImageOptions
    }
}
```

Another advantage is that other plugins can "find" the right extension by type: `extensions.findByType(GraalExtension)` instead of by name with the current state. The Micronaut plugin will typically need this (and this is one of the drivers for this change, as currently the plugin has to perform some _very nasty_ workarounds to get a handle on the extensions).

@graemerocher @lazar-mitrovic wdyt?

melix:
(moved to 0.9.3 to unblock the release)

graemerocher:
> @graemerocher @lazar-mitrovic wdyt?

Having another nesting level makes sense to me. Not sure `images` is the right name.

melix:
Not sure I like `images` either because of the potential confusion with docker images (in particular with plugins like Micronaut which would configure both). Alternatives:

- `native`: short, makes it clear we're talking about `graal -> native`
- `nativeImages`: longer but makes it clearer it talks about the `native-image` tool
- `executables`: a more abstract way to think about the images we're producing: they are actually (native)  _executables_
- `nativeExecutables`: more explicit than the previous one

graemerocher:
`nativeImages` makes sense to me

vjovanov:
First thoughts:
1) This is a Java Native plugin that currently uses GraalVM, but will hopefully one day become an integral part of `java`. That said, name `graal` doesn't make sense here. `native` would make the most sense if not taken in the Gradle ecosystem.
2) What would be the options that are not specific to native compilation? This is a Java native plugin for Gradle. Are we drifting towards a general GraalVM plugin? 
3) I would like this plugin to have a deprecation policy of at least few months. Breaking peoples code unannounced is never a good thing.

melix:
> This is a Java Native plugin that currently uses GraalVM, but will hopefully one day become an integral part of java. That said, name graal doesn't make sense here. native would make the most sense if not taken in the Gradle ecosystem.

Native is definitely taken in the Gradle ecosystem. Gradle itself has a native plugin (for C++) and for Swift. There's also the Kotlin native plugin. If `graal` shouldn't appear, maybe something like `nativeJava` would make more sense, to avoid the overload.

> What would be the options that are not specific to native compilation? This is a Java native plugin for Gradle. Are we drifting towards a general GraalVM plugin?

In Gradle we tend to model "applications", "libraries", "microservices", ... rather than tasks. This means that you have "higher order" concepts and that it's better to provide a _model_ of what you build, rather than explain how to build it. The goal of this extension is therefore to declare what _deliverables_ we'd like to have as native Java executables. We'd have the main application and the tests for now. For tests, we already know that we'd like to be able to declare more source sets as participating to the deliverables as native tests, so it makes sense to have this additional level to make this _declarative_. In the future, however, I can definitely imagine more kinds of deliverables. For example, even though to my knowledge this isn't possible yet, it would make a lot of sense to expose _libraries_ as shared libraries. That is to say, be able to provide "native libraries" instead of just native applications. There would be lots of advantages doing this: exposing code written in Java to other native languages, reducing the time to build a native image, ... So I can definitely envision in the future something which would require a configuration block to expose all kind of options (what target platforms to build for, so that Gradle creates a matrix automatically, etc...).

> I would like this plugin to have a deprecation policy of at least few months. Breaking peoples code unannounced is never a good thing.

I understand this and I would definitely agree post-1.0. Before, I think it's a bonus if you can, but it's more important for me to reach a good model, idiomatic, and to document the plugin rather than doing backwards compatibility: my experience with Gradle users is that they are relatively keen on upgrading, even if it includes breaking changes, as long as they are documented and that the benefit is obvious (we conducted a survey in the beginning of this year on this topic). But, in this particular case, there are difficulties with regards to deprecation in any case:

- there's no way, in Gradle, to deprecate the use of an _extension_. Currently we create 2 extensions, and there's no possibility to capture the use of the extension and throw a deprecation warning.
- there's nothing _static_ that we can use to explain that the extension in use is deprecated either. Even if using the Kotlin DSL, the generated code to access extensions wouldn't have deprecation markers, so the user wouldn't know

So I would prefer to focus on improving the plugins and writing good documentation (using Asciidoctor for example) than providing backwards compatibility pre-1.0. Once 1.0 is out, we'll clearly have to be more conservative, which makes it even more important for me to be able to break in pre-releases.

melix:
So after discussion, we need:

- a new name for the top-level container: not `graal`, but something like `nativeJava` or `javaNative`
- a new name for the "build" task, currently called `nativeBuild` but corresponding more to the `jar` task. `nativeBuild` could actually be created too, mapping to the `build` task in Gradle but for native instead

ideas welcome!

sdeleuze:
Those tricky issues were the main reason why we took the `nativeBuild` + `nativeTest` road, but point taken on the limitations raised by @melix.

For a top level block, maybe we can use keep using `nativeBuild` as a shortcut for `nativeBuildTools` (using the longer `nativeBuildTools` is also an option). If we rename the tasks that seems a reasonable outcome to me.

vjovanov:
I vote for either `javaNative` (analogy with react.native, Scala Native, Kotlin Native) or simply `nativeImage` given we can't use `native`. I don't think `nativeCompile` is accurate as this block contains test information as well.

graemerocher:
`nativeImage` makes sense to me

melix:
If we use `nativeImage` there's a bit of redundancy in the DSL:

```
nativeImage {
    nativeImages {
       main {
           // configure main image
       }
       test {
          // configure test image
       }
       integTest {
         // configure integ test image
       }
}
```

so possibly I'd be more keen on:

```
javaNative {
    registerTestSourceSet(sourceSets.functionalTest)
    images {
       ...
    }
}
```

which looks nicer to me.

vjovanov:
Agree with the redundancy, but if we go with `nativeImage` there would be no nested scope.

Why do we need the nested scope?

melix:
> Why do we need the nested scope?

Because as explained before there will be more DSL methods to create, for example, additional images from different source sets (e.g `functionalTest`) and potentially more configuration which is not directly related to native images (imagine configuring defaults or the target OS matrix). If we don't have that indirection level, then we conflate the configuration of individual images with the configuration of the plugin itself, and we can't provide a nice way for other plugins (e.g the Micronaut plugin) to interact with this one.

That's why I went with:

```
javaNative {
    images {
        ...
    }
}
```

Which allows other plugins to get a handle on the native plugin configuration (the top level extension), then get the list of configured images from it (via `.images`), etc.

vjovanov:
OK, so we introduce the extra scope (giving away some brevity) to allow easier future extensions. Is there a way to get the brevity back by stating `registerTestSourceSet(sourceSets.functionalTest)` inline with `functionalTest {` in a more compact form.

Overall, I like the final `javaNative` proposal.

melix:
> Is there a way to get the brevity back by stating registerTestSourceSet(sourceSets.functionalTest) inline with functionalTest { in a more compact form.

At this stage I am not sure how to make it more compact without breaking idiomatic Gradle or principle of least surprise. The `functionalTest` block would typically configure the functional test image, but there's no reason it _implies_ a source set (could be a native image you build from the main source set with different options, in theory).

On a side note, the PR contains the mentioned changes and I added deprecation of the `nativeBuild` and `nativeTest` extensions (meaning you can still use them but it warns that you should use the `javaNative` block), as well as the deprecation of `nativeBuild` in favor of `nativeAssemble`.

common/docs/ANNOUNCEMENT.md
```diff
@@ -43,19 +43,21 @@ pluginManagement {
 ```
 _(this step will be redundant once this plugin is published to the Gradle Plugin Portal)._
 
-After that, we can configure the image build by using a `nativeBuild` configuration block:
+After that, we can configure the image build by using a `graal` configuration block:
```
vjovanov, line 46: javaNative

common/docs/ANNOUNCEMENT.md
```diff
@@ -43,19 +43,21 @@ pluginManagement {
...
-  mainClass = "org.test.Main"
-  verbose = true
-  fallback = false
+javaNative {
+    images {
+        main {
+            imageName = "my-app"
+            mainClass = "org.test.Main"
+            verbose = true
+            fallback = false
```
vjovanov, line 54: Can we make this the default?

common/docs/ANNOUNCEMENT.md
melix: Good catch!

common/docs/ANNOUNCEMENT.md
melix: Can you clarify what default you're talking about? The defaults are configured at https://github.com/melix/native-build-tools/blob/40c44217afd8b5e9f0e272190740c0f331ee8140/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java#L218-L231

You'd want `verbose` to be true by default?

sdeleuze:
For the task name, could we discuss `nativeAssemble` versus `nativeCompile`?

melix:
> For the task name, could we discuss `nativeAssemble` versus `nativeCompile`?

Sure. We're trying to figure out what's the _closest_, semantically speaking, to what a `jar` is for the native java world, and what `build` is in the Gradle world. `build` is confusing to Gradle users because it corresponds to the `build` lifecycle task, which compiles, runs test and other verifications. So, what is the native image task output? It's a single deliverable, corresponding either to a runnable executable (for applications) or a library file (for libraries). In both cases, in the Java world, this corresponds to the JAR deliverable. So the "closest" would be `jar`, but naming the task `nativeJar` or `nativeTestJar` would sound awkward, since there's no jar generated. However, there's this `assemble` task in Gradle which is similar to `jar` except that:

- it's independent of the ecosystem: it's used for native C++ to build the executable/libraries
- it's a lifecycle task (does nothing by itself but delegates to `jar` for Java and `nativeXXX` for C++

As such, it's indeed very close to what we need. In comparison, `compile` in the Java world has a strong connotation of _compiling Java sources_, hence using the Java compiler, and the output is `.class` files. I don't think it matches as closely the semantics of building a native image.

sdeleuze:
>Sure. We're trying to figure out what's the closest, semantically speaking, to what a jar is for the native java world, and what build is in the Gradle world. build is confusing to Gradle users because it corresponds to the build lifecycle task, which compiles, runs test and other verifications

Ok, build is not what we need.

> So, what is the native image task output? It's a single deliverable, corresponding either to a runnable executable (for applications) or a library file (for libraries).

Ok with that as well.

> In both cases, in the Java world, this corresponds to the JAR deliverable. So the "closest" would be jar, but naming the task nativeJar or nativeTestJar would sound awkward, since there's no jar generated. However, there's this assemble task in Gradle which is similar to jar except that:
>
>it's independent of the ecosystem: it's used for native C++ to build the executable/libraries
>it's a lifecycle task (does nothing by itself but delegates to jar for Java and nativeXXX for C++
>
>As such, it's indeed very close to what we need. In comparison, compile in the Java world has a strong connotation of compiling Java sources, hence using the Java compiler, and the output is .class files. I don't think it matches as closely the semantics of building a native image.

Couldn't we have `assemble` that delegates to a `nativeCompile` task?

> As such, it's indeed very close to what we need. In comparison, compile in the Java world has a strong connotation of compiling Java sources, hence using the Java compiler, and the output is .class files. I don't think it matches as closely the semantics of building a native image.

Invoking native image compiler is really a compilation, and retaining this term work make more sense from my point of view. The native build tools declaration and the `native` prefix makes it clear it is not about `javac` but `native-image` compiler.

melix:
> Couldn't we have assemble that delegates to a `nativeCompile` task?

Yes, we could, and in theory, we _should_ bound the `assemble` task to whatever `nativeCompile` task exists. I'm not super happy with `assemble` either for that reason.

> Invoking native image compiler is really a compilation, and retaining this term work make more sense from my point of view. The native build tools declaration and the native prefix makes it clear it is not about javac but native-image compiler.

I agree with the "compilation" aspect (compiles to native code). However, `nativeCompile` doesn't say much about the underlying native: is it Graal? Is it Gradle native (C++)? We could think about `nativeJavaCompile` and `nativeJavaTestCompile`, but that would IMO be confusing because we're not compiling Java sources to native, we're compiling _JVM bytecode classes_ to native. Therefore, Kotlin, Scala, Groovy classes are also candidates for native compilation, so I'm not happy with the term "Java", but we're a bit abusing it in any case when we declare the `javaNative` extension. This confusion isn't specific to this plugin, tbh, there are several places in Gradle itself when we say "Java" but we should say "Java Ecosystem" (not even JVM).

So I _think_ I could be convinced with `nativeJavaCompile` and `nativeJavaTestCompile`.

docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -226,3 +246,60 @@ The generated configuration files will be found in the `${buildDir}/native/agent
 == Javadocs
 
 In addition, you can consult the link:javadocs/native-gradle-plugin/index.html[Javadocs of the plugin].
+
+[[changelog]]
+== Changelog
+
+=== Release 0.9.3
```
sdeleuze, line 253: To be updated to 0.9.4 once 0.9.3 released.

sdeleuze:
> I agree with the "compilation" aspect (compiles to native code). However, nativeCompile doesn't say much about the underlying native: is it Graal? Is it Gradle native (C++)? We could think about nativeJavaCompile and nativeJavaTestCompile, but that would IMO be confusing because we're not compiling Java sources to native, we're compiling JVM bytecode classes to native. Therefore, Kotlin, Scala, Groovy classes are also candidates for native compilation, so I'm not happy with the term "Java", but we're a bit abusing it in any case when we declare the javaNative extension. This confusion isn't specific to this plugin, tbh, there are several places in Gradle itself when we say "Java" but we should say "Java Ecosystem" (not even JVM).
>
> So I think I could be convinced with nativeJavaCompile and nativeJavaTestCompile.

The point you raised about Scala, Kotlin and other languages is a concern of mine as well, and I was not able to convince myself  those developers will be happy to use `nativeJavaCompile` or `nativeJavaTestCompile`. + the fact that the reverse order compared to the DSL block will be confusing.

That's why I ended up with `nativeCompile`, I fully realize it is not perfect semantically speaking, but in the context of a project where we use this plugin, I think that's ok to use this kind of shortcut. At least that's from my POV the most reasonable trade off i can think about.

melix:
> the fact that the reverse order compared to the DSL block will be confusing

Then, considering that the DSL block is named `javaNative`, what about `javaNativeCompile` and `javaNativeTestCompile`?

vjovanov:
I think `java` is offputting to people that don't write Java. How about we change java to `jvm` or `j`? This would shorten all the commands and make them both explicit and concise: 
```
jvmNative {
    registerTestSourceSet(sourceSets.functionalTest)
    images {
        // configure the built-in images here
        ...
        functionalTest {
            // configure the native image for the source set previously registered
        }
    }
}
```
And then I would also use `jvmNativeCompile` for consistency with the block.

melix:
This is ready for review: renames have been made according to our latest conversations.

common/docs/ANNOUNCEMENT.md
graemerocher: Looks like fallback is already false by default so we are good here

docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -173,58 +185,65 @@ The following configuration options are available for building images:
...
+      // Runtime options
+      runtimeArgs.add("--help") // Passes '--help' to built image, during "nativeRun" task
+
+      // Development options
+      agent.set(true) // Enables the reflection agent. Can be also set on command line using '-Pagent'
+
+      useFatJar.set(true) // Instead of passing each jar individually, builds a fat jar
+    }
+  }
 }
```
graemerocher, line 247: Do we need an example somewhere on how to configure an additional native image build other than `main`?

graemerocher (approved):

docs/src/docs/asciidoc/gradle-plugin.adoc
melix: I think this should be done in a separate PR with an API allowing to register a new image from a different source set.

PR #103, opened 2021-07-13
Replace container based testing with "isolated" Maven testing

### LINKED ISSUE #102: Remove Docker dependency from Maven testing
ATM we are using Docker to create layered images in order to have a reproducible builds for testing the Maven plugin.
See the rationale behind this decision [here](https://github.com/graalvm/native-build-tools/pull/87/commits/d96237b8242f1c1054c76ec4ba925590c873e940), and related discussion [here](https://github.com/graalvm/native-build-tools/pull/87#discussion_r663271744).

This is creating an overhead by fetching an entire docker image with a duplicate GraalVM installation, a loads of additional logic related to Gradle <-> Docker interaction, but the biggest problem is that it limits Maven testing to Linux (since guest Docker image will always be Linux based).

I propose that whole Docker related part gets removed and replaced with Groovy / Kotlin logic which would invoke `mvnw` to download separate maven installation, custom `settings.xml` and environment variables which would change `.m2` location, and a "folder-juggling" mechanism to simulate repository caching between each run. 

I'll work on PoC for this and link it to this issue as soon as it's ready.

melix:
This commit reworks functional testing of the Maven plugin by executing
Maven in a sandbox, isolated "as much as possible". We cannot, unfortunately,
make sure that it's fully isolated but the following steps were taken:

- the libraries published by this repository are published in the "common"
repository
- the external dependencies required for functional testing are seeded into
another, shared repository
- when tests are executed, a local repository is set so that dependencies
are copied into that repository rather than using the local m2 repository

This should make functional testing faster and fix the problem that the
Maven plugin couldn't be tested on something else than Linux.

Fixes #102


lazar-mitrovic:
This looks great, thanks for looking into this!
I would potentially set M2_HOME environment variable in order to prevent embedded maven from using default `.m2` location.

From maven wrapper docs:
```bash
# Optional ENV vars
# -----------------
#   M2_HOME - location of maven2's installed home dir
```

native-maven-plugin/src/seeding-build/src/test/java/org/graalvm/demo/CalculatorTest.java
```diff
@@ -0,0 +1,38 @@
+package org.graalvm.demo;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
lazar-mitrovic, line 3: Do we need these files in this project? 
I guess we only need to specify dependencies and have a basic `HelloWorld` in order for it to pull them?

gradle/libs.versions.toml
```diff
@@ -1,9 +1,9 @@
 [versions]
 # Project versions
-junitPlatformNative = "0.9.2"
-nativeGradlePlugin = "0.9.2"
-nativeMavenPlugin = "0.9.2"
-utils = "0.9.2"
+junitPlatformNative = "0.9.3-SNAPSHOT"
+nativeGradlePlugin = "0.9.3-SNAPSHOT"
+nativeMavenPlugin = "0.9.3-SNAPSHOT"
+utils = "0.9.3-SNAPSHOT"
```
lazar-mitrovic, line 6: If I saw correctly, you haven't run `updateSamples` task? 
We might want to write a comment here that reminds developer that samples should be kept in sync with this file

melix: Actually `master` should probably _never_ use a fixed version but always be `SNAPSHOT`.

native-maven-plugin/src/seeding-build/src/test/java/org/graalvm/demo/CalculatorTest.java
lazar-mitrovic: I understand the purpose of the `seeding-build` project, but I don't get why we need to fill that project with code. 
AFAIK we could just have a `HelloWorld.java` and specify that it depends on all of the needed dependencies.

gradle/libs.versions.toml
melix: fixed

melix:
> I would potentially set M2_HOME environment variable in order to prevent embedded maven from using default .m2 location.

done

native-maven-plugin/src/seeding-build/src/test/java/org/graalvm/demo/CalculatorTest.java
melix: Yes, we could also do that, but if you don't have tests, then the Maven surefire plugin wouldn't be downloaded, for example. So it's actually easier to use a "fake" project with real code so that all plugins we use in the similar samples are downloaded. It's easier to reason about IMHO.

gradle/libs.versions.toml
lazar-mitrovic: We need to have one commit that uses a fixed version in order to link to it when we publish a release.
ATM my process is that I create that one release commit, and then immediately push another one with a version bump. I haven't done it now since we have a release delay.
 
We _could_ have that commit on a separate branch. but it seems weird and confusing. I'll push version bumping commit today anyway.

lazar-mitrovic: Oh you already did it, great!

melix: Yes, there needs to be one _commit_, but ideally this should only be pushed on release.

melix:
@lazar-mitrovic please merge at your convenience, I guess you're waiting for 0.9.2 release to be officially out.

Issue #105, opened 2021-07-13
Ensure that samples work out-of-the-box


lazar-mitrovic:
Preferably we would want to support users copying project from the `samples` directory just building upon that.

For this we need to do two things:
* ensure that `mavenCentral` is included in the `pluginManagement` block in the Gradle plugin
    We already specify this [in the readme](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/README.md). 
* pin plugin version in the Gradle plugin to the repository one.

PR #113, opened 2021-07-30
Set mainClass as optional, fix samples to be usable standalone

### LINKED ISSUE #108: Regression: mainClass property should not be mandatory
This 0.9.2 regression (maybe coming from [this commit](https://github.com/graalvm/native-build-tools/commit/633c34cad8e169b832c8ccfb729ccefaed1a5535#diff-4b478e87d3e244b734743d63e08f0640211913a7973e1bb5073a7018297a2868) with a missing `@Optional`) breaks most Spring Native Gradle projects, since we set the main class via `-H:Class=` in `native-image.properties` and this is not consistent with Maven.

lazar-mitrovic:
Before this commit if mainClass was missing, Gradle plugin would abort the build.
This is wrong since mainClass isn't required for building shared libraries (also,
mainClass can be defined using properties file). This change fixes #108

Second change that is introduced in this commit fixes issue #105


build.gradle.kts
```diff
@@ -88,6 +88,7 @@ tasks.register<Zip>("releaseZip") {
 
 tasks.register<org.graalvm.build.samples.SamplesUpdateTask>("updateSamples") {
     inputDirectory.set(layout.projectDirectory.dir("samples"))
+    versions.put("nativeGradlePluginVersion", libs.versions.nativeGradlePlugin.get())
```
sdeleuze, line 91: Is there something preventing to use dot separators instead of camel case notation? That would be better for consistency.

samples/java-application-with-reflection/settings.gradle
```diff
@@ -39,4 +39,12 @@
  * SOFTWARE.
  */
 
+pluginManagement {
+    repositories {
+        mavenLocal()
```
melix, line 44: :x: Shouldn't use `mavenLocal`

samples/java-application-with-reflection/build.gradle
```diff
@@ -41,10 +41,11 @@
 
 plugins {
     id 'application'
-    id 'org.graalvm.buildtools.native'
+    id('org.graalvm.buildtools.native') version "${nativeGradlePluginVersion}"
```
melix, line 44: :x: instead of using a version here, the version should be declared in `settings.gradle` plugin management block, which would let you use dotted versions.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/dsl/NativeImageOptions.java
```diff
@@ -91,6 +92,7 @@
      * @return mainClass The main class.
      */
     @Input
+    @Optional
```
melix, line 95: :x: This change alone isn't sufficient. The code needs to be updated to remove the default value, and test if a value was set instead.

melix (changes requested):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/dsl/NativeImageOptions.java
lazar-mitrovic: Default value for `mainClass` property is set only when related property in the `application` block is set, as seen here:
https://github.com/graalvm/native-build-tools/blob/1456c696a3782ff2dfaf4c76fb225095aa3b8b0a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java#L132-L134

In other cases this value would be omitted, as expected.
https://github.com/graalvm/native-build-tools/blob/1456c696a3782ff2dfaf4c76fb225095aa3b8b0a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java#L134-L136

IMO this will handle all of the use cases.

build.gradle.kts
lazar-mitrovic: Dot separated properties cannot be used for versions in `plugins` group in `build.gradle`.
However as [Cédric pointed out](https://github.com/graalvm/native-build-tools/pull/113#discussion_r679803753) this is possible in `settings.gradle`, so I'll use that instead.

samples/java-application-with-reflection/settings.gradle
lazar-mitrovic: I found it useful when I have published the gradle plugin itself to the `mavenLocal` (the same way I would use those samples with maven plugin). I moved `mavenLocal` under the `mavenCentral`. Is there some specific reason why we would want to avoid this?

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -93,7 +93,7 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractFunctionalTest {
     @Unroll("agent property takes precedence on Gradle #version with JUnit Platform #junitVersion")
     def "agent property takes precedence"() {
         gradleVersion = version
-
```
sdeleuze, line 96: If you can rollback for cleaner diff that would be great.

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/KotlinApplicationWithTestsFunctionalTest.groovy
```diff
@@ -50,7 +50,7 @@ class KotlinApplicationWithTestsFunctionalTest extends AbstractFunctionalTest {
     @Unroll("can execute Kotlin tests in a native image directly on Gradle #version with JUnit Platform #junitVersion")
     def "can execute Kotlin tests in a native image directly"() {
         gradleVersion = version
-
```
sdeleuze, line 53: If you can rollback for cleaner diff that would be great.

sdeleuze (changes requested):
Works for me with my additional comments on formatting and after clarifying with @melix the `mavenLocal()`, I admit I added it as well locally for my `gradle publishToMavenLocal` but I guess he has another workflow to recommend.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/dsl/NativeImageOptions.java
melix: :+1:

samples/java-application-with-reflection/settings.gradle
melix: There are many reasons to avoid `mavenLocal()`, main one being non reproducibility. There are 2 cases we want to handle:

1. a user building this project, in which case the tests will automatically inject the required artifacts via composite builds, so `mavenLocal()` isn't required
2. a user willing to try a sample, in which case they should use a release (or pre-release) version, in which case `mavenLocal()` is also redundant, and worse, could introduce non-determinism

IMO it's a bad practice to encourage using `mavenLocal()` as people tend to copy and paste samples, and that `mavenLocal()` would be a bad thing to do. Note that this change is also unrelated to this PR, so I would leave it out.

lazar-mitrovic: I removed `mavenLocal` from this PR so we should be OK.

melix (approved):

PR #115, opened 2021-08-09
Default to the current Java version

### LINKED ISSUE #109: Regression: Gradle plugin requires GRAALVM_HOME to be set
I think this is a regression we previously saw and fix with @melix but it seems to be back with 0.9.2. When using https://sdkman.io to install and use GraalVM, `GRAALVM_HOME` env variable is not set and when using the Gradle plugin and I run `./gradlew clean nativeBuild` I get the following error:

```

* What went wrong:
Execution failed for task ':nativeBuild'.
> Expected to find native-image executable but it wasn't found. Make sure to declare the GRAALVM_HOME environment variable or install GraalVM with native-image in a standard location recognized by Gradle Java toolchain support
```

The same use case works correctly with 0.9.1. If I set `GRAALVM_HOME` with 0.9.2 it works.

melix:
This is a workaround for Gradle < 7 not being able to set the vendor
requirement, which causes hard to diagnose errors.

Fixes #109

### DIFF
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeImageOptionsTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeImageOptionsTest.groovy
@@ -1,0 +1,44 @@
+package org.graalvm.buildtools.gradle
+
+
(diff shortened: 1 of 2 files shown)


lazar-mitrovic (approved):
LGTM!

PR #118, opened 2021-08-10
Download native-image executable using gu tool

lazar-mitrovic:


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -154,22 +156,40 @@ public void exec() {
...
-                    getService().get();
-                    spec.setExecutable(executable);
+
+        try {
+            if (!executablePath.exists()) {
+                logger.log("Native Image executable wasn't found. We will now try to download it. ");
+                File graalVmHomeGuess = executablePath.getParentFile();
+
+                if (!graalVmHomeGuess.toPath().resolve(GU_EXE).toFile().exists()) {
+                    throw new GradleException("'" + GU_EXE + "' tool wasn't found. Is this JDK GraalVM distribution?");
```
melix, line 166: I would rephrase to something like "tool wasn't found. This probably means that JDK at <path> isn't a GraalVM distribution."

PR #120, opened 2021-08-11
Fix documentation CI script

lazar-mitrovic:
Since `eskatos/gradle-command-action@v1` is disabled due to security policy, lets invoke Gradle directly.


melix (approved):
:+1: In the meantime let's try to get this action approved, since it would improve build times on CI.

PR #121, opened 2021-08-12
Add a `sharedLibrary` configuration option for Gradle plugin

### LINKED ISSUE #119: Feature Request: Add support for generating native library
Currently this tool only allows to generate native images and not native library. We are currently using `com.palantir.graal` plugin that allows this feature and I would like to replace it with the official support.

lazar-mitrovic:
This commit adds `sharedLibrary` configuration option to the Gradle plugin as it was a requested feature (closes #119).
`java-library` test sample is also included as well as corresponding functional test.

### DIFF
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaLibraryFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaLibraryFunctionalTest.groovy
@@ -1,0 +1,83 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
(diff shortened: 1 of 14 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -134,6 +134,8 @@ private void configureJavaProject(Project project, Provider<NativeImageService>
                 project.getExtensions().findByType(JavaApplication.class).getMainClass()
         ));
 
+        project.getPlugins().withId("java-library", p -> buildExtension.getSharedLibrary().convention(true));
```
melix, line 137: You can set the convention to `false` in the `NativeImageOptions` constructor, which avoids the `@Optional` and `getOrElse` when you consume the value.

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaLibraryFunctionalTest.groovy
```diff
@@ -0,0 +1,68 @@
...
+        when:
+        run 'nativeBuild'
+
+        then:
+        tasks {
+            succeeded ':jar', ':nativeBuild'
+            doesNotContain ':build'
+        }
+
+        outputContains "-H:+SharedLibrary"
```
melix, line 63: Would be great to check that an actual shared library file is produced.

melix (approved):
I left a couple comments which would be great to fix, no blockers. And since you wrote a `pom.xml` file for the sample, would actually be good to add a test for Maven too.

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaLibraryFunctionalTest.groovy
lazar-mitrovic: Fixed

PR #126, opened 2021-08-18
Use a classpath jar under Windows

### LINKED ISSUE #85: CreateProcess error=206, The filename or extension is too long
Using 0.9.1, I received the following error on trying to build the native image:

[ERROR] Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.9.1:build (build-native) on project SBProject: Building image with C:\graalvm\bin\native-image.cmd failed: Cannot run program "C:\graalvm\bin\native-image.cmd" (in directory "D:\projects\SBProject\target"): CreateProcess error=206, The filename or extension is too long

melix:
This commit introduces a classpath jar, a jar which contains no
classes but only a Class-Path manifest entry, so that the classpath
to `native-image` is passed this way instead of on the CLI.

This would fix the issue of long classpath entries on Windows.

Fixes #85

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -159,6 +159,8 @@ nativeBuild {
 
   // Development options
   agent = true // Enables the reflection agent. Can be also set on command line using '-Pagent'
+
+  useFatJar = true // Instead of passing each jar individually, builds a fat jar
 }
 ```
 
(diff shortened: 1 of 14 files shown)


melix:
Unfortunately, I don't think this will fix the reported issue: we're apparently moving the problem one level down. While it fixes the problem of passing a long classpath to `native-image`, it seems that `native-image` itself invokes a `java` process and passes an expanded `-imagecp` argument which contains the entries found in the classpath jar. As a consequence, fixing this would also require a fix to `native-image` itself.

melix:
Following discussion with @lazar-mitrovic , the implementation was changed to use a _fat jar_ under Windows. This mode can be explicitly disabled.

lazar-mitrovic (approved):

PR #128, opened 2021-08-19
Add publication of snapshots to GitHub

melix:
This commit introduces publication of snapshot artifacts to
a branch, so that we can use GitHub as a repository.

### DIFF
--- a/.github/workflows/deploy-snapshots.yml
+++ b/.github/workflows/deploy-snapshots.yml
@@ -17,6 +17,20 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 1
+          ssh-key: "${{ secrets.SSH_PRIVATE_KEY }}"
+      - name: Setup git
+        env:
+          GIT_SSH_COMMAND: "ssh -o StrictHostKeyChecking=no"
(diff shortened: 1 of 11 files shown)


.github/workflows/deploy-snapshots.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Get GraalVM Nightly
+        run: |
+          source common/scripts/downloadGraalVM.sh
+          echo "$GRAALVM_HOME/bin" >> $GITHUB_PATH
+          echo "JAVA_HOME=$JAVA_HOME" >> $GITHUB_ENV
+          echo "GRAALVM_HOME=$GRAALVM_HOME" >> $GITHUB_ENV
+      - name: Run gradle
```
lazar-mitrovic, line 26: I believe that after adding:
https://github.com/graalvm/native-build-tools/blob/d51b4bed97351a2fabbc04de074cd321712dc6fd/.github/workflows/documentation.yml#L17-L29

GIt credentials should work out-of-the-box

melix: Thanks, added!

PR #137, opened 2021-09-28
Allow disabling testing support

### LINKED ISSUE #133: Pulling in JUnit Dependencies
Including the following line in the plugins section:
   id 'org.graalvm.buildtools.native' version "0.9.5"
results in dependencies on junit to get pulled into my project, that I would prefer not to be there, as I use TestNG.

melix:
This commit adds an option to disable testing support in the Gradle
plugin. This is useful whenever the test framework that the application
or library is using is not JUnit Platform (e.g TestNG) or in cases the
user doesn't want to run tests in native mode.

Fixes #133

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -329,7 +329,7 @@ The generated configuration files will be found in the `${buildDir}/native/agent
 There are cases where you might want to disable test support:
 
 - you don't actually want to run your tests in native mode
-- your library or application uses another test framework than JUnit Platform (currently this plugin only supports testing with JUnit Platform)
+- your library or application uses another test framework than JUnit 5 (currently this plugin only supports testing with JUnit Platform 5.8+)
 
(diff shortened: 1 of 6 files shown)


docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -324,6 +324,30 @@ Same can be achieved by setting corresponding DSL option, althought this isn't r
 
 The generated configuration files will be found in the `${buildDir}/native/agent-output/${taskName}` directory, for example, `build/native/agent-output/run`.
 
+=== Disabling test support
+
+There are cases where you might want to disable test support:
+
+- you don't actually want to run your tests in native mode
+- your library or application uses another test framework than JUnit Platform (currently this plugin only supports testing with JUnit Platform)
```
graemerocher, line 332: clarify JUnit 5 since we have tickets as well asking for JUnit 3 support

melix: done

graemerocher (approved):

PR #157, opened 2021-11-07
Ensure native test image exit code is in range 0-255

### LINKED ISSUE #154: NativeImageJUnitLauncher returns exit code that can cause issues for shell scripts
## Overview

The exit code is currently the value of the number of failures in the test build.

https://github.com/graalvm/native-build-tools/blob/15bea361faf74d2f36655aef688995be17e59918/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java#L138

For the `ConsoleLauncher` in the JUnit Platform (JUnit 5), we originally did the same thing, but users reported issues with their shell scripts that reacted to the exit code if the value was too big (i.e., more than 255 failed tests).

It turns out that [the value should be between 0 and 255](https://www.shellscript.sh/exitcodes.html).

Consequently, the JUnit team revised its exit code handling to limit the values to -1 through 2. See [ConsoleLauncherExecutionResult](https://github.com/junit-team/junit5/blob/f4ef4b9e607579dd0000d4bfcfe6a34b39823b53/junit-platform-console/src/main/java/org/junit/platform/console/ConsoleLauncherExecutionResult.java) for details.

## Proposal

Change:

```java
System.exit((int) failedCount);
```

To:

```java
System.exit(failedCount > 0 ? 1 : 0);
```

## Related Issues

- #153

## Deliverables

- [x] Ensure that the exit code returned by the `NativeImageJUnitLauncher` falls in the range of 0 to 255.

sbrannen:
Prior to this commit, the exit code for the native test image was based
on the number of test failures. For a small number of failures, this is
not an issue; however, if there is a large number of failures the exit
code can problematic for shell scripts processing the exit code.

In order to make the exit code more user friendly for shell scripts,
this commit returns an exit code of 1 if there are any test failures.

The actual number of failed tests and containers can be viewed in the
summary printed to STD_OUT, and the full details can be obtained in
the generated XML test reports.

Fixes #154

### DIFF
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
@@ -135,6 +135,6 @@ public class NativeImageJUnitLauncher {
         }
 
         long failedCount = summary.getTestsFailedCount() + summary.getTestsAbortedCount();
-        System.exit((int) failedCount);
+        System.exit(failedCount > 0 ? 1 : 0);
     }
 }


sbrannen:
Please do not merge this PR until @aclement has given his approval.

aclement:
looks good to me

PR #165, opened 2021-11-17
Add an option to disable toolchain support

### LINKED ISSUE #164: Make it possible to completely override toolchain support
Gradle's toolchain support currently causes a couple of issues:

- it's not possible to select a specific _version_ of GraalVM, for example to use 21.3 if 20.2 is also installed
- it _cannot_ detect GraalVM Enterprise, which completely breaks this plugin

In effect, the plugin makes use of the _vendor_ spec to match GraalVM. This works fine for GraalVM Community Edition, but the string for GraalVM Enterprise is actually "Oracle Corporation", which makes it impossible to differentiate from other Oracle JDKs.

In practice, it's actually the _implementation_ string which makes the difference, but there's no way to ask Gradle for a specific _implementation_ in the current Java toolchain spec.

As a consequence, we should add a way to bypass toolchain detection. Because it works in some cases, it would be a problem to always use the `GRAALVM_HOME` environment variable to override whatever is declared, so we suggest to add support for a boolean flag on the native image tasks which, when set, would override detection.

melix:
Fixes #164

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -90,8 +90,9 @@ Follow the <<graalvm-setup.adoc#,following instructions>> to install it properly
 
 Eventually, you also have the options to:
 
-1. Run Gradle itself with a GraalVM SDK
-2. Setup a `GRAALVM_HOME` environment variable pointing to your GraalVM installation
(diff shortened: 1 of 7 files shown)


graemerocher (approved):

sdeleuze:
Unless I missed something, based on my test it does not work, see https://github.com/graalvm/native-build-tools/issues/183.

Issue #162, opened 2021-11-11
Support configuration of custom agent options when using the agent


sbrannen:
## Overview

The Maven and Gradle plugins currently configure the options for the agent as follows.

https://github.com/graalvm/native-build-tools/blob/cd0e25a28e347dfcfae39c3c7f744a1a4fa680e0/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeExtension.java#L78-L82

https://github.com/graalvm/native-build-tools/blob/cd0e25a28e347dfcfae39c3c7f744a1a4fa680e0/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/AgentCommandLineProvider.java#L73-L77

It is not currently possible to configure [Class-based Filters](https://www.graalvm.org/reference-manual/native-image/Agent/#caller-based-filters) or [Access Filters](https://www.graalvm.org/reference-manual/native-image/Agent/#access-filters). This makes it impossible to use Native Build Tools' support for the agent in Spring Native's `commandlinerunner-agent` sample, since the `access-filter-file` option cannot be set.

https://github.com/spring-projects-experimental/spring-native/blob/7eef3c2447656c637c7b7a67f492b70633c1e032/samples/commandlinerunner-agent/pom.xml#L80

In addition, the Gradle plugin sets `-Dorg.graalvm.nativeimage.imagecode=agent`, but the Maven plugin does not.

## Related Issues

- https://github.com/spring-projects-experimental/spring-native/issues/1247

## Deliverables

- [x] Make it possible to configure agent options when using the agent with Maven. These should be configurable for the main application and tests.
- [x] Document Maven support for agent options in the user guide.
- [ ] Make it possible to configure agent options when using the agent with Gradle. These should be configurable for the main application and tests.
- [ ] Document Gradle support for agent options in the user guide.

sbrannen:
I also noticed that `experimental-class-loader-support` is always enabled, but perhaps some users do not want that. Or perhaps they want the option to disable it.

In light of that, one solution to this issue would be to allow the user to configure all options supplied to `-agentlib:native-image-agent` except the `config-output-dir` which is managed by the plugin.

If we go that route, the user could specify `experimental-class-loader-support`, `caller-filter-file`, `access-filter-file`, etc. as necessary.

sbrannen:
> - [ ] Set `-Dorg.graalvm.nativeimage.imagecode=agent` in the Maven plugin.

[...]

PR #167, opened 2021-11-22
Move test code to correct files

sbrannen:
Prior to this commit, the application and calculator tests were in
the wrong source files.

PR #168, opened 2021-11-22
Make it possible to configure agent options in Maven plugin

sbrannen:
Prior to this commit, users could not supply custom options to the JVM agent. In addition, the `experimental-class-loader-support` option was always supplied, which may not be desired by all users.

This commit introduces support for configuring the agent by allowing users to specify multiple sets of named `agentOptions` in the `configuration` block for the native-maven-plugin.

The following example configures two sets of agent options, one for execution of the application (exec) and one for tests (test).

```xml
<configuration>
  <agentOptions name="exec">
    <agentOption>experimental-class-loader-support</agentOption>
  </agentOptions>
  <agentOptions name="test">
    <agentOption>experimental-class-loader-support</agentOption>
    <agentOption>access-filter-file=${basedir}/src/test/resources/access-filter.json</agentOption>
  </agentOptions>
</configuration>
```

In addition, the `experimental-class-loader-support` option is no longer configured automatically by the native-maven-plugin.

See #162

### DIFF
--- a/samples/java-application-with-reflection/pom.xml
+++ b/samples/java-application-with-reflection/pom.xml
@@ -136,12 +136,23 @@
                                 <phase>package</phase>
                             </execution>
                         </executions>
+                        <!-- tag::native-plugin-agent-options[] -->
                         <configuration>
+                            <!-- end::native-plugin-agent-options[] -->
                             <imageName>${imageName}</imageName>
                             <buildArgs>
                                 <buildArg>--no-fallback</buildArg>
                             </buildArgs>
+                            <!-- tag::native-plugin-agent-options[] -->
+                            <agentOptions name="exec">
+                                <agentOption>experimental-class-loader-support</agentOption>
+                            </agentOptions>
+                            <agentOptions name="test">
+                                <agentOption>experimental-class-loader-support</agentOption>
+                                <agentOption>access-filter-file=${basedir}/src/test/resources/access-filter.json</agentOption>
+                            </agentOptions>
                         </configuration>
+                        <!-- end::native-plugin-agent-options[] -->
                     </plugin>
                 </plugins>
             </build>
--- a/samples/java-application-with-reflection/pom.xml
+++ b/samples/java-application-with-reflection/pom.xml
@@ -144,13 +144,18 @@
                                 <buildArg>--no-fallback</buildArg>
                             </buildArgs>
                             <!-- tag::native-plugin-agent-options[] -->
-                            <agentOptions name="exec">
-                                <agentOption>experimental-class-loader-support</agentOption>
-                            </agentOptions>
-                            <agentOptions name="test">
-                                <agentOption>experimental-class-loader-support</agentOption>
-                                <agentOption>access-filter-file=${basedir}/src/test/resources/access-filter.json</agentOption>
-                            </agentOptions>
+                            <agent>
+                                <!-- end::native-plugin-agent-options[] -->
+                                <enabled>true</enabled>
+                                <!-- tag::native-plugin-agent-options[] -->
+                                <options name="main">
+                                    <option>experimental-class-loader-support</option>
+                                </options>
+                                <options name="test">
+                                    <option>experimental-class-loader-support</option>
+                                    <option>access-filter-file=${basedir}/src/test/resources/access-filter.json</option>
+                                </options>
+                            </agent>
                         </configuration>
                         <!-- end::native-plugin-agent-options[] -->
                     </plugin>
--- a/samples/java-application-with-reflection/pom.xml
+++ b/samples/java-application-with-reflection/pom.xml
@@ -148,15 +148,18 @@
                                 <!-- end::native-plugin-agent-options[] -->
                                 <enabled>true</enabled>
                                 <!-- tag::native-plugin-agent-options[] -->
-                                <options><!-- unnamed, shared options -->
+                                <options><!-- shared options -->
                                     <option>experimental-class-loader-support</option>
                                 </options>
                                 <options name="main"><!-- main options -->
-                                    <option>config-write-period-secs=30</option>
-                                    <option>config-write-initial-delay-secs=5</option>
+                                    <option>access-filter-file=${basedir}/src/main/resources/access-filter.json</option>
                                 </options>
                                 <options name="test"><!-- test options -->
                                     <option>access-filter-file=${basedir}/src/test/resources/access-filter.json</option>
+                                </options>
+                                <options name="periodic-config"><!-- periodic-config options -->
+                                    <option>config-write-period-secs=30</option>
+                                    <option>config-write-initial-delay-secs=5</option>
                                 </options>
                             </agent>
                         </configuration>
--- a/samples/java-application-with-reflection/pom.xml
+++ b/samples/java-application-with-reflection/pom.xml
@@ -148,11 +148,14 @@
                                 <!-- end::native-plugin-agent-options[] -->
                                 <enabled>true</enabled>
                                 <!-- tag::native-plugin-agent-options[] -->
-                                <options name="main">
+                                <options><!-- unnamed, shared options -->
                                     <option>experimental-class-loader-support</option>
                                 </options>
-                                <options name="test">
-                                    <option>experimental-class-loader-support</option>
+                                <options name="main"><!-- main options -->
+                                    <option>config-write-period-secs=30</option>
+                                    <option>config-write-initial-delay-secs=5</option>
+                                </options>
+                                <options name="test"><!-- test options -->
                                     <option>access-filter-file=${basedir}/src/test/resources/access-filter.json</option>
                                 </options>
                             </agent>
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -95,8 +95,10 @@ You can use development versions of the plugin by adding our snapshot repository
 [[configuration-options]]
 === Configuration options
 
-If you use Native Image Maven plugin, it will pick up all the configuration for your application stored below the `META-INF/native-image/` resource location, as described in https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/[Native Image Build Configuration].
-It is also possible to customize the plugin within a
+If you use Native Image Maven plugin, it will pick up all the configuration for your
+application stored below the `META-INF/native-image/` resource location, as described in
+https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/[Native Image
+Build Configuration]. It is also possible to customize the plugin within a
 `<configuration>` node. The following configuration options are available.
 
 `<mainClass>`::
@@ -116,6 +118,9 @@ It is also possible to customize the plugin within a
 `<skipNativeBuild>`::
    To skip generation of the native image, supply
    `<skipNativeBuild>true</skipNativeBuild>` in the configuration of the plugin.
+`<agent>`::
+   Configuration of the <<agent-support, native agent>>. See <<agent-support-enabling>>
+   and <<agent-support-configuring-options>> for details.
 
 For example, to build a native image named `executable-name` that uses
 `org.example.ClassName` as its main class with assertions enabled, add the following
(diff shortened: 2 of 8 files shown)


sbrannen:
FYI: This PR currently (accidentally) includes the changes proposed in #167. So those can be ignored when reviewing this PR.

sbrannen:
Based on feedback from the team, I've pushed a [commit](https://github.com/graalvm/native-build-tools/pull/168/commits/773765e3627d7f031bdf0574c5990c963349ab5e) that introduces support for named groups of configuration blocks for agent options.

```xml
<configuration>
  <agentOptions>
    <config name="default">
      <agentOption>experimental-class-loader-support</agentOption>
    </config>
    <config name="test">
      <agentOption>experimental-class-loader-support</agentOption>
      <agentOption>access-filter-file=${basedir}/src/test/resources/access-filter.json</agentOption>
    </config>
  </agentOptions>
</configuration>
```

sbrannen:
I have pushed an additional [commit](https://github.com/graalvm/native-build-tools/pull/168/commits/c237f332c7bbc883101d6bb15394650e0b4f688e) that further simplifies the configuration from the user's perspective, resulting in a Maven command line similar to:

```shell
mvn -Pnative -Dagent=true -DagentOptions=test test
```

And configuration within `pom.xml` similar to:

```xml
<configuration>
    <agentOptions name="default">
        <agentOption>experimental-class-loader-support</agentOption>
    </agentOptions>
    <agentOptions name="test">
        <agentOption>experimental-class-loader-support</agentOption>
        <agentOption>access-filter-file=${basedir}/src/test/resources/access-filter.json</agentOption>
    </agentOptions>
</configuration>
```

sbrannen:
This PR is no longer a _draft_.

I have finalized the feature, tested it, documented it, squashed all commits, rebased on `master`, and force pushed the changes.

sbrannen:
Please refrain from merging this PR until @aclement has approved it.

melix (approved):
LGTM!

sdeleuze:
Thanks for working on this. 2 change proposal I deeply care about:

**1**  I like the powerfullness of being able to specify multiple `<agentOptions>`, but I am concerned by the ceremony required to use this feature (2 options need to be used!), and think we should make it easier to use for simple use case. After discussing with @sbrannen, our proposal is the following: it should be possible to specify only `mvn -Pnative -Dagent=true test` to enable a single unnamed `<agentOptions>` element, for example:
```
<configuration>
    <agentOptions>
        // ...
    </agentOptions>
</configuration>
```
When there is more than 1 `<agentOptions>` element, names should be required required and `-DagentOptions=foo` should be used (I guess we need to support `-DagentOptions=foo -DagentOptions=bar` as well).

**2** While I agree that the majority of use cases will use `-Dagent=true` to enable the agent configuration, I think we should support a related option in `pom.xml` to enable the agent options by default when the user want to do that (some project will require that to run and that's not OK from my POV to force them to type some custom mvn command).

```
<configuration>
    <agentOptions enabled=true> // By default enabled=false
        // ...
    </agentOptions>
</configuration>
```
That will be applied with using `mvn -Pnative test`.

sbrannen:
With regard to allowing the user to enable the agent by default, I would prefer an explicit XML element to configure (analogous to the `agent = true` support in the Gradle DSL.

```
<configuration>
    <agent>true</agent> // By default enabled=false
    <agentOptions>
        // ...
    </agentOptions>
</configuration>
```

We could also go with something like `<agentEnabled>true</agentEnabled>`.

In summary, having the "enable agent" flag standalone allows users to enable the agent within the build file even if they are not defining custom agent options.

sbrannen:
In order to better align with the support in the Gradle DSL, I propose that the unnamed agent options always be applied (first) with named agent options applied after that (if the named set of agent options is selected via `-DagentOptions=<NAME>`).

```xml
<configuration>
    <agentOptions> <!-- always applied first -->
        <!-- ... -->
    </agentOptions>
    <agentOptions name="test"> <!-- applied after the unnamed options -->
        <!-- ... -->
    </agentOptions>
</configuration>
```

In the above example, the `test` options would _extend_ the global (unnamed) options.

I don't think we should allow the `test` options to _override_ the global options, since that would require that we parse all of the options with knowledge of the semantics of each option. For example, if both sets configure the `access-filter-file` option, we would have to detect that two options start with `access-filter-file` and let the latter override the former.

aclement:
looks good to me.

sdeleuze:
@sbrannen Does https://github.com/graalvm/native-build-tools/pull/168#issuecomment-978028540 replace what you proposed in https://github.com/graalvm/native-build-tools/pull/168#issuecomment-978017722? In other words, do you plan to not have `<agent>true</agent>` but just enable unamed `<agentOptions>` automatically?

Could you please elaborate on the consistency with the latest version of the Gradle PR (https://github.com/graalvm/native-build-tools/pull/169)?

sbrannen:
> Does [#168 (comment)](https://github.com/graalvm/native-build-tools/pull/168#issuecomment-978028540) replace what you proposed in [#168 (comment)](https://github.com/graalvm/native-build-tools/pull/168#issuecomment-978017722)?

No. It's complementary. 

> In other words, do you plan to not have `<agent>true</agent>` but just enable unamed `<agentOptions>` automatically?
> 
> Could you please elaborate on the consistency with the latest version of the Gradle PR (#169)?

With the Gradle plugin, agent support is disabled by default, but it can be enabled via `agent=true` in the DSL or via `-Pagent` on the command line.

My proposal is for parity in the Maven plugin:

- agent support disabled by default
- agent support can be enabled in `pom.xml` via `<agent>true</true>` in the `<configuration>` for the plugin
- agent support can be enabled via `-Dagent-true` on the command line

samples/java-application-with-reflection/pom.xml
```diff
@@ -143,6 +143,7 @@
                             <buildArgs>
                                 <buildArg>--no-fallback</buildArg>
                             </buildArgs>
+                            <agentEnabled>true</agentEnabled>
```
sdeleuze, line 146: I am in favor of naming it `agent` to be consistent with Maven command line (`-Dagent=true`) and Gradle side (`agent = true`).

samples/java-application-with-reflection/pom.xml
```diff
@@ -136,12 +136,23 @@
...
                         </executions>
+                        <!-- tag::native-plugin-agent-options[] -->
                         <configuration>
+                            <!-- end::native-plugin-agent-options[] -->
                             <imageName>${imageName}</imageName>
                             <buildArgs>
                                 <buildArg>--no-fallback</buildArg>
                             </buildArgs>
+                            <!-- tag::native-plugin-agent-options[] -->
+                            <agentOptions name="exec">
```
sdeleuze, line 148: I think we miss an enabled attribute like `<agentOptions enabled=true name="exec">` here, similar to what we have on Gradle side.

sdeleuze (dismissed):
I think `<agentEnabled>` element should be renamed to `<agent>` and that we miss an `enabled` attribute in `<agentOptions>`.

samples/java-application-with-reflection/pom.xml
sbrannen: On the Gradle side `agent = true` is only remaining for legacy purposes (backwards compatibility).

The _new_ way to set that is:

```groovy
agent {
    enabled = true
}
```

So I was aligning with that: agent/enabled --> agentEnabled.

sbrannen: As an alternative, we could go with an `<agent>` container in the POM, something like this:

```xml
<configuration>
  <agent enabled="true">
    <agentOptions>
      <agentOption><!-- ... --></agentOption
    </agentOptions>
  </agent>
</configuration>
```

samples/java-application-with-reflection/pom.xml
melix: I don't think it makes sense. The `agent=true` flag in Gradle _delegates to_ `agentOptions.enabled=true`. So it's effectively the _same flag_. It's there to enable the agent. Then you have the options. The options are defined in the build file (build.gradle or pom.xml) per "profile", where the profile is explicit in Maven (`name="exec"`) and _implicit_ in Gradle (because you run the `main` native image, it will use the `main` options).

So what @sbrannen did looks good to me.

sbrannen: In https://github.com/graalvm/native-build-tools/pull/168#issuecomment-978017722, I made the following point.

> having the "enable agent" flag standalone allows users to enable the agent within the build file even if they are not defining custom agent options.

We do not want to force users to declare an `<agentOptions>` block if they are not using custom agent options.

sbrannen: This is in line with the Gradle support.

The Gradle DSL does not force users to declare agent options to enable the agent.

samples/java-application-with-reflection/pom.xml
sbrannen: > ```groovy
> agentOptions {
>     enabled = true
> }
> ```

Though I admit that is confusing.

In the Gradle DSL `agentOptions` is really a container for agent configuration.

melix: We could rename to:

```
agentConfiguration {
    enabled = true
    options.add("...")
}
```

if you think that's better. More letters to type, but maybe clearer?

sbrannen: If we really want parity with the Gradle DSL, we could even use the "arguments" terminology instead of "options.

Then we might end up with something like this:

```xml
<configuration>
  <agentOptions>
    <enabled>true</enabled>
    <args><!-- "global" agent arguments -->
      <arg><!-- ... --></arg>
    </args>
    <args name="test"><!-- named agent arguments -->
      <arg><!-- ... --></arg>
    </args>
  </agentOptions>
</configuration>
```

samples/java-application-with-reflection/pom.xml
sdeleuze: What I care about is allowing to use agent options out of the box with Maven without custom command line parameters. It seems that will be achieved via applying unamed agent options (not yet done) so I will close this conversation since that will be handled differently.

samples/java-application-with-reflection/pom.xml
sdeleuze: @melix As just discussed with you, @vjovanov and @sbrannen, let's update the Gradle side with the following idiomatic DSL (your proposal + renaming `agentConfiguration` to `agent`) for clearer semantics to what is defined in https://docs.oracle.com/javase/1.5.0/docs/guide/jvmti/jvmti.html#starting and https://www.graalvm.org/reference-manual/native-image/Agent/ :

```
agent {
    enabled = true
    options.add("...")
}
```

That's a breaking change, but that's IMO reasonable to do it since:
 -  We should not provide suboptimal DSL for backard compatiblity reasons since we are still pre 1.0 (and here we have no choice)
 - This is an advanced feature
 - Migration path is easy and will be documented in released notes

On Maven side, as discussed, let's use:
```
<agent>
    <enabled>true</enabled>
    <options name="main">
        <option>experimental-class-loader-support</option>
    </options>
    <options name="test">
        <option>experimental-class-loader-support</option>
        <option>access-filter-file=${basedir}/src/test/resources/access-filter.json</option>
    </options>
</agent>
```

sbrannen: > On Maven side, as discussed, let's use:
> 
> ```xml
> <agent>
>     <enabled>true</enabled>
>     <options name="main">
>         <option>experimental-class-loader-support</option>
>     </options>
>     <options name="test">
>         <option>experimental-class-loader-support</option>
>         <option>access-filter-file=${basedir}/src/test/resources/access-filter.json</option>
>     </options>
> </agent>
> ```

This support has been added in the last two commits.

sbrannen:
This PR was temporarily a draft while I finalized the support based on feedback and team discussions.

This PR is now ready for review again, and the individual commits tell the story of the features implemented.

@sdeleuze, @melix, and @aclement, please take another look.

sdeleuze (approved):
LGTM

melix (approved):

aclement:
looks good to me.

PR #169, opened 2021-11-24
Make it possible to configure agent options in Gradle plugin

melix:
Before this commit, the `experimental-class-loader-support` was
systematically added to the agent options and it wasn't possible
to configure the options passed to the agent.

Now, each image can configure agent options independently. As
part of this work, cleanup was done in the way agent configuration
was injected. In particular, each native image now has to explicitly
state what _task_ should be instrumented when the agent is enabled.
The plugin automatically configures the `main` image to instrument
the `run` task, while the `test` image instruments the `test` task.

The agent options are configured via a new `agentOptions` block.
The old `agent.set(true)` method is delegating to the `agentOptions`
block.

See #162

This is the Gradle plugin counterpart to #168

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -246,7 +246,7 @@ The generated configuration files will be found in the `${buildDir}/native/agent
 The native agent can be configured https://www.graalvm.org/reference-manual/native-image/Agent/[with additional options].
 This can be done using the `agentOptions` configuration block:
 
-.Disabling the fat jar creation
+.Configuring agent options for every binary
 [source, groovy, role="multi-language-sample"]
 ----
 include::../snippets/gradle/groovy/build.gradle[tags=add-agent-options]
@@ -257,6 +257,19 @@ include::../snippets/gradle/groovy/build.gradle[tags=add-agent-options]
 include::../snippets/gradle/kotlin/build.gradle.kts[tags=add-agent-options]
 ----
 
+You may also define distinct agent options for different images.
+In the following example, the agent used for instrumention for the main image has distinct options from the test ones:
+
+.Configuring agent options specifically
+[source, groovy, role="multi-language-sample"]
+----
+include::../snippets/gradle/groovy/build.gradle[tags=add-agent-options-individual]
+----
+
+[source, kotlin, role="multi-language-sample"]
+----
+include::../snippets/gradle/kotlin/build.gradle.kts[tags=add-agent-options-individual]
+----
 
 === Disabling test support
 
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -17,56 +17,55 @@ For upgrading please take a look at the <<changelog,changes section>>.
 Add following to `plugins` section of your project's `build.gradle` / `build.gradle.kts`:
 
 .Applying the plugin
-[subs="verbatim,attributes", role="multi-language-sample"]
-```groovy
+[source,groovy,subs="verbatim,attributes", role="multi-language-sample"]
+----
 plugins {
   // ...
 
   // Apply GraalVM Native Image plugin
   id 'org.graalvm.buildtools.native' version '{gradle-plugin-version}'
 }
-```
+----
 
-[subs="verbatim,attributes", role="multi-language-sample"]
-```kotlin
+[source,kotlin,subs="verbatim,attributes",role="multi-language-sample"]
+----
 plugins {
   // ...
 
   // Apply GraalVM Native Image plugin
   id("org.graalvm.buildtools.native") version "{gradle-plugin-version}"
 }
-```
+----
 
 The plugin isn't available on the https://plugins.gradle.org[Gradle Plugin Portal] yet, so you will need to declare a plugin repository in addition:
 
 Add the following to your `settings.gradle` / `settings.gradle.kts`:
 
-[role="multi-language-sample"]
 .Declaring the plugin repository
-```groovy
+[source,groovy,role="multi-language-sample"]
+----
 include::../snippets/gradle/groovy/settings.gradle[tags=plugin-management]
-```
+----
 
-[role="multi-language-sample"]
-```kotlin
+[source,kotlin,role="multi-language-sample"]
+----
 include::../snippets/gradle/kotlin/settings.gradle.kts[tags=plugin-management]
-
-```
+----
 
 [TIP]
 .Testing pre-releases
 ====
 You can use the development versions of the plugin by adding our snapshot repository instead. Pre-releases are provided for convenience, without any guarantee.
-[role="multi-language-sample"]
-```groovy
+[source, groovy, role="multi-language-sample"]
+----
 include::../snippets/gradle/groovy/settings.gradle[tags=pre-release, indent=0]
+----
 
-```
-
-[role="multi-language-sample"]
-```kotlin
+[source,kotlin,role="multi-language-sample"]
+----
 include::../snippets/gradle/kotlin/settings.gradle.kts[tags=pre-release, indent=0]
-```
+----
+
 ====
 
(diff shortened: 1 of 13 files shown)


sdeleuze:
Thanks a lot for working on this @melix.

Related to my comment on [Maven side](https://github.com/graalvm/native-build-tools/pull/168#issuecomment-977994744), I really think we should support a `enabled = true` property at `agentOptions` level to allow users to enable some by default when required (some project will require that to run). By putting the default to `enabled = false`, we send already a clear signal about the fact that agent is not the first-class way to go.

sdeleuze:
Oh but there is already `agent = true`, so maybe I would refine my proposal to just rename it to `enabled = true`.

sbrannen (changes requested):
Looks good.

I reviewed the code changes and didn't notice anything, but I'll defer to your Gradle expertise regarding the details.

I've only requested a few minor non-functional changes.

sbrannen:
> Could we support enabling a single unnamed `agentOptions` with `gradle -Pagent`?

With the proposed support in the Gradle DSL, this is already possible by applying agent options to all _binaries_ (e.g., `main`, `test`, `integrationTest`, etc.). See https://github.com/graalvm/native-build-tools/pull/169#discussion_r756206932 for details.

docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -227,19 +227,37 @@ The Gradle plugin makes it easy to generate the required configuration files by
...
+==== Configuring agent options
+
+The native agent can be configured https://www.graalvm.org/reference-manual/native-image/Agent/[with additional options].
+This can be done using the `agentOptions` configuration block:
+
+.Disabling the fat jar creation
+[source, groovy, role="multi-language-sample"]
+----
+include::../snippets/gradle/groovy/build.gradle[tags=add-agent-options]
+----
```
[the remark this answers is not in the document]
melix: done

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -142,4 +142,20 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractFunctionalTest {
...
     }
 
+    def "can configure extra options to the agent"() {
+        gradleVersion = version
+        debug = true
+        given:
+        withSample("java-application-with-reflection")
+
+        when:
+        fails 'nativeTest', '-DagentOption=will-fail'
```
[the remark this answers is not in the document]
melix: done

melix:
> Could we rename agent = true to enabled = true

Done, but I kept the top level `agent = true` as a shortcut (in case you don't need agent options).

> Could we support enabling a single unamed agentOptions with gradle -Pagent?

There is already an option to use `-Pagent`. By default it will inject no agent parameters. Then if you want the same parameters in every binary, use `binaries.configureEach`, otherwise configure the options separately as indicated in the (updated) docs.

graemerocher (approved):

sbrannen (approved):

samples/java-application-with-reflection/build.gradle
```diff
@@ -69,7 +69,10 @@ graalvmNative {
     binaries {
         test {
             verbose = true
-            agent = true
+            agentOptions {
+                enabled = true
+                args.add(providers.systemProperty("agentOptions").forUseAtConfigurationTime())
+            }
```
sbrannen, line 75: Now that I've taken a second look at this, it seems a little strange to me to call that block `agentOptions`.

Semantically, it's a container for agent configuration in general.

Also, the `args` are the actual "agent options" -- as in the options supplied to the agent.

What do you think about renaming `agentOptions` to `agent`, `agentConfiguration`, or something similar and renaming `args` to `agentOptions` or simply `options`?

Here's a proposal:

```suggestion
            agent {
                enabled = true
                options.add(providers.systemProperty("agentOptions").forUseAtConfigurationTime())
            }
```

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/AgentCommandLineProvider.java
```diff
@@ -66,13 +70,18 @@ public AgentCommandLineProvider() {
...
+    @Input
+    @Optional
+    public abstract ListProperty<String> getAgentOptions();
+
     @Override
     public Iterable<String> asArguments() {
         if (getEnabled().get()) {
             File outputDir = getOutputDirectory().getAsFile().get();
+            List<String> agentOptions = new ArrayList<>(getAgentOptions().getOrElse(Collections.emptyList()));
+            agentOptions.add("config-output-dir=" + outputDir.getAbsolutePath());
```
sbrannen, line 82: Since the plugin is responsible for configuring `config-output-dir`, I think an attempt to set that by the user should result in a build failure.

In in the Maven support, I added a note to the reference documentation pointing that out, and in the code I have the following check.

```java
if (value.contains("config-output-dir")) {
    throw new IllegalStateException("config-output-dir cannot be supplied as an agent option");
}
```

I think we should add the same sort of check to the Gradle plugin.

sbrannen (changes requested):
Thanks for pushing the previous round of changes!

I made another pass and came across two additional topics to discuss.

samples/java-application-with-reflection/build.gradle
melix: We can't use `agent` because there's _already_ a property called `agent` (the old one). I also thought about using it, but we can't really for backwards compatibility. So I moved to `agentOptions` because it's similar to other patterns we find in Gradle: `forkOptions` for example.

sbrannen: > We can't use agent because there's already a property called agent (the old one).

sbrannen: > We can't use `agent` because there's already a property called `agent` (the old one).

Yes, of course: backwards compatibility.

Sorry I overlooked that.

> So I moved to `agentOptions` because it's similar to other patterns we find in Gradle: `forkOptions` for example.

Since there's a precedent for that, it's OK as-is.

It's just unfortunate that the actual "options" are called "arguments", but in the end the semantics are effectively the same.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/AgentCommandLineProvider.java
```diff
@@ -79,6 +79,9 @@ public AgentCommandLineProvider() {
         if (getEnabled().get()) {
             File outputDir = getOutputDirectory().getAsFile().get();
             List<String> agentOptions = new ArrayList<>(getAgentOptions().getOrElse(Collections.emptyList()));
+            if (agentOptions.contains("config-output-dir")) {
```
sbrannen, line 82: That check won't suffice. 

You'll need to iterate (or stream) over the supplied options and check if any of the individual strings contains that, since a violation would be in the form of `"config-out-dir=<user supplied output dir>"`.

sbrannen (changes requested):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/AgentCommandLineProvider.java
melix: Of course you're right :facepalm: Fixed!

samples/java-application-with-reflection/build.gradle
melix: I could use `options` in the `agentOptions` too, but for some reason I don't like much the redundancy.

sbrannen: Understood. Redundancy is bad; naming is difficult. 🤷‍♂️

sbrannen (approved):

PR #170, opened 2021-11-26
Rename `agentOptions` -> `agent`

melix:
In order to be more consistent with the actual naming in GraalVM,
the `agentOptions` block which was effectively agent configuration
is renamed to `agent`, at the cost of a breaking change: the
`agent` property was removed from the extension and now returns
the agent configuration.


sbrannen (approved):
LGTM 👍

PR #184, opened 2021-12-20
Rework how disabling of toolchain detection works

### LINKED ISSUE #183: Introduce disableToolchainDetection DSL property and fix toolchain handling in BaseNativeImageOptions
Unless I miss something, based on my tests with Liberica NIK only installed (but apply likely to GraalVM EE as well) and no GraalVM CE installation present locally (not easy because `native-build-tools` requires it to be built), even when applying `disableToolchainDetection = true` to `BuildNativeImageTask` implemented as part of #165 I get the following error:

```
> Task :nativeCompile FAILED

* What went wrong:
Execution failed for task ':nativeCompile'.
> Error while evaluating property 'options.javaLauncher' of task ':nativeCompile'
   > Failed to query the value of property 'javaLauncher'.
      > No compatible toolchains found for request filter: {languageVersion=11, vendor=matching('GraalVM'), implementation=vendor-specific} (auto-detect true, auto-download true)
```

After a deeper analysis, I found that [toolchain relate code in `BaseNativeImageOptions`](https://github.com/graalvm/native-build-tools/blob/00a8cc3bb9755b23f7c3f8d65612885e3afa7da4/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java#L221-L228) need to be executed conditionally as well.

As a consequence, I propose to introduce `disableToolchainDetection` at DSL level (since it will be a popular option at least on Spring side for a while) to be able to just configure:
```
graalvmNative {
    disableToolchainDetection = true
}
```
And to apply it on both `BaseNativeImageOptions` and `BuildNativeImageTask` to be effective.

Be careful, to test you need to temporarily remove GraalVM CE from what is recognized by Gradle toolchain, for my tests I used Liberica NIK (available on SDKMAN) but GraalVM EE can also probably be used.

This issue is critical for Spring Native as it prevents Gradle users to use Liberica NIK which is now our default GraalVM distribution.

melix:
Disabling the toolchain detection by tasks wasn't sufficient,
because Gradle would use the nested `BaseImageOptions` as an input,
and even if the particular `JavaLauchner` input wasn't used if that
flag was disabled, it would still be used internally by Gradle to
snapshot inputs.

As a consequence, the fix only worked if Gradle actually detected
a GraalVM installation somewhere, even if it didn't use it!

To fix this, this commit introduces a top-level DSL method called
`disableToolchainDetection()` which effectively sets the toolchain
to `null` by convention. By making the toolchain optional and null
now Gradle wouldn't complain anymore.

Fixes #183

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -20,7 +20,7 @@ If you are interested in contributing, please refer to our https://github.com/gr
 ==== Gradle plugin
 
 * Fixed resource inference not working on custom binaries
-* Fixed `disableToolchainDetection` not working if a GraalVM installation isn't present. Please use `graalvmNative.disableToolchainDetection()` instead.
+* Fixed `disableToolchainDetection` not working if a GraalVM installation isn't present. Please use `graalvmNative.toolchainDetection.set(false)` instead.
 
 === Release 0.9.8
 
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -20,6 +20,7 @@ If you are interested in contributing, please refer to our https://github.com/gr
 ==== Gradle plugin
 
 * Fixed resource inference not working on custom binaries
+* Fixed `disableToolchainDetection` not working if a GraalVM installation isn't present. Please use `graalvmNative.disableToolchainDetection()` instead.
 
 === Release 0.9.8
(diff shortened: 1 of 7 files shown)


sdeleuze (approved):
Works as expected with 0c4e30b1cb4bc830eedbbce4ef983fd5a7176c62, thanks!

Issue #135, opened 2021-09-28
Support junit3 test framework


lusou-zhangquan:
When we are testing BouncyCastle java project (https://github.com/bcgit/bc-java) based on this native build tool, we found that test cases were not compiled and tested as expected with `gradle nativeTest` command. The reason is that the test cases of this project is based on junit3, while this plugin support junit4+ only. Therefore, we did some work on the basis of this tool to enable test cases based on junit3 framework, and now test cases in BouncyCastle could be built and run automatically. There are many java projects whose test cases are based on junit3 and developers want to test them with this native build tool. Finally, we are glad to contribute this feature to this project if it could be accepted.

lusou-zhangquan:
Thanks for your patient explanation firstly.

As mentioned above, I have tried to run junit3 test cases using this plugin but none was executed.
In my understanding, this plugin works like this:
1. Collect class and method information during executed in jvm mode;
2. Export to config file whose name starts with `junit-platform-unique-ids`;
3. Enable `JUnitPlatformFeature` during graal native image building;
4. Register classes parsed from config file and add them to `testPlan` of `DefaultLauncher`;
5. Launch it and finish test;
Please point it out directly if there is anything incorrect.

And we think the reason that junit3 test cases was not executed is:
1. Test cases base on junit3 are not registered because they are not found in config file exported during jvm execution;
2. `junit-vintage-engine` supports junit3 but its entry `VintageTestEngine` was not enabled in this plugin;

And we tried to fix them by:
1. Enable newly-added `JUnit3PlatformFeature` during graal native image building which collect and register junit3 test cases;
2. Enable `VintageTestEngine` or original `TestRunner` to start junit3 test.

Hope for your better advices. Thanks!

lazar-mitrovic:
Have you managed to run your project using `junit-platform` dependency in JVM mode?

As far as I see, you are still using
```
testCompile group: 'junit', name: 'junit', version: '4.11'
```
and that won't work, and will not be supported. You'll need to [migrate to `junit-platform`](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-running) in order for listener to pick up your tests. You also might need to configure `junit-vintage-engine`. See [this project](https://github.com/junit-team/junit5-samples/tree/main/junit5-migration-gradle) as an example of proper migration to JUnit Platform.

lusou-zhangquan:
[...]

PR #187, opened 2021-12-29
Add scanner to discover junit3 test cases

lusou-zhangquan:
When executing native test, if failed in test listener mode because of missing `junit-platform-unique-ids` config file, it will fall back to test discovery mode at present. This commit provides the third solution: discover test cases in class path and generate test plan manually. With this solution, BouncyCastle java project test cases, which are based on junit3 framework and missing unique id config file after executing in jvm mode, could run and report in native image mode.

#135

### DIFF
--- a/common/junit-platform-native/build.gradle
+++ b/common/junit-platform-native/build.gradle
@@ -59,7 +59,7 @@ dependencies {
     implementation libs.test.junit.platform.console
     implementation libs.test.junit.platform.launcher
     implementation libs.test.junit.jupiter.core
-    testImplementation libs.test.junit.vintage
+    implementation libs.test.junit.vintage
 }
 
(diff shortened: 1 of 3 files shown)


lazar-mitrovic:
I'm not sure about this solution as our original intention was to force users to upgrade to `junit-vintage` in order to reduce maintenance on our side by integrating with latest JUnit releases. 

It seems to me that this approach might create some hard to spot bugs and edge cases, since you are discovering JUnit 4 tests in JVM mode (made using JUnit 4 dependencies) and then executing them in JUnit 5 in native-image. WDYT @sbrannen ?

If we go this route I would prefer if we add this as a opt-in feature (via system property). Another note is that AFAIK JUnit selectors are in internal JUnit API, so they aren't something that we should rely upon.

PR #190, opened 2022-01-06
Do not add option "config-output-dir" when conflicting options are pr…

### LINKED ISSUE #189: Cannot configure "trace-output" agent option because plugins add "config-output-dir" option by default
Using the following in the maven plugin's &lt;configuration> section:

```xml
                            <agent>
                                <options>
                                    <option>trace-output=${basedir}/trace-output.json</option>
                                </options>
                            </agent>
```
 causes `mvn test` to fail with the following message:

```
native-image-agent: can only once specify exactly one of trace-output=, config-output-dir= or config-merge-dir=.
```

Looking at the sources of the maven plugin it appears that the plugin explicitly forbids adding the `config-output-dir` but also does not allow me to remove it. I suggest to allow overriding the `config-output-dir` option by the user and not adding the `config-output-dir` option by default when one of the other options is present.

hannesd:
…esent

This pull request prevents the maven plugin from adding the "config-output-dir" agent option when one of options "config-output-dir", "trace-output", or "config-merge-dir" is already present in the options array. Otherwise the agent would bail and complain: `native-image-agent: can only once specify exactly one of trace-output=, config-output-dir= or config-merge-dir=.`

Since the documentation explicitly states that the "config-output-dir" cannot be overwritten, I assume that this behavior is required, which is why I left this untouched.

Fixes #189

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -41,6 +41,7 @@
 
 package org.graalvm.buildtools.gradle;
 
+import java.util.List;
 import org.graalvm.buildtools.VersionInfo;
 import org.graalvm.buildtools.gradle.dsl.AgentConfiguration;
 import org.graalvm.buildtools.gradle.dsl.GraalVMExtension;
@@ -74,6 +75,7 @@ import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.file.DuplicatesStrategy;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.file.FileSystemLocation;
+import org.gradle.api.internal.provider.AbstractMinimalProvider;
 import org.gradle.api.model.ObjectFactory;
 import org.gradle.api.plugins.ApplicationPlugin;
 import org.gradle.api.plugins.JavaApplication;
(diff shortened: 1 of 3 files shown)


hannesd:
Added changes to both maven and gradle plugin

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -528,6 +530,26 @@ private static void configureAgent(Project project,
         nativeImageOptions.getConfigurationFileDirectories().from(files);
     }
 
+    private static Provider<Directory> agentOutputDirectoryFor(Project project, NativeImageOptions nativeImageOptions, TaskProvider<? extends JavaForkOptions> instrumentedTask) {
```
melix, line 533: This isn't correct, as you are effectively eagerly reading the provider (`orElse(AGENT_...`), which defeats the laziness.

hannesd: It's my first time working with gradle, so I might be mistaken, but as far as I can tell, there are no lazy input values to the computation of the "outputDirUnresolved" variable (the part of the code ending with `orElse(AGENT_...`). All the values involved come from the plugin's configuration which I expect to be constant from start to finish. And since `.getBuildDirectory().dir(outputDirUnresolved)` returns a provider I would assume that whatever it does with outputDirUnresolved will be done lazily.

In case I'm missing something, I have pushed another version where the whole computation is wrapped into one of the Provider implementations I found in gradle's plugin code.

hannesd:
I'm sorry, but I cannot find this repo in the OCA singing form. There are other GraalVM related repos I can choose from, but it's not clear to me which one I should choose. Please advice.

lazar-mitrovic:
> I'm sorry, but I cannot find this repo in the OCA singing form. There are other GraalVM related repos I can choose from, but it's not clear to me which one I should choose. Please advice.

AFAIK you can select GraalVM there.

hannesd:
Hi, I have signed the OCA and received an email saying that all parties have completed the OCA. However, the oca-check does not pick up on it.

lazar-mitrovic:
> Hi, I have signed the OCA and received an email saying that all parties have completed the OCA. However, the oca-check does not pick up on it.

@gilles-duboscq can you please take a look?

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
melix: :x: this isn't correct either. You should not implement your own provider type. I will take a stab at it once the CLA issues are fixed.

gilles-duboscq:
I don't find anything in the [signatories list](https://oca.opensource.oracle.com/?ojr=contrib-list).

hannesd:
Here's the answer I received from docusign/oracle:
See anything suspicious?

oca

gilles-duboscq:
I think what's confusing here is that you get a confirmation email from docusign that sounds like everything is done but actually it just means that the document was sent, not that it was actually processed internally at Oracle.

Thanks for the feedback, it helps debug the process! We'll get your OCA processed and recognized by the bot asap.

melix:
@hannesd While reviewing your code and thinking a bit more about this, I think this PR should fix the problem differently. Instead of looking for the options and trying to figure out if one of `config-output-dir`, or `config-merge-dir` is set, it should probably change how the `trace-output` property is set by the plugins.

That is, because the 3 options are mutually exclusive, there should be a configuration option to tell if you want to generate configuration, if you want to merge configuration, or if you want the trace, in which case we'd use respectively `config-output-dir`, `trace-output` or `config-merge-dir`. That is, we keep the agent output directory, which is used in any case by Gradle, for example, for up-to-date checking, but we simply change the configuration option name based on the profiling mode.

I would then suggest to have a configuration property which defaults to configuration generation, which is what we have today.

Does that make sense?

hannesd:
You mean something like this?
```
<agent>
    <mode>trace</mode>  <!-- replace|merge|trace, default = replace -->
    <options>
        <option>caller-filter-file=${basedir}/src/main/native-image/caller-filter-file.json</option>
        <option>access-filter-file=${basedir}/src/main/native-image/access-filter-file.json</option>
        <!-- config/merge-output-dir, trace-output is set implicitly by the plugin depending on "mode" -->
    </options>
</agent>
```

Sounds good to me. Shall I give it a try?

melix:
Yes, absolutely. If you can't make it I can do it.

PR #193, opened 2022-01-22
Introduce skipNativeTests flag in Maven plugin

### LINKED ISSUE #179: Introduce `skipNativeTests` in Maven plugin
## Overview

`NativeBuildMojo` has a `skipNativeBuild` property.

Whereas, `NativeTestMojo` has a `skipTests` property.

This means that you can skip the tests, but that skips the JVM tests **and** the native tests, since Maven Surefire looks for the same [`skipTests` property](https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html).

Some workflows have the need to execute tests on the JVM using the `agent` in order to generate native configuration files but don't have the need to execute the tests within a native image.

We should therefore support such use cases via a command line similar to the following.

```shell
mvn -Pnative -Dagent=true -DskipNativeTests test
```

## Deliverables

- [x] Introduce a `skipNativeTests` property to skip only the native tests.

sbrannen:
Prior to this commit, the `NativeTestMojo` had a `skipTests` property.
This meant that you could skip the tests, but that skipped the JVM
tests as well as the native tests, since Maven Surefire looks for the
same `skipTests` flag. In other words, there was previously no way to
execute the JVM tests and skip only the native tests.

This commit addresses this issue by introducing a new `skipNativeTests`
flag that allows the user to explicitly disable only native testing
support.

For example, this new feature enables workflows that need to execute
tests on the JVM using the agent in order to generate native
configuration files without having to execute the tests within a native
image. See the updated reference documentation for additional use cases.

Closes #179

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -229,16 +229,21 @@ Currently, this feature requires the execution of the tests in the classic "JVM"
 
 NOTE: This plugin requires JUnit Platform 1.8 or higher.
 
-=== Disabling test support
+[[testing-support-disabling]]
+=== Disabling testing support
 
-There are cases where you might want to disable test support:
+There are cases where you might want to disable native testing support:
 
-- you don't actually want to run your tests in native mode
-- your library or application uses a testing framework that is not supported on the JUnit Platform
+- You don't actually want to run your tests in native mode.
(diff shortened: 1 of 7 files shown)


sbrannen:
Please refrain from merging this PR until @aclement has granted his approval.

aclement:
looks good to me

melix (approved):
LGTM!

PR #223, opened 2022-03-18
Rework argfile generation

melix:
Fixes the fact that paths may contain spaces.

Relates to https://github.com/micronaut-projects/micronaut-gradle-plugin/issues/401

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
@@ -48,6 +48,7 @@ import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static org.graalvm.buildtools.utils.SharedConstants.GRAALVM_EXE_EXTENSION;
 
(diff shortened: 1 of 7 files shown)


alvarosanchez:
@melix I think escaping also needs to be done here: https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java#L270

alvarosanchez:
@melix I think escaping also needs to be done here: https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java#L270

melix:
> @melix I think escaping also needs to be done here: https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java#L270

If I understand correctly, the intent of that line was to support multiple build arguments in a single call, say "-X -Y" should be split to "-X", "-Y". There shouldn't be escaping for build arguments, as those are _extra free arguments_. They shouldn't duplicate arguments which are handled directly by the plugin, like `-cp`.

alvarosanchez:
You are probably right 👍

graemerocher (approved):

PR #226, opened 2022-03-21
Support applications with custom packaging types

alvarosanchez:
If an application has a custom packaging type, but its artefact is really a JAR file,
I will be skipped by the classpath calculation algorithm.

This PR allows the main project artifact to have a different packaging type, which is
the case of Micronaut applications packaged as native images (`<packaging>native-image</packaging>`).


melix:
Thanks, would you mind adding an integration test if it's not too difficult? See `org.graalvm.buildtools.maven.JavaApplicationFunctionalTest` for examples.

melix (approved):

PR #240, opened 2022-05-25
JVM Reachability Metadata support for Maven

alvarosanchez:
Fixes #231

Generated documentation:

wsl%24_Ubuntu_home_alvaro_graalvm_native-build-tools_docs_build_docs_asciidoc_maven-plugin html (1)

PR #241, opened 2022-05-27
Add metadata archives for testing remote repositories

alvarosanchez:
Required for #240 

The contents of the archives are:

```
index.json
org/
org/graalvm/
org/graalvm/internal/
org/graalvm/internal/library-with-reflection/
org/graalvm/internal/library-with-reflection/index.json
org/graalvm/internal/library-with-reflection/1/
org/graalvm/internal/library-with-reflection/1/reflect-config.json
```


melix:
I don't think we should do this. Introducing archives in the repository isn't great since they are opaque and could be hacked. The approach taken by the `org.graalvm.buildtools.gradle.NativeConfigRepoFunctionalTest` is to create those archives _on demand_ during the test.

alvarosanchez:
I already did that in #240

This is to test remote repositories and HTTP downloads.

melix:
Still, if we need an archive then the build should be responsible for building it. Ideally, we don't even need this and can test with a pre-release of the metadata repo.

alvarosanchez:
Closed in favour of [`731b768` (#240)](https://github.com/graalvm/native-build-tools/pull/240/commits/731b7687f8589e64548529e4807e8f263d3611ab)

PR #245, opened 2022-06-01
Set `useArgFile` to be true by default only on Windows

### LINKED ISSUE #243: Avoid using @argfile on Linux and macOS
This makes it hard to see the command line that was used and makes people confused. We should use `@argfiles` only on Windows where it is necessary.

lazar-mitrovic:
Since argument files make native-image invocation non-transparent
we disable them by default except on Windows (where they are needed
in order to avoid errors due to limitation on number of characters).

Fixes #243

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
@@ -81,6 +81,8 @@ import java.util.function.BiFunction;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import static org.graalvm.buildtools.utils.SharedConstants.IS_WINDOWS;
+
 @Mojo(name = "build", defaultPhase = LifecyclePhase.PACKAGE)
 public class NativeBuildMojo extends AbstractNativeMojo {
 
@@ -108,8 +110,8 @@ public class NativeBuildMojo extends AbstractNativeMojo {
     @Parameter(property = "classpath")
     private List<String> classpath;
 
-    @Parameter(property = "useArgFile", defaultValue = "true")
-    private boolean useArgFile;
+    @Parameter(property = "useArgFile")
+    private Boolean useArgFile;
 
     @Parameter(alias = "metadataRepository")
     private MetadataRepositoryConfiguration metadataRepositoryConfiguration;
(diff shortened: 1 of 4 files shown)


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
```diff
@@ -108,8 +110,8 @@ public class NativeBuildMojo extends AbstractNativeMojo {
     @Parameter(property = "classpath")
     private List<String> classpath;
 
-    @Parameter(property = "useArgFile", defaultValue = "true")
-    private boolean useArgFile;
+    @Parameter(property = "useArgFile")
+    private Boolean useArgFile;
```
lazar-mitrovic, line 114: @alvarosanchez I'm not sure if this is the proper way to handle this. 
Per [PluginParameterExpressionEvaluator](https://maven.apache.org/ref/3.8.5/maven-core/apidocs/org/apache/maven/plugin/PluginParameterExpressionEvaluator.html) we could do this by binding this property to something that will be initialized on inject-time, but I don't know where would optimal place for that initialization code be?

melix (changes requested):
The change should be documented in the release notes.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
alvarosanchez: This would certainly work. However, I'd initialise the value (what is done in lines 164-166) in the constructor.

alvarosanchez (changes requested):
Apart from the inline review comment, I miss tests for this change.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
alvarosanchez: Actually, now that I think about it, maybe the constructor is too early. I think Plexus will set via reflection the value of annotated fields after constructing the object, but I'm not really sure.

I prefer [constructor injection](https://github.com/micronaut-projects/micronaut-maven-plugin/blob/master/src/main/java/io/micronaut/build/DockerMojo.java#L52-L56) to avoid that.

lazar-mitrovic:
Closed in favor of #247

PR #247, opened 2022-06-07
Rework Maven plugin

### LINKED ISSUE #221: Better error when `classpath` is empty.
I tried setting up the native build tools in an existing Maven project and put the config into the wrong `pom.xml`. I then ended up with a weird looking problem that the build args were put into the middle of the command that invokes native-image.

As it turned out, the `classpath` was empty and in `NativeBuildMojo.java`, the classpath is unconditionally added to the list of arguments:

https://github.com/graalvm/native-build-tools/blob/551c859f485319eabc7124c47ee02cd41628d963/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java#L135-L138

An empty `classpath` isn't very useful and I didn't realize this was causing the build to fail in weird ways. It would be great if the Maven plugin could fail with an error in such a case, informing users to double check the classpath and their `pom.xml` configs.

### LINKED ISSUE #234: Duplicated classpath entries break tests with GraalVM 22.1+
It seems that at least with the Maven plugin, NBT provide duplicated classpath entries leading to issues like the one described with https://github.com/oracle/graal/issues/4493 or https://github.com/spring-projects-experimental/spring-native/issues/1584.

After a deeper look in the logs, and as pointed one by @olpaw, I see that the plugin pass to `-imagecp` both `/home/seb/workspace/spring-native/samples/jdbc-tx/target/classes` and `/home/seb/workspace/spring-native/samples/jdbc-tx/src/main/resources` leading to duplicated resources in the classpath with GraalVM 22.1 (not with previous versions including 22.0).

As a result, several Spring samples are broken.

To reproduce:
 - Clone `https://github.com/spring-projects-experimental/spring-native`
 - With Java 11 run `./build.sh` at the root
 - `cd samples/jdbc-tx`
 - `mvn clean test -P native` works with GraalVM 22.0 and fails with GraalVM 22.1

### LINKED ISSUE #237: Allow classpath to be built without relying on the main jar artifact
The plugin [currently lets us](https://github.com/graalvm/native-build-tools/blob/92a0aee58c93513e2b556943690df875b1e55ba7/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java#L118-L127) build the classpath ourselves or use a default that includes compile and runtime dependencies, as well as the main artifact.

The main artifact may not be the regular jar of the project, and NBT therefore fails to find the main class (or any class required by the project). This is case with Spring Boot as our repackage goal makes the "application" jar the main artifact. There is a way to configure this differently but we'd like this to work out-of-the-box.

This is not the first time a plugin has to chose between using `target/classes` as the "classpath of the project" over the jar file produced for the main artifact. For instance, the `failsafe-maven-plugin` offers an option that let users chose between those two options, see https://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#classesDirectory. We also document this in [our documentation](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#integration-tests.no-starter-parent).

Building the classpath ourselves but doing so in the parent looks really unpleasant.

**Additional context**
None.

### LINKED ISSUE #243: Avoid using @argfile on Linux and macOS

lazar-mitrovic:
- [Maven] Completely reworked Maven plugin (should fix many of previous issues and inconsistencies between main and test builds).

    Most of the build logic is now moved to `AbstractNativeMojo` in order to be reused by both `NativeBuildMojo` and `NativeTestMojo`.
- [Maven] Added `classesDirectory`, `debug`, `fallback`, `verbose`, `sharedLibrary`, `configurationFileDirectories`, `excludeConfig` and `jvmArgs` properties in order to match those present in the Gradle plugin.
- [Maven and Gradle] `useArgFile` is now set to true by default only on Windows
- [Maven and Gradle] Added `excludeConfig` configuration option that allows skipping of configuration files that are present in classpath `jar` s.

This should fix following issues:
Fixes #221
Fixes  #234
Fixes #237
Fixes #243

Draft until I iron out documentation and tests.


alvarosanchez (approved):

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -114,10 +114,118 @@ Build Configuration]. It is also possible to customize the plugin within a
...
+   If you want to use argument file for native-image building supply the following in the configuration of the plugin:
+[source,xml]
+----
+<useArgFile>true</useArgFile>
+----
+`<quickBuild>`::
+   If you want to build the image using https://blogs.oracle.com/java/post/graalvm-enterprise-221--faster-smarter-leaner[quick build mode], supply the following in the configuration of the plugin:
+[source,xml]
+----
+<quickBuild>true</quickBuild>
```
alvarosanchez, line 164: So in addition to this, what if we allowed users to set this mode also from an environment variable? 

For example, the Micronaut Maven Plugin has quite a few integration tests that involve creating a native image. Instead of going POM by POM adding this config, I wish I could simply pass an environment variable from the GitHub Actions workflow.

lazar-mitrovic: Well you could pass `-DquickBuild` and that should just work. 
I don't know if there is some nice way to also consume properties from env, but I could add a explicit check for `System.getenv("quickBuild")` as well?

alvarosanchez: > Well you could pass `-DquickBuild` and that should just work.

That will involve changing a few places too.

> I don't know if there is some nice way to also consume properties from env, but I could add a explicit check for `System.getenv("quickBuild")` as well?

For plugin authors, there is no other way than `System.getenv()` as far as I know.

I would, however, look for a more specific environment variable name, e.g. `GRAALVM_QUIICK_BUILD`

.github/workflows/test-native-gradle-plugin.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradle-version: ["current", "7.3.3", "7.2", "7.1", "6.8.3","6.7.1"]
```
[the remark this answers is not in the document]
lazar-mitrovic: I wanted to reduce consumption of hosted runners, since there is a limited number of them available for the entire organization at any given time, and 6 of them would be used for each push on each PR just for Gradle testing (and those are long running tests that keep runners busy for a long time - thus preventing others from testing their stuff).

What I believe to be the optimal CI setup:
* On a push to PR we only test first and last supported Gradle version, but configure matrix so that each runner runs only a certain part of tests. That way we get a much quicker response time.
* On a push to master we test more supported versions but configure it so that maybe more of them get tested on few runners (perhaps one or two that would be active for a long time).

I agree that this change shouldn't be here, so I'll revert it and open issue for further discussion.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
```diff
@@ -108,24 +106,33 @@ private List<String> buildExcludeConfigArgs(NativeImageOptions options) {
...
+                                        mid.getGroup(),
+                                        mid.getModule(),
+                                        mid.getVersion()
+                                );
+                                return gav.startsWith(dependency);
+                            }
+                            return false;
+                        });
+                    }).getFiles().getFiles())
+                    .flatMap(Set::stream) // merge resulting sets of files
```
melix, line 128: :x: You can use `.distinct()` and I would collect to a `List` so that the order is reproducible.

docs/src/docs/asciidoc/maven-plugin.adoc
alvarosanchez: Implementation is great! The only thing left is documenting it here

alvarosanchez (changes requested):

alvarosanchez (approved):

melix (approved):

PR #252, opened 2022-06-17
Add docs for Gradle agent modes DSL

lazar-mitrovic:


melix:
This comes too late for the release, it's already in progress.

melix:
Note that because those are only docs, we can 'hotfix' them on GH pages.

PR #256, opened 2022-06-27
Restore `NativeImagePlugin.NATIVE_TEST_EXTENSION`

### LINKED ISSUE #253: Restore `NativeImagePlugin.NATIVE_TEST_EXTENSION`
`NativeImagePlugin.NATIVE_TEST_EXTENSION` removed [here](https://github.com/graalvm/native-build-tools/commit/4fe4210657ee53b945cc037ab268fe41157c0eb6#diff-3ca19e7e49fe64feca3a1e88279c961b56e03450666583025c711da44cfb73bbL138) by mistake should be restored as it is part of NBT public API and can be used by other plugins like [the Spring Native one](https://github.com/spring-projects-experimental/spring-native/blob/main/spring-aot-gradle-plugin/src/main/java/org/springframework/aot/gradle/SpringAotGradlePlugin.java#L173) which is now broken by this change.

lazar-mitrovic:
Fixes #253

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -139,6 +139,7 @@ public class NativeImagePlugin implements Plugin<Project> {
     public static final String NATIVE_TEST_COMPILE_TASK_NAME = "nativeTestCompile";
     public static final String NATIVE_TEST_TASK_NAME = "nativeTest";
     public static final String NATIVE_MAIN_EXTENSION = "main";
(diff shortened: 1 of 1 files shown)


lazar-mitrovic:
cc: @sdeleuze

melix (changes requested):
We should use the constant in https://github.com/graalvm/native-build-tools/blob/b96b0079fd644fb3882f1c6befc1ced7bb834ae8/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java#L251

sdeleuze:
Ok for me, please merge it.

melix:
There are many build failures which I think are related to using latest GraalVM. This isn't giving us good signal. We should probably build the PRs against the latest release of GraalVM, and have separate, cron workflows for `dev`.

PR #257, opened 2022-06-27
Change reachability metadata fallback path

lazar-mitrovic:


melix (changes requested):
This isn't enough. You will also need to update the test resources in https://github.com/graalvm/native-build-tools/tree/f4bb059fc59b4f0c34fecca781cd91bd4034849b/common/jvm-reachability-metadata/src/test/resources

melix (commented):
I think the file paths in the test repositories have to change too, right?

PR #258, opened 2022-06-28
Fix JUnit annotation scanning during the image build

gradinac:


gradinac:
That's a good question - it should be compatible with older GraalVM versions as well. It seems like annotations are still allowed to be initialized at build time without explicitly marking them as such. However, if the annotations then initialize enums, these have to be marked as build time initialized manually. Initializing them at build-time manually is a no-op for older versions (as it would happen there as well)

samples/java-application-with-custom-packaging/pom.xml
```diff
@@ -61,7 +61,7 @@
         <micronaut.version>3.3.4</micronaut.version>
         <exec.mainClass>org.graalvm.demo.Application</exec.mainClass>
         <micronaut.runtime>netty</micronaut.runtime>
-        <maven.native.plugin.version>0.9.12-SNAPSHOT</maven.native.plugin.version>
+        <maven.native.plugin.version>0.9.13-SNAPSHOT</maven.native.plugin.version>
```
melix, line 64: You can run `./gradlew updateSamples` to automatically upgrade all those version numbers.

gradinac: I think they were all updated, but this one was skipped

PR #266, opened 2022-07-05
Final PR for the metadata repo

vjovanov:


samples/native-config-integration/pom.xml
```diff
@@ -148,22 +148,21 @@
             </build>
         </profile>
         <profile>
-            <id>metadataLocal</id>
+            <id>metadataDefault</id>
```
alvarosanchez, line 151: Can you add a new profile (and test) instead of renaming this? Because otherwise we lose testing this case

alvarosanchez (changes requested):

samples/native-config-integration/pom.xml
gradinac: This should now be reverted - I've also removed the test that expects the app to fail when the metadata repo is enabled but not configured, we now use a default repo configuration

alvarosanchez (changes requested):

samples/native-config-integration/pom.xml
vjovanov: Added new tests here, this was just a transient change.

melix (approved):

alvarosanchez (approved):

Issue #273, opened 2022-07-22
Allow to run `native:build` from the command-line


snicoll:
**Is your feature request related to a problem? Please describe.**
It is not possible to run `./mvnw native:build` consistently from the command-line. If you run this goal from a project that you've never built before (i.e. `target` does not exist yet), this yield the following:

```
[ERROR] Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.9.13:build (default-cli) on project demo: Execution default-cli of goal org.graalvm.buildtools:native-maven-plugin:0.9.13:build failed: Cannot invoke "java.io.File.exists()" because "artifactFile" is null
```

However, if `package` has run before (potentially with outdated source) it will pick the state of the "latest" time `package` ran. To avoid that from happening, one can run `./mvnw package native:build`.

**Describe the solution you'd like**
There is a way to instruct a goal that it has to run a phase first. Core maven plugins [use this feature](https://github.com/apache/maven-source-plugin/blob/37ffefea6f479bbd78e89a2df97b4acd5b57f5b3/src/main/java/org/apache/maven/plugins/source/SourceJarMojo.java#L33) so that invoking the goal alone is all that it takes to produces an accurate result.

Spring Boot uses this feature as well for their buildpack integration (i.e. `./mvnw spring-boot:build-image` will first build the application and then invoke buildpack on the updated repackaged archive). 

This makes then the goal not really suitable to being added  to a lifecycle execution, as package would then run twice. Core plugins implement [a `no-fork` goal](https://github.com/apache/maven-source-plugin/blob/37ffefea6f479bbd78e89a2df97b4acd5b57f5b3/src/main/java/org/apache/maven/plugins/source/SourceJarNoForkMojo.java#L38) that does exactly the same thing as the original goal, except it doesn't execute the phase that it needs.

Spring Boot would like to provide a `native` profile that configures both our buildpack integration, and our NBT integration so that users can choose the tool they want. They can currently run `./mvnw -Pnative spring-boot:build-native`. However `./mvnw -Pnative native:build` does not work as expected for the reason explained above.

**Describe alternatives you've considered**
The alternative would be inconsistent command-line options where NBT requires an extra `package`.

**Additional context**
See https://github.com/spring-projects/spring-boot/issues/31770

lazar-mitrovic:
Fixed by #274, as now we have `native:compile` goal.

snicoll:
[...]

PR #274, opened 2022-07-24
Maven plugin fixes

### LINKED ISSUE #270: `NullPointerException` when using 0.9.12 and later to run tests with native image compilation
**Describe the bug**
Running a simple test with `native-maven-plugin` 0.9.12 and later is resulting a `NullPointerException` at build time. 

**To Reproduce**
Reproduce: https://github.com/mpeddada1/junit-testrule

**Expected behavior**
The tests run successfully with the versions in question.

**Logs**
Following the steps in the README of the reproducer is resulting in this stacktrace:

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running com.example.MySampleTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.028 s - in com.example.MySampleTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] 
[INFO] --- native-maven-plugin:0.9.13:test (test-native) @ child-project ---
[INFO] ====================
[INFO] Initializing project: child-project
[INFO] ====================
[INFO] Found GraalVM installation from JAVA_HOME variable.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.770 s
[INFO] Finished at: 2022-07-20T17:43:43-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.9.13:test (test-native) on project child-project: Execution test-native of goal org.graalvm.buildtools:native-maven-plugin:0.9.13:test failed.: NullPointerException -> [Help 1]
```
[Snippet of stacktrace with -X flag ](https://gist.github.com/mpeddada1/9b145382933c837489ec8d1aa56e3cc9)

However, **note** that this test runs successfully with `0.9.11`

 - OS: `Linux`
 - GraalVM Version: 22.1.0 CE
 - Java Version: 11
 - Plugin version: native-maven-plugin 0.9.12 and 0.9.13

I'm wondering if anyone has run into this problem? Thank you!

cc @suztomo @meltsufin

lazar-mitrovic:
Allow running native:build from the command-line #273
NullPointerException fix #270

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) 2020, 2022, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
+ * The above copyright notice and either this complete permission notice or at a
+ * minimum a reference to the UPL must be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package org.graalvm.buildtools.maven;
+
+import org.apache.maven.plugins.annotations.Execute;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+
+/**
+ * This goal builds native images.
+ * It can be invoked using `mvn native:compile`.
+ */
+@Mojo(name = "compile", defaultPhase = LifecyclePhase.PACKAGE,
+        requiresDependencyResolution = ResolutionScope.RUNTIME,
+        requiresDependencyCollection = ResolutionScope.RUNTIME)
+@Execute(phase = LifecyclePhase.PACKAGE)
+@SuppressWarnings("unused")
+public class NativeCompileMojo extends NativeCompileNoForkMojo {
+    // no-op
+}
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
@@ -49,7 +49,7 @@ import org.apache.maven.plugins.annotations.ResolutionScope;
 
 /**
  * This goal builds native images.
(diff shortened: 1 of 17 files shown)


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildNoForkMojo.java
```diff
@@ -0,0 +1,159 @@
...
+import org.codehaus.plexus.component.configurator.expression.ExpressionEvaluationException;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.BiFunction;
+
+
+/**
+ * This goal runs native tests. It functions the same as the native:build goal, but it
```
snicoll, line 63: is "This goal runs native tests" accurate?

snicoll (commented):
Thanks for the quick turnaround!

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildNoForkMojo.java
lazar-mitrovic: Yep that is a mistake (copied and pasted from the test mojo), thanks for noticing!

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeBuildMojo.java
```diff
@@ -41,114 +41,21 @@
...
-    private void maybeSetMainClassFromPlugin(BiFunction<String, String[], String> mainClassProvider, String pluginName, String... nodeNames) {
-        if (mainClass == null) {
-            mainClass = mainClassProvider.apply(pluginName, nodeNames);
-
-            if (mainClass != null) {
-                getLog().info("Obtained main class from plugin " + pluginName + " with the following path: " + String.join(" -> ", nodeNames));
-            }
-        }
-    }
+@Execute(phase = LifecyclePhase.COMPILE)
```
lazar-mitrovic, line 57: @snicoll This works in my testing but I'm not sure if this is a correct thing to do.
Are there cases where resources / additional code is added after the COMPILE phase?

I didn't want to execute PACKAGE phase here since I cannot prevent it from building native-image twice.

snicoll: Compile won't work as it wouldn't have produced the `jar` that NBT requires by default. It should be `package`. 

> I didn't want to execute PACKAGE phase here since I cannot prevent it from building native-image twice.

That's exactly what the `no-fork` variant is for.

alvarosanchez:
I don't see this working. The default configuration is already binding `native:build` to the package phase:

```xml
<execution>
  <id>build-native</id>
  <goals>
    <goal>build</goal>
  </goals>
  <phase>package</phase>
</execution>
```

With the above and `@Execute(phase = LifecyclePhase.PACKAGE)`, the package phase will run twice.

snicoll:
Yes. That's exactly what the "no-fork" goal is all about. I am afraid this is a limitation in Maven and nothing we can do about it.

alvarosanchez:
But then, this is a breaking change. `native:build` should be left without forking, and then another goal (e.g. `native:build-package`) should be added with the `@Execute` annotation.

snicoll:
As I've said in my opening comment that's what core maven plugins do. Yes it's a breaking change but that's ok imo given the benefit. 

The obvious goal to use on the command line is "build" not "build-and-package" or something. 

This isn't the first time breaking changes were made pre-1.0 for the benefit of the community.

alvarosanchez:
Since the version is still 0.9.x, I guess it's ok to break things.

However, documentation needs to reflect this, and the breaking change needs to be made clear in the release notes.

lazar-mitrovic:
- [ ] Change samples and docs before merging this PR

snicoll:
@lazar-mitrovic I don't think a parity for the test mojo is warranted. Contrary to other core plugins, the surefire plugin does not execute the test phase. This change would mean that users having an execution should move to the no-fork variant. 

Can we please consider doing this change for the `build` goal only?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java
```diff
@@ -70,7 +70,8 @@
 import static org.graalvm.buildtools.Utils.NATIVE_TESTS_EXE;
 
 /**
- * @author Sebastien Deleuze
+ * This goal runs native tests. It functions the same as the native:test goal, but it
+ * does not fork the build, so it is suitable for attaching to the build lifecycle.
```
snicoll, line 74: I think that's outdated now.

snicoll (commented):
Thanks for the update!

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
```diff
@@ -0,0 +1,61 @@
...
+ * This goal builds native images.
+ * It can be invoked using `mvn native:compile`.
+ */
+@Mojo(name = "compile", defaultPhase = LifecyclePhase.PACKAGE,
+        requiresDependencyResolution = ResolutionScope.RUNTIME,
+        requiresDependencyCollection = ResolutionScope.RUNTIME)
+@Execute(phase = LifecyclePhase.PACKAGE)
+@SuppressWarnings("unused")
+public class NativeCompileMojo extends NativeCompileNoForkMojo {
+    // no-op
```
snicoll, line 60: A bit odd... What does "no-op" mean here?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileMojo.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.apache.maven.plugins.annotations.Execute;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+
+/**
+ * This goal builds native images.
+ * It can be invoked using `mvn native:compile`.
```
snicoll, line 52: Perhaps worth clarifying that it is meant to be invoked on the command-line as a single goal or something. You're doing it already but the command-line bit isn't explicit.

PR #278, opened 2022-07-27
Fix how the GRAALVM_QUICK_BUILD env var is read

### LINKED ISSUE #277: Value for property 'quickBuild' is final and cannot be changed any further
If the `GRAALVM_QUICK_BUILD` environment variable is set, then calling `gradle nativeCompile` will fail with the following error:

```
Caused by: java.lang.IllegalStateException: The value for property 'quickBuild' is final and cannot be changed any further.
        at org.gradle.api.internal.provider.AbstractProperty$FinalizedValue.beforeMutate(AbstractProperty.java:489)
        at org.gradle.api.internal.provider.AbstractProperty.assertCanMutate(AbstractProperty.java:263)
        at org.gradle.api.internal.provider.AbstractProperty.setSupplier(AbstractProperty.java:212)
        at org.gradle.api.internal.provider.DefaultProperty.set(DefaultProperty.java:71)
        at org.graalvm.buildtools.gradle.tasks.BuildNativeImageTask.exec(BuildNativeImageTask.java:175)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:125)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:51)
        at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:29)
```

This is caused by this line: https://github.com/graalvm/native-build-tools/blob/bed8b18e30db3f67472f8360d7cc40a5ad5acc14/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java#L175

Configuration _must not_ be changed at task execution time. Instead the value of the property should simply be read, which means that the `NativeImageOptions` should be configured to read their default value from the `GRAALVM_QUICK_BUILD` environment variable instead.

melix:
It was being mutated at execution time, instead of read at configuration
time, which causes an error (see https://github.com/micronaut-projects/micronaut-sql/runs/7538645832?check_suite_focus=true#step:6:907).

Fixes #277

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -663,15 +663,16 @@ public class NativeImagePlugin implements Plugin<Project> {
                                 FileSystemOperations fileOperations,
                                 Task taskToInstrument,
                                 JavaForkOptions javaForkOptions) {
+        Provider<AgentConfiguration> agentConfiguration = AgentConfigurationFactory.getAgentConfiguration(agentMode, graalExtension.getAgent());
         //noinspection Convert2Lambda
         taskToInstrument.doFirst(new Action<Task>() {
             @Override
             public void execute(Task task) {
-                logger.logOnce("Instrumenting task with the native-image-agent: " + task.getName());
+                if (agentConfiguration.get().isEnabled()) {
+                    logger.logOnce("Instrumenting task with the native-image-agent: " + task.getName());
+                }
             }
         });
-        Provider<AgentConfiguration> agentConfiguration = AgentConfigurationFactory.getAgentConfiguration(agentMode, graalExtension.getAgent());
-
         AgentCommandLineProvider cliProvider = project.getObjects().newInstance(AgentCommandLineProvider.class);
         cliProvider.getInputFiles().from(agentConfiguration.map(AgentConfiguration::getAgentFiles));
         cliProvider.getEnabled().set(agentConfiguration.map(AgentConfiguration::isEnabled));
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -663,7 +663,13 @@ public class NativeImagePlugin implements Plugin<Project> {
                                 FileSystemOperations fileOperations,
                                 Task taskToInstrument,
                                 JavaForkOptions javaForkOptions) {
-        logger.lifecycle("Instrumenting task with the native-image-agent: " + taskToInstrument.getName());
+        //noinspection Convert2Lambda
+        taskToInstrument.doFirst(new Action<Task>() {
+            @Override
+            public void execute(Task task) {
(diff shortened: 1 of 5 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java
```diff
@@ -240,7 +243,13 @@ public BaseNativeImageOptions(String name,
         getDebug().convention(false);
         getFallback().convention(false);
         getVerbose().convention(false);
-        getQuickBuild().convention(false);
+        getQuickBuild().convention(providers.environmentVariable("GRAALVM_QUICK_BUILD").map(env -> {
+            LOGGER.warn("Quick build environment variable is set.");
```
alvarosanchez, line 247: In a multi-project build, this is too verbose IMO:

image

alvarosanchez (approved):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java
melix: Good call, let me fix this.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -663,7 +663,13 @@ private void configureAgent(Project project,
                                 FileSystemOperations fileOperations,
                                 Task taskToInstrument,
                                 JavaForkOptions javaForkOptions) {
-        logger.lifecycle("Instrumenting task with the native-image-agent: " + taskToInstrument.getName());
+        //noinspection Convert2Lambda
+        taskToInstrument.doFirst(new Action<Task>() {
```
gradinac, line 667: Thank you for fixing this! :) One note - maybe we could log this message only if we're running with the agent enabled - while technically we are instrumenting these tasks every time, it only has an effect if we're running with the agent

gradinac (approved):
The changes look good! :)

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
melix: I'm having troubles figuring out the reliable way to tell if the agent is enabled from this method. Is it possible?

gradinac: Hmm, I think we could capture the `AgentConfiguration` provider from below inside the `doFirst` lambda and check if it's enabled through the `isEnabled`call

PR #296, opened 2022-09-12
Add support for override flag

sdeleuze:
@gradinac @melix Could you please review this PR intended to be used with oracle/graalvm-reachability-metadata#51?

@aclement Could you please give your agreement for this PR and further refinements to be merged?


common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/Configuration.java
```diff
@@ -0,0 +1,63 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.reachability;
+
+import java.nio.file.Path;
+
+public class Configuration {
```
melix, line 45: Would you mind renaming this to `ConfigurationDirectory` or something similar? `Configuration` is a bit overloaded, in particular in Gradle, and it can be confusing to read.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -348,6 +350,48 @@ private void configureJvmReachabilityConfigurationDirectories(Project project, G
...
+                                                artifact.forceConfigVersion(forcedVersions.get(module));
+                                            }
+                                        });
+                                    }
+                                    query.useLatestConfigWhenVersionIsUntested();
+                                }).stream()
+                                .filter(org.graalvm.reachability.Configuration::isOverride)
+                                .flatMap(configuration -> Stream.of(
+                                        "--exclude-config",
+                                        "/" + moduleVersion.getName() + "-" + moduleVersion.getVersion() + "\\.jar$",
```
melix, line 386: This is quite fragile, but this has to be "fixed' somehow in `native-image`: relying on file names is not great. This would break for classifiers for example, or even for different variants of a single module (e.g guava).

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/Configuration.java
sdeleuze: Sure I will.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
sdeleuze: On Maven side, I think I retrieve the exact dependency file name associated with an artifact where the override flag is true, do we have a way to do the same on Gradle side?

lazar-mitrovic: @sdeleuze Gradle supports excluding using GAV coordinates:
```
excludeConfig.put("org.example.test:artifact:version", ["META-INF/native-image/*", "config/*"]) // Excludes configuration that matches one of given regexes from JAR of dependency with said coordinates.
```

So you might be able to utilize this :)

sdeleuze: Very good point, thanks for sharing @lazar-mitrovic, I will update the PR accordingly.

aclement:
From my point of view, looks good.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
sdeleuze: Ok so I did the change in 8d9478cd0c003a3e43fc6a0a29f367a328065075, exclusion are passed as expected to `native-image` but they did not work in practice because `excludeConfig` is broken. I fixed it via f8dfca52a059d28fe761c117be36fb9556f7bac9 and now it works as expected.

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/Configuration.java
sdeleuze: Renamed to `DirectoryConfiguration`.

sdeleuze:
I have rebased the PR on top of `master`, squashed the commits that made sense. Should be ready to merge.

melix:
@sdeleuze there are failing checkstyle rules and tests, would you mind taking a look?

sdeleuze:
@melix Should be all green now.

melix:
Merged, thank you!

PR #304, opened 2022-09-16
Fix compatibility with configuration cache

### LINKED ISSUE #302: Agent instrumentation prevents the use of Gradle's configuration cache
**Describe the bug**
When the agent instruments a task, it breaks Gradle's configuration cache.

**To Reproduce**
```groovy
plugins {
	id 'java'
	id 'org.graalvm.buildtools.native' version '0.9.13'
}

repositories {
	mavenCentral()
}

graalvmNative {
    toolchainDetection = false
}
```
```bash
$ gradle test --configuration-cache
Configuration cache is an incubating feature.
Calculating task graph as no configuration cache is available for tasks: test
[native-image-plugin] Instrumenting task with the native-image-agent: test

* What went wrong:
Configuration cache problems found in this build.

7 problems were found storing the configuration cache, 4 of which seem unique.
- Task `:test` of type `org.gradle.api.tasks.testing.Test`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.5.1/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:test` of type `org.gradle.api.tasks.testing.Test`: cannot serialize object of type 'org.gradle.api.internal.tasks.DefaultSourceSetContainer', a subtype of 'org.gradle.api.tasks.SourceSetContainer', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.5.1/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:test` of type `org.gradle.api.tasks.testing.Test`: cannot serialize object of type 'org.gradle.api.internal.tasks.DefaultTaskContainer', a subtype of 'org.gradle.api.tasks.TaskContainer', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.5.1/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
- Task `:test` of type `org.gradle.api.tasks.testing.Test`: value 'map(map(org.gradle.api.file.Directory property(org.gradle.api.file.Directory, fixed(class org.gradle.api.internal.file.DefaultFilePropertyFactory$FixedDirectory, /Users/awilkinson/dev/temp/native-image-instrumentation-problem/build)) org.gradle.api.internal.file.DefaultFilePropertyFactory$PathToDirectoryTransformer@76ce53fc))' failed to unpack provider
```

**Expected behavior**
Agent instrumentation does not prevent the use of Gradle features.

**Logs**
[configuration-cache-report.zip](https://github.com/graalvm/native-build-tools/files/9582523/configuration-cache-report.zip)

 - OS: N/A
 - GraalVM Version: N/A
 - Java Version: N/A
 - Plugin version: 0.9.13
 - Gradle version: 7.5.1

**Additional context**
Disabling the agent does not help:

```groovy
graalvmNative {
	agent {
		enabled = false
	}
	toolchainDetection = false
}
```

7 problems are still reported. However, configuring the predicate to avoid any instrumentation does help:

```groovy
graalvmNative {
	agent {
		java.util.function.Predicate<Test> predicate = { t -> false }
		tasksToInstrumentPredicate = predicate
	}
	toolchainDetection = false
}
```

With this configuration in place, 0 problems are reported.

melix:
This commit reworks the agent task actions so that they are compatible with the configuration cache. The problem was that when the configuration cache is enabled, it will try to serialize the task actions that we add, so that if the task is not up-to-date and that the configuration didn't change, it can replay the action.

The problem was that our task action used, as an input, a `Provider<List<String>>`, where those strings represented paths which do _not_ exist when the action is serialized yet. In practice, we don't care, because those are implementation details of the task action: the directories will not exist _before_ the task is executed, and not _after_, they only exist during task execution.

The configuration cache, however, wants to know about all inputs. We could have changed the action so that it uses a _root directory_ instead of a list of directories as input, and make the computation of the `session-` directories _during_ the task action: this would have worked because the only input would be the "output directory" that we pass as root. However, the task action is also used in another task which uses completely different inputs, which are not derived from a single directory.

At the same time, we _don't_ need the `Provider` semantics in the task action: all we care about is to have a supplier which will generate a list of directories. Therefore, we changed the `Provider` type to a `Supplier` one: the lambda will be serialized, but the only thing it will capture is the `outputDir` provider which is fine, configuration-cache wise.

Now, let me take a pill.

Fixes #302


sdeleuze (approved):
Not sure I am fully qualified to evaluate properly this PR, but the code changes LGTM. I will double check before the release it does not introduce regressions with Spring apps.

PR #320, opened 2022-09-28
Add .sdkmanrc

philwebb:
This commit adds a `.sdkmanrc` file to the root of the project so that users of https://sdkman.io/ don't need to remember to switch VMs when working on the project.


aclement:
looks good to me.

alvarosanchez (approved):

PR #321, opened 2022-09-28
Fix functional tests for MacOS users

philwebb:
Whilst working on the codebase I found that functional tests fail on MacOS. The folling pull-request fixes the issue.


aclement:
looks good to me

PR #322, opened 2022-09-28
Include buildtool support for bundling reachability metadata into jar files

philwebb:
I believe that this request was discussed recently with @bclozel who kindly wrote the following:

## Motivation

Currently, Native Build Tools build plugins fetch reachability metadata from the official repository right before building a native image of the application.
While the library in charge of fetching the metadata repository is available and can be reused in different contexts, there is still a non-trivial amount of code required to do the following:

* get reachability metadata for the libraries in the application classpath
* apply filtering options described in the plugin configuration

Frameworks using NBT can generate their own reachability metadata before packaging the application as a JAR.
It would be useful if vanilla or shaded JARs could be packaged with all the required reachability metadata (Framework + metadata repository) so that such JARs could be used by CI or cloud infrastructure to generate a native image directly.

This would provide the following advantages:

* the application JAR is the single source for running the application in JVM mode, or building it as a native binary
* this enables build reproducibility - without this change, building the JAR as a native binary at different times could yield different results if the shared repo has changed.

## Requested behavior

The current change proposal would require a new build tool task, for the Maven and Gradle plugins. Those tasks would:

1. Fetch the repository, collect the relevant metadata, apply filters and in general the plugin configuration related to this step
2. Write the resulting metadata to a location configurable in the plugin. By default the location should align with the build tool defaults to pack these files with the resulting JAR.
3. Be designed to be run independently of other NBT tasks
4. Be designed to be called by Framework plugins or as a task/goal directly by the developer

### DIFF
--- a/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/DirectoryConfiguration.java
+++ b/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/DirectoryConfiguration.java
@@ -40,15 +40,34 @@
  */
 package org.graalvm.reachability;
 
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardCopyOption;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.Collection;
 
 public class DirectoryConfiguration {
+
+    private static final String PROPERTIES = "reachability-metadata.properties";
+
+    private final String groupId;
+
+    private final String artifactId;
+
+    private final String version;
 
     private final Path directory;
 
     private final boolean override;
 
-    public DirectoryConfiguration(Path directory, boolean override) {
+    public DirectoryConfiguration(String groupId, String artifactId, String version, Path directory, boolean override) {
+        this.groupId = groupId;
+        this.artifactId = artifactId;
+        this.version = version;
         this.directory = directory;
         this.override = override;
     }
@@ -60,4 +79,49 @@ public class DirectoryConfiguration {
     public boolean isOverride() {
         return override;
     }
+
+    public static void copy(Collection<DirectoryConfiguration> configurations, Path destinationDirectory) throws IOException {
+        Path nativeImageDestination = destinationDirectory.resolve("META-INF").resolve("native-image");
+        for (DirectoryConfiguration configuration : configurations) {
+            Path target = nativeImageDestination
+                    .resolve(configuration.groupId)
+                    .resolve(configuration.artifactId)
+                    .resolve((configuration.version != null) ? configuration.version :
+                            configuration.getDirectory().getFileName().toString());
+            copyFileTree(configuration.directory, target);
+            writeConfigurationProperties(configuration, target);
+        }
+    }
+
+    private static void copyFileTree(Path source, Path target) throws IOException {
+        Files.walkFileTree(source, new SimpleFileVisitor<Path>() {
+
+            @Override
+            public FileVisitResult preVisitDirectory(Path directory, BasicFileAttributes attrs) throws IOException {
+                Files.createDirectories(target.resolve(source.relativize(directory)));
+                return FileVisitResult.CONTINUE;
+            }
+
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                if (!"index.json".equalsIgnoreCase(file.getFileName().toString())) {
+                    Files.copy(file, target.resolve(source.relativize(file)), StandardCopyOption.REPLACE_EXISTING);
+                }
+                return FileVisitResult.CONTINUE;
+            }
+        });
+    }
+
+    private static void writeConfigurationProperties(DirectoryConfiguration configuration, Path target)
+            throws IOException {
+        StringBuilder content = new StringBuilder();
+        if (configuration.isOverride()) {
+            content.append("override=true\n");
+        }
+        if (content.length() > 0) {
+            Files.write(target.resolve(PROPERTIES), content.toString().getBytes(StandardCharsets.ISO_8859_1));
+        }
+    }
 }
+
+
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -660,6 +660,23 @@ This may be interesting if there's no specific metadata available for the partic
 include::../../../../samples/native-config-integration/pom.xml[tag=metadata-force-version]
 ----
 
+=== Adding metadata respoistory files
+
+By default, repository metadata will be used only when your native image is generated.
+In some situations, you may want to include the metadata directly inside your jar.
+
+Adding metadata to your jar can be useful when some other process is responsible for converting your jar into a native image.
+For example, you might be generating a shaded jar and using a https://paketo.io/[Paketo buildpack] to convert it to a native image.
+
+To include metadata repository inside your jar you can use the `add-reachability-metadata` goal.
+Typically the goal will be included in an execution step where by default it will be bound to the `generate-resources` phase:
+
+.Configuring the `add-reachability-metadata` goal to execute with the `generate-resources` phase
+[source,xml,indent=0]
+----
+include::../../../../samples/native-config-integration/pom.xml[tag=add-reachability-metadata-execution]
+----
+
 [[javadocs]]
 == Javadocs
 
--- a/docs/src/docs/snippets/gradle/groovy/build.gradle
+++ b/docs/src/docs/snippets/gradle/groovy/build.gradle
@@ -207,3 +207,9 @@ graalvmNative {
     }
 }
 // end::specify-metadata-version-for-library[]
+
+// tag::include-metadata[]
+tasks.named("jar") {
+    from collectReachabilityMetadata
+}
+// end::include-metadata[]
--- a/docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
+++ b/docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
@@ -221,3 +221,9 @@ graalvmNative {
     }
 }
 // end::specify-metadata-version-for-library[]
+
+// tag::include-metadata[]
+tasks.named("jar", Jar) {
+	from(collectReachabilityMetadata)
+}
+// end::include-metadata[]
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -77,6 +77,7 @@ import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ConfigurationContainer;
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
+import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.AttributeContainer;
 import org.gradle.api.file.ArchiveOperations;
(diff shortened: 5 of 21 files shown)


philwebb:
There are four commits in this pull-request. The first updates common support code to enable copying. The second adds Maven support and the third adds Gradle support.

The fourth commit is a minor refactoring that attempts to remove duplicated code. It isn't directly related to this pull-request and could be dropped or submitted as another PR if required.

aclement:
approved by me as a contribution. Will let the engineers comment on the details :)

philwebb:
> I'm still not convinced that bundling metadata in the jar is the right thing to do. I understand that this is aimed at final applications, not libraries, but it could create a precedence for what we ask folks not to do.

The Maven goal and Gradle task are pretty generic at the moment, the only problem is the defaults that I picked. If we change things so that the destination always needs to be specified, would that be better?

That way we're adding general purpose tasks/goals without guiding the user in potentially bad practices. I can also change the documentation to remove any suggestion of adding files to jars.

> This is only the case if you don't use the native build tools, right? Because we are reproducible

Right. If you're always using NBT then the metadata repository is pinned and things are repeatable. The reproducibility only happens if you want some other process to create the native image. In that case, for the other process to work it will either need the metadata, or it will need to know the GAVs of dependencies and the version of the metadata repository to use.

> For Micronaut for example, we provide a way to build native images in Docker, using the same build process, so it doesn't matter if you build locally or on CI.

I think that's something we'd like to do eventually as well. There's a lot less ceremony in creating a Docker image for a native executable. It's basically a `from JVM` and here's a layer with the executable.

There is one benefit of having Source -> Jar -> Native Image, namely that you can take the same jar and create native images for lots of different architectures. As I understand it, you can't currently create an MacOS executable on a Linux OS.

> So I'm not against this PR, but we should be clearer with regards to what it is suitable for, and I would not recommend to put the metadata in the jar (actually I would suggest to publish another artifact, possibly with a classifier, which only contains metadata, and possibly other "build reproducibility" metadata).

If we make the goal/task general purpose then I think folks could use it to create copy the files wherever they see fit. I suspect it would be quite easy to configure a Gradle build to create a distinct artifact. I imagine it's also possible with Maven, albeit with more configuration.

Thanks for taking the time to look it. I'll await your feedback then update the PR.

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/DirectoryConfiguration.java
```diff
@@ -60,4 +80,54 @@ public Path getDirectory() {
...
+                    .resolve(configuration.groupId)
+                    .resolve(configuration.artifactId)
+                    .resolve((configuration.version != null) ? configuration.version :
+                            configuration.getDirectory().getFileName().toString());
+            copyFileTree(configuration.directory, target);
+            if (configuration.isOverride()) {
+                overrides.put(target, configuration);
+            }
+        }
+        for (Map.Entry<Path, DirectoryConfiguration> entry : overrides.entrySet()) {
```
sdeleuze, line 98: As detailed on https://github.com/spring-projects/spring-boot/issues/31782#issuecomment-1266435509, I don't think `--exclude-config` is supported in `native-image.properties` so this part probably needs to be removed from that PR.

docs/src/docs/snippets/gradle/groovy/build.gradle
```diff
@@ -207,3 +207,11 @@ graalvmNative {
     }
 }
 // end::specify-metadata-version-for-library[]
+
+// tag::include-metadata[]
+task reachabilityMetadata(type: org.graalvm.buildtools.gradle.tasks.ReachabilityMetadataCopyTask)
```
melix, line 212: :x: This should use the lazy APIs:

```
tasks.register("reachabilityMetadata", org.graalvm.buildtools.gradle.tasks.ReachabilityMetadataCopyTask)
```

docs/src/docs/snippets/gradle/groovy/build.gradle
```diff
@@ -207,3 +207,11 @@ graalvmNative {
...
+
+jar {
```
melix, line 214: :x: This should use the lazy APIs

```
tasks.named("jar") {
   from reachabilityMetadata
}
```

docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
```diff
@@ -221,3 +221,11 @@ graalvmNative {
     }
 }
 // end::specify-metadata-version-for-library[]
+
+// tag::include-metadata[]
+var reachabilityMetadata = tasks.register<org.graalvm.buildtools.gradle.tasks.ReachabilityMetadataCopyTask>("reachabilityMetadata")
+
+tasks.withType<Jar> {
```
melix, line 228: :x: This is inconsistent with the Groovy version. Either you want to update _only_ the jar, and this should be:

```
tasks.named("jar", Jar) {
...
}
```
or, you want to update all the jar files (I wouldn't recommend) and this should be:

```
tasks.withType<Jar>().configureEach {
    ...
}
```

(note the `configureEach` to ensure laziness)

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -271,6 +273,21 @@ private void configureJavaProject(Project project, Provider<NativeImageService>
             task.getMergeWithExisting().set(graalExtension.getAgent().getMetadataCopy().getMergeWithExisting());
             task.getToolchainDetection().set(graalExtension.getToolchainDetection());
         });
+
+        GraalVMReachabilityMetadataRepositoryExtension metadataRepositoryExtension = reachabilityExtensionOn(graalExtension);
+        TaskCollection<ReachabilityMetadataCopyTask> reachabilityMetadataCopyTasks = project.getTasks()
```
melix, line 282: :x: I think I would change this to basically _always_ register the task. Currently the configuration works a bit backwards: it waits for the user to create a task of type `ReachabilityMetadataCopyTask` and then configures it. This isn't quite Gradle idiomatic: I think the plugin should register the task, always, and configure some defaults (convention). This would also simplify documentation.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/ReachabilityMetadataCopyTask.java
```diff
@@ -0,0 +1,152 @@
...
+    public Configuration getClasspath() {
+        return classpath;
+    }
+
+    public void setClasspath(Configuration classpath) {
+        this.classpath = classpath;
+    }
+
+    @Internal
+    public abstract Property<GraalVMReachabilityMetadataService> getMetadataService();
```
melix, line 88: I am unsure if this is compatible with the configuration cache. @eskatos can you put some light on this?

samples/metadata-repo-integration/build.gradle
```diff
@@ -66,3 +67,9 @@ graalvmNative {
         enabled = true
     }
 }
+
+task reachabilityMetadata(type: org.graalvm.buildtools.gradle.tasks.ReachabilityMetadataCopyTask)
```
melix, line 71: :x: Same comment here: both the task creation and the jar task configuration should use the lazy APIs.

melix (changes requested):
I have added comments on the implementation for the Gradle plugin, I haven't looked at the Maven side yet (possibly @alvarosanchez can look?).

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/ReachabilityMetadataCopyTask.java
eskatos: It is compliant to use a build service for a task property.
We are working on a lower ceremony way but for now don't forget to do `task.usesService(serviceProvider)` as shown in  https://docs.gradle.org/current/userguide/build_services.html#registering_a_build_service

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/DirectoryConfiguration.java
philwebb: I think we're going to need some way to discover if metadata had the `override` flag set so that we can reconstruct the exclude patterns later. The easiest way I can think of to do that is to write a specific properties file next to the metadata JSON. I'll update the PR to do that.

docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
philwebb: Thanks. I've not used the Kotlin DSL much so this as all a bit alien to me.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
philwebb: I'll add a "nativeReachabilityMetadata" (most of the other task names start with "native") then let the conventions apply as before.

philwebb:
@melix I've forced pushed an update addressing the previous review comments. The only one I'm not too sure about is registering the task by default. I've gone with a simple `getTasks().register(...)` call and left the conventions as a `configureEach` callback. I hope that's what you meant, but let me know if I've got it wrong.

I picked `nativeReachabilityMetadata` as that task name. There's already `nativeBuild`, `nativeTest`, `nativeBuild`, `nativeTestBuild` so the `native` prefix seemed sensible. I noticed that there's also a `metadataCopy` task that's registered. I wonder if at some point this should be renamed to something like `nativeAgentMetadata` or `nativeAgentMetadataCopy`?

sdeleuze:
Not sure I would prefix pure metadata generation by `native` since this prefix implies a native compilation and metadata are designed to be not tied to native, but I agree that's a grey area given the name of the plugin. I think I would be fine with `metadataGenerate` or something like that, up to @melix to decide.

I will discuss the `reachability-metadata.properties` thing with @vjovanov and share my feedback here later.

melix:
It happens that we _already_ have a `metadataCopy` task (https://github.com/oracle/graalvm-reachability-metadata/blob/master/docs/CollectingMetadata.md) which is aimed at copying metadata collected by the agent into the local sources. It's unfortunate that we have a name clash because this would have made sense to use this task name.

So maybe, for this new task, we should use a different name like `collectReachabilityMetadata`?

sdeleuze:
What about `metadataGenerate` since here it is IMO more a generation than a copy? Also that would be closer to current naming pattern.

melix:
> What about metadataGenerate since here it is IMO more a generation than a copy? Also that would be closer to current naming pattern.

The task already exists as `metadataCopy`, so it would be a breaking change. One could argue that it _is_ a copy, since its responsibility is to copy the metadata from the agent output directory into another place. So I tend to think it's ok to keep `metadataCopy` as is today, and use `collectReachabilityMetadata` for the new task, which has a more explicit meaning imo.

philwebb:
I like `collectReachabilityMetadata`. I'll update the PR and take another look at the naming used in the Maven plugin. Thanks @melix!

sdeleuze:
After discussing with @vjovanov, I think we should move forward with what you implemented @philwebb with `reachability-metadata.properties`

philwebb:
Forced push with the suggested task name.

melix:
@alvarosanchez do you have time to review the Maven side of this PR or should I give it a try?

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -660,6 +660,23 @@ This may be interesting if there's no specific metadata available for the partic
...
+=== Adding metadata respoistory files
+
+By default, repository metadata will be used only when your native image is generated.
+In some situations, you may want to include the metadata directly inside your jar.
+
+Adding metadata to your jar can be useful when some other process is responsible for converting your jar into a native image.
+For example, you might be generating a shaded jar and using a https://paketo.io/[Paketo buildpack] to convert it to a native image.
+
+To include metadata repository inside your jar you can use the `add-reachability-metadata` goal.
+Typically the goal will be included an execution step where by default it will be bound to the `generate-resources` phase:
```
alvarosanchez, line 672:
```suggestion
Typically the goal will be included in an execution step where by default it will be bound to the `generate-resources` phase:
```

alvarosanchez (approved):

melix (approved):

sdeleuze:
Thanks to you for this epic contribution!

PR #326, opened 2022-10-05
Rich output should also enable progress printing.

fniephaus:
I think the rich output option should also enable progress printing. That's also enabled by default within Native Image and works in Maven. It's weird if it's not behaving the same way in Gradle.


melix (approved):

fniephaus:
It turns out that the interactive Gradle output and the progress printing from Native Image are incompatible. So this unfortunately cannot be merged.

PR #343, opened 2022-10-24
Add maven support for agent

dnestoro:


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/MetadataCopyMojo.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@Mojo(name = "metadata-copy", defaultPhase = LifecyclePhase.PREPARE_PACKAGE)
+public class MetadataCopyMojo extends AbstractMojo {
+
+    @Parameter(alias = "metadataCopy")
```
alvarosanchez, line 23: `alias` is to specify a different name than the Java field. If they are the same, you don't need it.

```suggestion
    @Parameter
```

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/MetadataCopyMojo.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
+        ArrayList<Path> sources = new ArrayList<>(Arrays.asList(Paths.get(project.getBuild().getOutputDirectory()), Paths.get(project.getBuild().getTestOutputDirectory())));
+        ArrayList<Path> destinations = metadataCopy.getOutputDirectories();
+
+        for (Path source: sources) {
+            for (Path destination : destinations) {
+                try {
+                    Files.copy(source, destination);
```
alvarosanchez, line 37: This isn't correct. `Files.copy()` copies a single file.

Also, `project.getBuild().getOutputDirectory()` is `target/classes` and `project.getBuild().getTestOutputDirectory()` is `target/test-classes`, but that is not where the agent dumps the configuration files. 

You need to copy **all files** from `target/native/agent-output/main` and `target/native/agent-output/test` into the destination(s).

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeExtension.java
```diff
@@ -129,24 +133,19 @@ public void afterProjectsRead(MavenSession session) {
...
-                assert x != null;
-                System.out.println(x.getName() + " ------- " + x.getParent().getName());
+                AgentConfiguration agent = AgentUtils.collectAgentProperties(session, (Xpp3Dom) nativePlugin.getConfiguration());
 
                 // Test configuration
                 withPlugin(build, "maven-surefire-plugin", surefirePlugin -> {
                     configureJunitListener(surefirePlugin, testIdsDir);
-                    if (isAgentEnabled) {
-                        List<String> agentOptions = getAgentOptions(nativePlugin, Context.test, selectedOptionsName);
+                    if (agent != null) {
```
alvarosanchez, line 141: It would be better if `AgentUtils.collectAgentProperties()` always return a non-null `AgentConfiguration` instance with the enabled flag set so that you can do:

```java
if (agent.isEnabled()) { ... }
```

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/config/MetadataCopyConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.File;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class MetadataCopyConfiguration {
+
+    @Parameter
+    private List<Path> outputDirectories;
```
alvarosanchez, line 14: Please add javadoc since that goes into the generated `plugin.xml` and shows in the autocomplete documentation for users.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/config/MetadataCopyConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @Parameter(defaultValue = "${project.basedir}/src/main/resources/META-INF/native-image")
+    private Path outputDirectory;
```
alvarosanchez, line 17: Please add javadoc since that goes into the generated `plugin.xml` and shows in the autocomplete documentation for users.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/MetadataCopyMojo.java
alvarosanchez: Please add javadoc since that goes into the generated `plugin.xml` and shows in the autocomplete documentation for users.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/MetadataCopyMojo.java
```diff
@@ -0,0 +1,45 @@
...
+    @Parameter(alias = "metadataCopy")
+    private MetadataCopyConfiguration metadataCopy;
+
+    @Parameter(defaultValue = "${project}", readonly = true, required = true)
+    private MavenProject project;
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
+        ArrayList<Path> sources = new ArrayList<>(Arrays.asList(Paths.get(project.getBuild().getOutputDirectory()), Paths.get(project.getBuild().getTestOutputDirectory())));
+        ArrayList<Path> destinations = metadataCopy.getOutputDirectories();
```
alvarosanchez, line 32: Do not use `ArrayList` for declaring variables, always use its interface (`List` in this case)

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/config/MetadataCopyConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public ArrayList<Path> getOutputDirectories() {
+        ArrayList<Path> tmp = new ArrayList<>(outputDirectories);
+        tmp.add(outputDirectory);
+        return tmp;
```
alvarosanchez, line 22: `outputDirectories` and `outputDirectory` can be null.

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/AgentUtils.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.maven.execution.MavenSession;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.graalvm.buildtools.agent.*;
+
+import java.util.ArrayList;
+import java.util.stream.Collectors;
+
+import static org.graalvm.buildtools.utils.Utils.parseBoolean;
+
+public class AgentUtils {
```
alvarosanchez, line 12: Make it final and add a private default constructor

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/Utils.java
```diff
@@ -0,0 +1,24 @@
+package org.graalvm.buildtools.utils;
+
+public class Utils {
```
alvarosanchez, line 3: Final and private default constructor

samples/java-application-with-reflection/pom.xml
```diff
@@ -138,27 +138,40 @@
                         </executions>
                         <!-- tag::native-plugin-agent-options[] -->
                         <configuration>
-                            <!-- end::native-plugin-agent-options[] -->
```
alvarosanchez, line 141: You have removed Asciidorctor tags that are used to render snippets in the documentation

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -57,6 +57,18 @@ public class AgentConfiguration implements Serializable {
...
 
+    // This constructor should be used only to specify that we have instance of agent that is disabled (to avoid using null for agent enable check)
+    public AgentConfiguration() {
+        this.callerFilterFiles = null;
+        this.accessFilterFiles = null;
+        this.builtinCallerFilter = false;
+        this.builtinHeuristicFilter = false;
+        this.experimentalPredefinedClasses = false;
+        this.experimentalUnsafeAllocationTracing = false;
+        this.trackReflectionMetadata = false;
```
alvarosanchez, line 68: The fields already have those values by default. Remove.

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -78,6 +78,54 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractGraalVMMavenFunctio
         outputDoesNotContain "containers found"
     }
 
+    def "simple test of agent usage"() {
```
alvarosanchez, line 81: The test is actually testing metadata copy

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -78,6 +78,54 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractGraalVMMavenFunctio
...
+
+        then:
+        buildSucceeded
+        outputDoesNotContain "Cannot merge agent files because native-image-configure is not installed. Please upgrade to a newer version of GraalVM."
+        outputDoesNotContain "returned non-zero result"
+        outputDoesNotContain "Agent files cannot be copied."
+        outputContains "Metadata copy process finished."
+    }
+
+    def "simple test of agent usage with direct mode"() {
```
alvarosanchez, line 97: Improve name too

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -78,6 +78,54 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractGraalVMMavenFunctio
...
+        mvn  '-PagentConfigurationDirectMode', '-DskipNativeTests', 'native:metadata-copy'
+
+        then:
+        buildSucceeded
+        outputDoesNotContain "Cannot merge agent files because native-image-configure is not installed. Please upgrade to a newer version of GraalVM."
+        outputDoesNotContain "returned non-zero result"
+        outputContains "You are running agent in direct mode. Skipping both merge and metadata copy tasks."
+    }
+
+    def "simple test of agent usage with disabled stages"() {
```
alvarosanchez, line 112: Improve name too

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/Xpp3DomParser.java
```diff
@@ -0,0 +1,79 @@
...
+            if (retVal != null) {
+                return retVal;
+            }
+        }
+
+        return null;
+    }
+
+    public static LinkedList<Xpp3Dom> getAllTagsByName(Xpp3Dom root, String name) {
+        LinkedList<Xpp3Dom> listOfTags = new LinkedList<>();
```
alvarosanchez, line 66: Use interfaces for declaring variables

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
dnestoro: This class is also used by gradle plugin, so I am not sure if we can remove default values.

alvarosanchez: https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html

dnestoro:
Style issues will be fixed once main logic gets approval.

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -509,43 +509,35 @@ experimental features such as `experimental-class-loader-support` or advanced fe
...
+* `conditional` - in this mode you can provide additional files that can be used as a filter for the agent output (https://github.com/oracle/graalvm-reachability-metadata/blob/master/docs/CollectingMetadata.md[see this]). You
+can read more about conditional mode https://www.graalvm.org/22.1/reference-manual/native-image/ExperimentalAgentOptions/#generating-conditional-configuration-using-the-agent[here].
+
+Each option and how you should use it is described bellow:
+
+* `enabled` - a simple flag that specifies whether the agent is on or off. Can be set to true or false.
+* `defaultMode` - agent mode switcher. Can be set to: standard, direct or conditional.
+* `modes` - list of additional mode options, specific for the mode type. Inside this tag, you can specify options for
+ direct or conditional modes. Standard mode doesn't have any specific options.
+* `commonOptions` - list of options that can be specify independent from agent mode. More about common options can be found
```
sdeleuze, line 534: Could you elaborate on the choice of `commonOptions` versus `options`?

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -509,43 +509,35 @@ experimental features such as `experimental-class-loader-support` or advanced fe
...
+This example shows all possibilities you can use for native agent configuration. If you don't need some of the options,
+just remove them.
+
+Agent can be run in one of the following modes:
+
+* `standard` - in this mode you run agent only with options provided in `commonOptions` section.
+* `direct` - in this mode you can provide command line that you want to be executed as an agent configuration. In this
+     mode, user is fully responsible for agent configuration, and the rest of the config file will be ignored.
+* `conditional` - in this mode you can provide additional files that can be used as a filter for the agent output (https://github.com/oracle/graalvm-reachability-metadata/blob/master/docs/CollectingMetadata.md[see this]). You
+can read more about conditional mode https://www.graalvm.org/22.1/reference-manual/native-image/ExperimentalAgentOptions/#generating-conditional-configuration-using-the-agent[here].
```
sdeleuze, line 526: Outdated link

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/AgentUtils.java
```diff
@@ -0,0 +1,207 @@
...
+                agentMode = new DirectAgentMode(options);
+                break;
+            default:
+                throw new Exception("Unknown agent mode selected: " + mode);
+        }
+
+        return agentMode;
+    }
+
+    public static AgentConfiguration collectAgentProperties(MavenSession session, Xpp3Dom rootNode) throws RuntimeException {
```
sdeleuze, line 122: I thought one of the goal of the refactoring was to stop doing manual parsing of the nodes and leverage regular binding to also benefit from autocomplete in the IDE. Could you please elaborate on the limit that prevent us to use regular binding and if if we now have a good DevXP (no "everything in red" when we configure the agent and full auto-complete in the IDE)?

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/NativeImageConfigurationUtils.java
```diff
@@ -57,7 +56,7 @@
 /**
  * Utility class containing various native-image and JVM related methods.
  */
-public abstract class Utils implements SharedConstants {
+public abstract class NativeImageConfigurationUtils implements SharedConstants {
```
sdeleuze, line 59: Worth to test that with Spring/Micronaut to be sure it does not break current integrations (we had that kind of regression in the past).

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/Utils.java
```diff
@@ -0,0 +1,66 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.buildtools.utils;
+
+public final class Utils {
+
+    private Utils() { }
```
sdeleuze, line 45: Maybe better to remove that private constructor and make the class abstract.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
```diff
@@ -67,7 +67,7 @@ class IsolatedMavenExecutor {
         cliArgs.add("-cp")
         cliArgs.add(classpath)
         if (debug) {
-            cliArgs.add("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005")
+            cliArgs.add("-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005")
```
sdeleuze, line 70: Could you explain why this change is needed?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -41,42 +41,42 @@
...
-import org.apache.maven.plugin.MojoExecutionException;
-import org.apache.maven.plugin.descriptor.PluginDescriptor;
-import org.apache.maven.plugins.annotations.Component;
-import org.apache.maven.plugins.annotations.Parameter;
-import org.apache.maven.toolchain.ToolchainManager;
-import org.graalvm.buildtools.Utils;
-import org.graalvm.buildtools.maven.config.ExcludeConfigConfiguration;
-import org.graalvm.buildtools.utils.NativeImageUtils;
-import org.graalvm.buildtools.utils.SharedConstants;
+import static org.graalvm.buildtools.utils.NativeImageConfigurationUtils.getNativeImage;
```
sdeleuze, line 79: Probably more consistent and explicit to not use static import here to know what Utils class we use.

sdeleuze (commented):
Please provide feedback on this first set of questions, thanks!

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/NativeImageConfigurationUtils.java
dnestoro: @alvarosanchez could you please test this PR with Micronaut?
@bclozel could you please test this PR with Spring?

docs/src/docs/asciidoc/maven-plugin.adoc
dnestoro: First reason for this choice is to align with `gradle` implementation and have uniform agent manipulation mechanism. Also, with `options` there are no semantic meaning for any option (you can put whatever you want inside '<option>` tag). On the other hand, with `commonOptions` we have named options, which makes the file more readable, and easier for parsing.

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/AgentUtils.java
dnestoro: Since we are using agent configuration in the `NativeExtension` as well, we need to parse pom file manually, to collect configuration in this stage. Also we are using regular binding as well, to prevent "everything in red" and to use agent options in two mojos.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
dnestoro: This option is used for debugging. With this option enabled (`y`) process will wait for us to start the debugger. If the option is set to `n` you need to click debugger button (in IDE) multiple times to catch the moment when process allows you to attach the debugger (you don't know when it will be possible). I don't see a reason for this option to be `n`. If there is a need for that, please explain me.

docs/src/docs/asciidoc/maven-plugin.adoc
sdeleuze: > First reason for this choice is to align with gradle implementation and have uniform agent manipulation mechanism.

Not sure what you mean about Gradle implementation, but I see no usage of `commonOptions` in https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html.

> Also, with options there are no semantic meaning for any option (you can put whatever you want inside <option> tag). On the other hand, with commonOptions we have named options, which makes the file more readable, and easier for parsing.

What prevents us to use an `<options>` element with named options?

docs/src/docs/asciidoc/maven-plugin.adoc
sdeleuze: Can you please update the link?

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
sdeleuze: Maybe because this is used for test.

@melix Could you provide a feedback on this since I think you committed those lines of code?

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/NativeImageConfigurationUtils.java
sdeleuze: Compilation of Boot seems ok.

@alvarosanchez Could you please check on Micronaut side?

docs/src/docs/asciidoc/maven-plugin.adoc
alvarosanchez: I think what David is trying to say is better understood with an example:

```xml                              
<modes>
	<conditional>
		<userCodeFilterPath>user-code-filter.json</userCodeFilterPath>
		<parallel>true</parallel>
	</conditional>
</modes>
<commonOptions>
	<builtinCallerFilter>true</builtinCallerFilter>
	<builtinHeuristicFilter>true</builtinHeuristicFilter>
	<enableExperimentalPredefinedClasses>false</enableExperimentalPredefinedClasses>
	<enableExperimentalUnsafeAllocationTracing>false</enableExperimentalUnsafeAllocationTracing>
	<trackReflectionMetadata>true</trackReflectionMetadata>
</commonOptions>
```

`modes.conditional` is mapped with a `java.util.Properties`, therefore, children tags can be whatever. IIRC, they are passed as-is to GraalVM.

`commonOptions`, on the other hand, is mapped with `CommonOptionsConfiguration` class, which has a well-know set of options mapped as `@Parameters`.

Having said that, I think all Sébastien is asking (please correct me if I'm wrong) is to simply rename it as `options`, something I would agree with.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
alvarosanchez: It is used in the tests, yes. When you want to attach a debugger, and suspend is `n`, you need to watch a gigantic log for the corresponding message. If you try too early, it will fail. If you try too late, you may have missed your mojo.

Suspending will basically wait for you to attach so that you don't miss the correct moment to do it.

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/NativeImageConfigurationUtils.java
alvarosanchez: @sdeleuze as per your last review comment, it doesn't really work, does it?

sdeleuze: I meant compilation of Boot itself depending on the NBT snapshot of this PR was fine, usage in a project was broken but now fixed I think and likely unrelated to this class rename.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
sdeleuze: So should this value be changed from `n` to `y` as suggested above from your POV?

docs/src/docs/asciidoc/maven-plugin.adoc
sdeleuze: Yep

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/IsolatedMavenExecutor.groovy
alvarosanchez: Yes

sdeleuze:
When configuring:
```xml
<configuration>
	<agent>
		<enabled>true</enabled>
	</agent>
</configuration>
```

I see the following error:
```
GraalVM native-image is missing from your system.
 Make sure that GRAALVM_HOME environment variable is present.
```
Like the other parts of the plugin, I think it is important that it works without making `GRAALVM_HOME` mandatory, just with `JAVA_HOME` defined.

dnestoro:
> When configuring:
> 
> ```
> <configuration>
> 	<agent>
> 		<enabled>true</enabled>
> 	</agent>
> </configuration>
> ```
> 
> I see the following error:
> 
> ```
> GraalVM native-image is missing from your system.
>  Make sure that GRAALVM_HOME environment variable is present.
> ```
> 
> Like the other parts of the plugin, I think it is important that it works without making `GRAALVM_HOME` mandatory, just with `JAVA_HOME` defined.

I agree with this idea. The only thing I am thinking about is should I do this inside this PR or I should create separate PR. It looks like this is commonly used in many parts of the repo and not only related to the agent. What is your opinion?

sdeleuze:
If this is not a regression introduced by that PR, we could probably fix it in another PR but IMO we should do that before releasing NBT `0.9.20`.

native-maven-plugin/src/main/java/org/graalvm/buildtools/utils/NativeImageConfigurationUtils.java
alvarosanchez: I tested this PR with Micronaut too, doesn't seem to break anything

dnestoro:
I tested plugin with `netty` project, and discovered two new edge cases. Last commits cover those cases.

sdeleuze (approved):

alvarosanchez (approved):

linghengqian:
> I tested plugin with `netty` project, and discovered two new edge cases. Last commits cover those cases.

- Is there a draft PR showing on the `netty/netty` side?  I'm curious when the version of NBT that includes this PR will be released.

Issue #346, opened 2022-10-24
Add a check for the minimal GraalVM version


sdeleuze:
In order to allow users and frameworks to deal with GraalVM compatibility explicitly and provide meaningful errors to end users, Maven and Gradle plugins should allow to specify the minimal GraalVM versions supported.

sdeleuze:
I will focus for now on minimal version. See my proposal for the configuration bellow;

Maven
```
<configuration>
    <requireVersion>22.3</requireVersion>
</configuration>
```

Gradle
```
graalvmNative {
    requireVersion = "22.3"
}
```
For now `requireVersion` could be `MAJOR`, `MAJOR.MINOR` or `MAJOR.MINOR.PATCH` and means that the version specified and more recent will be accepted.

We can introduce later the range syntax of our choice to be able to specify upper bounds.

melix:
What worries me, at least for Gradle, is that we're inventing a system which will interfere with the toolchain support (in other words, this belongs to the toolchain support, not the NBT configuration). I am tempted to say we should wait for the issue to be fixed by the Gradle team, so that we can reason on the "secondary" version, rather than the language version (/cc @ljacomet). Also with the merge of `native-image` in OpenJDK, I wonder if that "secondary" version will make sense in the future.

sdeleuze:
It may not make sense when the OpenJDK integration is complete, and this is GraalVM specific so not sure Gradle toolchain should deal with that as a first class concern, but there are concrete needs for this today and next year at least.

PR #353, opened 2022-10-28
Add a check for the minimal GraalVM version

sdeleuze:
This commit adds a new Gradle and Maven requiredVersion configuration option that specifies the minimal GraalVM version and can be set to `MAJOR`, `MAJOR.MINOR` or `MAJOR.MINOR.PATCH`.

Closes gh-346

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
@@ -49,11 +49,18 @@ import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static org.graalvm.buildtools.utils.SharedConstants.GRAALVM_EXE_EXTENSION;
 
 public class NativeImageUtils {
+
+    private static final Pattern requiredVersionPattern = Pattern.compile("^([0-9]+)(?:\\.([0-9]+)?)?(?:\\.([0-9]+)?)?$");
+
+    private static final Pattern graalvmVersionPattern = Pattern.compile("^GraalVM ([0-9]+)\\.([0-9]+)\\.([0-9]+).*");
+
     public static void maybeCreateConfigureUtilSymlink(File configureUtilFile, Path nativeImageExecutablePath) {
         if (!configureUtilFile.exists()) {
             // possibly the symlink is missing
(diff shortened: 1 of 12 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -220,4 +229,28 @@ public void exec() {
         }
     }
 
+    private void checkRequiredVersionIfNeeded(File executablePath, NativeImageOptions options) {
+        if (!options.getRequiredVersion().isPresent()) {
+            return;
+        }
+        try {
```
melix, line 236: :x: This should use `getExecOperations()` instead, as using `ProcessBuilder` is known to have issues (in particular under Windows). Should be quite straightforward API.

common/utils/src/test/java/org/graalvm/buildtools/utils/NativeImageUtilsTest.java
```diff
@@ -0,0 +1,105 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.graalvm.buildtools.utils;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class NativeImageUtilsTest {
```
melix, line 46: Maybe you want to add a test with a version from GraalVM EE. For example I get this:

> GraalVM 22.0.0.1 Java 11 EE (Java Version 11.0.14+8-LTS-jvmci-22.0-b03)

or 

> GraalVM 22.2.0 Java 17 EE (Java Version 17.0.4+11-LTS-jvmci-22.2-b05)

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
```diff
@@ -102,4 +109,44 @@ public static String escapeArg(String arg) {
...
+    public static void checkVersion(String requiredVersion, String versionToCheck) {
+        Matcher requiredMatcher = requiredVersionPattern.matcher(requiredVersion);
+        if (!requiredMatcher.matches()) {
+            throw new IllegalArgumentException("Invalid version " + requiredVersion + ", should be for example \"22\", \"22.3\" or \"22.3.0\".");
+        }
+        Matcher checkedMatcher = graalvmVersionPattern.matcher(versionToCheck);
+        if (!checkedMatcher.matches()) {
+            throw new IllegalArgumentException("Version to check " + versionToCheck + " can't be parsed.");
+        }
+        int requiredMajor = Integer.parseInt(requiredMatcher.group(1));
```
melix, line 128: :o: This could be refactored to have a method which checks a single number, then call it for major, minor and patch.

melix (changes requested):
I'm wondering if it works for dev (snapshot) versions too.

common/utils/src/test/java/org/graalvm/buildtools/utils/NativeImageUtilsTest.java
sdeleuze: Already done see https://github.com/graalvm/native-build-tools/pull/353/files/ccb866e7be63f2c4bfc5a693ccd1070729ee2f58#diff-9fe652c1f823c1c1ec47896ec8c5d857a2f2fed41391f86434eb7e95ced50f59R79-R83

melix: oh thanks, looks like I'm blind :D

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
sdeleuze: Ok done

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
sdeleuze: Given the custom message in the logs (which is IMO important), I am not sure, could you please double check?

sdeleuze:
It works for snapshots, this is tested in `checkGraalVMCEDevVersion()`.

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
melix: Yeah, we could use a static method with lambdas, but I'm not sure it would be easier to read.

sdeleuze:
@melix I think I have taken in account your feedback, please let me know if that's ok or if you think I should change something else.

PR #372, opened 2022-11-25
Fix compile task out of date when runtime args change

### LINKED ISSUE #371: Using `runtimeArgs` causes the _compile_ task to be out of date
Changing the _runtime arguments_ of the native image extension will cause the _compilation_ of the image to be out-of-date. This is a problem because runtime args should only be used... at runtime.

A possible solution is to rework the inputs of the task so that it doesn't use the extension directly.

melix:
The `NativeImageOptions` domain object is used in the DSL to configure both compile and runtime arguments for native images: the compile options are used to invoke `native-image` and build a binary, while the runtime args are used to pass extra arguments when executing the native image.

The full `NativeImageOptions` domain object was passed to the native image compilation task as a `@Nested` input, which meant in practice that if the _runtime_ arguments changed, then we would rebuild the image although it isn't necessary.

This commit fixes the problem in a binary compatible way, by introducing a couple of interfaces: one for the compile options, the other for the runtime options. It makes the "options" internal on the compile task, and uses a delegated view of compile options for input snapshotting.

Fixes #371

### DIFF
--- a/.github/workflows/test-native-gradle-plugin.yml
+++ b/.github/workflows/test-native-gradle-plugin.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        gradle-version: ["current", "6.7.1"]
+        gradle-version: ["current", "7.1"]
         # Following versions are disabled temporarily in order to speed up PR testing
         # "7.3.3", "7.2", "7.1", "6.8.3"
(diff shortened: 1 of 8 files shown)


sdeleuze (approved):
Please reach me when merged to allow me to check snapshots does not introduce regression with Spring Boot 3.

PR #374, opened 2022-11-25
Fix repository not downloaded from Maven Central

### LINKED ISSUE #373: Default metadata repository isn't downloaded from Maven Central
If the user doesn't configure a metadata repository version **and** doesn't configure a specific URL, we should download the metadata repository from Maven Central (or, more specifically, from the repositories declared in the build).

Unfortunately, there are a couple places where the metadata repository service can be registered, and one of those places doesn't set the correct URL.

I think this comes from a merge conflict which wasn't done properly, as the service initialization should have been done in a single method called from 2 places instead.

melix:
Fixes #373


sdeleuze (commented):
Does Maven plugin downloads metadata from Maven Central already?

melix:
> Does Maven plugin downloads metadata from Maven Central already?

It does: https://github.com/graalvm/native-build-tools/blob/51baf9fe0db80eefee127451e26be5aa2018202f/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java#L208-L217

PR #396, opened 2023-02-08
Set regex to accept both GraalVM and native-image as version names

dnestoro:


fniephaus (approved):
LGTM. I can confirm this works with a non-native SVM driver.

Issue #403, opened 2023-02-16
For Maven Plugin 0.9.20, the `native:metadata-copy` task cannot be executed normally


linghengqian:
**Before reporting**

- This repository should be used to report **issues on the Maven or Gradle plugins for GraalVM**.
- Please report issues which are specific to [the Spring Framework](https://spring.io/) or [the Micronaut framework](https://micronaut.io/) to their specific repositories.
- Do not report issues with building your specific application, e.g errors which happen at image build time like classes initialized at build time, or missing classes as run time: those are not related to the plugins but problems with configuration. You can refer to the [GraalVM native image documentation](https://www.graalvm.org/latest/reference-manual/native-image/) for available options and the [plugins documentation](https://graalvm.github.io/native-build-tools) for how to use them with the plugin.

**Describe the bug**
**Make sure that you have read [the documentation](https://graalvm.github.io/native-build-tools) and that you are using the latest plugin version.**

- For Maven Plugin 0.9.20, the `native:metadata-copy` task cannot be executed normally. Although there is no error in the log, the final json file is not output to the custom `outputDirectory`.
- ![image](https://user-images.githubusercontent.com/20187731/219303653-b65e8391-9d14-44e5-bc1f-bcb228069da8.png)

**To Reproduce**

When possible, provide a link to a repository which reproduces the issue, with instructions on how to use.
The reproducer **must** make use of either the Maven or Gradle plugin.

- Minimal reproducible unit tests and steps to reproduce are at https://github.com/linghengqian/maven-v0920-metadatacopy-test .

```bash
sudo apt install unzip zip curl sed -y
curl -s "https://get.sdkman.io" | bash
source "$HOME/.sdkman/bin/sdkman-init.sh"
sdk install java 22.3.1.r17-grl
sdk use java 22.3.1.r17-grl
gu install native-image
sudo apt-get install build-essential libz-dev zlib1g-dev -y

git clone git@github.com:linghengqian/maven-v0920-metadatacopy-test.git
cd ./maven-v0920-metadatacopy-test/
./mvnw -Pnative -Dagent=true clean test native:metadata-copy
```
Please use backticks to [properly format code](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting).
[...]

PR #407, opened 2023-02-27
Enable metadata copy when agent is enabled from cmd

dnestoro:
This PR should fix [this](https://github.com/graalvm/native-build-tools/issues/403) issue. Previously metadata-copy mojo didn't check if agent is enabled via command line


alvarosanchez (changes requested):
This is not the way to fix the issue.

Instead, you need to use the `property` attribute of the `@Parameter` annotation in `AgentConfiguration`:

```java
@Parameter(property = "agent")
private boolean enabled;
```

Quoting their documentation:

> Property to use to retrieve a value. Can come from <code>-D</code> execution, setting properties or pom properties.

alvarosanchez (approved):

PR #408, opened 2023-02-28
Add a new write-args-file goal

alvarosanchez:
This PR introduces a goal that only writes the args file, so that other plugins further in the lifecycle can use it. In particular, the Micronaut Maven Plugin needs this goal.

The path to the args file is stored in the `MavenProject` properties, an approach that is used by other plugins such as the JaCoCo Maven Plugin.

### DIFF
--- a/docs/src/docs/asciidoc/index.adoc
+++ b/docs/src/docs/asciidoc/index.adoc
@@ -21,10 +21,6 @@ If you are using alternative build systems, see <<alternative-build-systems.adoc
 
 === Release 0.9.21
 
-==== Gradle plugin
-
-- TBD
-
 ==== Maven plugin
 
 - Add a new `native:write-args-file` goal that can be used to write the arguments passed to `native-image` to a file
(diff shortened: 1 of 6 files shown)


docs/src/docs/asciidoc/index.adoc
```diff
@@ -19,6 +19,16 @@ If you are using alternative build systems, see <<alternative-build-systems.adoc
 [[changelog]]
 == Changelog
 
+=== Release 0.9.21
+
+==== Gradle plugin
+
+- TBD
```
melix, line 26: suggestion: remove this as it is not necessary that there will be Gradle fixes in this release

melix (approved):
question: this only works for the "main" arguments. Do you need something similar for `test`? Also, will this approach work with Micronaut AOT?

alvarosanchez:
> Do you need something similar for test?

The use case is persisting the arguments that `native:compile|no-fork` would use to produce a native executable, so that they can be reused within Docker, for example.

> will this approach work with Micronaut AOT?

It does work! 🎉 
It is being tested in https://github.com/micronaut-projects/micronaut-maven-plugin/pull/643 (which needs the new NBT snapshot to pass)

dnestoro (approved):

PR #420, opened 2023-04-03
Disable toolchain detection by default

melix:
With the changes coming in GraalVM regarding versioning, in particular the fact that the GraalVM version becomes internal and that the vendor string is changing, toolchain detection will clearly be broken for all users.

This commit therefore disables toolchain detection by default, and makes sure to document properly the new behavior.


fniephaus (changes requested):

fniephaus (approved):
LGTM, thanks!

PR #425, opened 2023-04-06
Fix lookup of metadata repository

### LINKED ISSUE #424: Native gradle plugin - Default metadata used even when a custom version provided
If a custom metadata version is set without uri, the default metadata version is used instead of custom one.
```
graalvmNative {
    metadataRepository {
        version = "0.2.7"
    }
}
```

The `computeMetadataRepositoryUri` method (`NativeImagePlugin` class) compares `defaultUri` and `configuredUri` which are equal when only `version` is set like in the example above therefore the maven artifact is used which contains default metadata.

**System Info:**
 - OS: Mac
 - GraalVM Version: 22.3.0 CE
 - Java Version: 17
 - Plugin version: native-gradle-plugin:0.9.20

melix:
This commit fixes the URI which is constructed when only a version is set in the metadata repository. Before this change, whatever version was used in the configuration, we would ignore it, because we were testing against the wrong URI to figure out if the user had configured something explicitly or not.

Now, we will prioritize the user URI, then construct a GitHub release URI from the version which is configured, and eventually use Maven Central if that URI is the same as the default one.

Fixes #424

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
@@ -132,6 +132,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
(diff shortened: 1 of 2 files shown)


gradinac (approved):

melix:
If the build passes I will leave merging to @dnestoro since we're in the process of releasing 0.9.21 and that it would introduce a merge conflict.

dnestoro:
Since we finished the `0.9.21` release, I am merging this PR

PR #432, opened 2023-04-28
Add support for `default-for` attribute

sdeleuze:
This new `index.json` attribute is design to allow more flexible version matching to avoid breakage due to switch on latest version for untested ones.

It should have a `default-for` key and a regexp as a string value, for example "1\\.0\\..*" which should match versions like "1.0.0", "1.0.1", "1.0.2", etc.

Please do not merge before @aclement has given his +1.

See oracle/graalvm-reachability-metadata#62
See oracle/graalvm-reachability-metadata#275

### DIFF
--- a/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
+++ b/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
@@ -91,7 +91,8 @@ public class SingleModuleJsonVersionToConfigDirectoryIndex implements VersionToC
      */
     @Override
     public Optional<DirectoryConfiguration> findConfiguration(String groupId, String artifactId, String version) {
-        return findConfigurationFor(groupId, artifactId, version, artifact -> artifact.getVersions().contains(version));
+        return findConfigurationFor(groupId, artifactId, version,
+                artifact -> artifact.getVersions().contains(version) || artifact.isDefaultFor(version));
     }
 
     @Override
@@ -101,15 +102,16 @@ public class SingleModuleJsonVersionToConfigDirectoryIndex implements VersionToC
     }
 
     /**
-     * Returns the latest configuration directory for the requested artifact.
+     * Returns the matching configuration directory for the requested artifact.
      *
      * @param groupId the group ID of the artifact
      * @param artifactId the artifact ID of the artifact
+     * @param version the version of the artifact
      * @return a configuration directory, or empty if no configuration directory is available
      */
     @Override
     public Optional<DirectoryConfiguration> findLatestConfigurationFor(String groupId, String artifactId, String version) {
-        return findConfigurationFor(groupId, artifactId, version, Artifact::isLatest);
+        return findConfigurationFor(groupId, artifactId, version, artifact -> artifact.isDefaultFor(version) || artifact.isLatest());
     }
 
     private Optional<DirectoryConfiguration> findConfigurationFor(String groupId, String artifactId, String version,
--- a/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
+++ b/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
@@ -91,8 +91,9 @@ public class SingleModuleJsonVersionToConfigDirectoryIndex implements VersionToC
      */
     @Override
     public Optional<DirectoryConfiguration> findConfiguration(String groupId, String artifactId, String version) {
-        return findConfigurationFor(groupId, artifactId, version,
-                artifact -> artifact.getVersions().contains(version) || artifact.isDefaultFor(version));
+        Optional<DirectoryConfiguration> exactMatch = findConfigurationFor(groupId, artifactId, version, artifact -> artifact.getVersions().contains(version));
+        return exactMatch.isPresent() ? exactMatch :
+                findConfigurationFor(groupId, artifactId, version, artifact -> artifact.isDefaultFor(version));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
```diff
@@ -101,15 +102,16 @@ public Optional<DirectoryConfiguration> findLatestConfigurationFor(String groupI
...
      *
      * @param groupId the group ID of the artifact
      * @param artifactId the artifact ID of the artifact
+     * @param version the version of the artifact
      * @return a configuration directory, or empty if no configuration directory is available
      */
     @Override
     public Optional<DirectoryConfiguration> findLatestConfigurationFor(String groupId, String artifactId, String version) {
-        return findConfigurationFor(groupId, artifactId, version, Artifact::isLatest);
+        return findConfigurationFor(groupId, artifactId, version, artifact -> artifact.isDefaultFor(version) || artifact.isLatest());
```
fniephaus, line 114: Doesn't this assume that entries are ordered by version? Is that a valid assumption or should `findConfigurationFor(String groupId, String artifactId, String version, Predicate<? super Artifact> predicate)` sort entries before filtering by `predicate`?

melix: Yes, I agree that there is a wrong assumption here. A similar assumption is made in `findConfiguration`: we should search for an exact match first, and only fallback to the default if not found.

melix (changes requested):

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
sdeleuze: Good catch, I agree we need to perform 2 different invocations for `findConfigurationFor`, but let's maybe clarify in what order.

> we should search for an exact match first, and only fallback to the default if not found.

By "exact match first" you mean searching first `Artifact::isLatest` then `artifact.isDefaultFor(version)`? I am asking because if you have such metadata:
```
[
  { "module": "com.foo:bar", "tested-versions": ["1.0"], "metadata-version": "1.0", "default-for": "1\\..*" },
  { "module": "com.foo:bar", "tested-versions": ["2.0", "2.1"], "metadata-version": "2.0", "latest": true }
]
```
I am wondering what result we want when searching the latest configuration for `1.0`.

melix: By exact match I mean that we must first look for metadata which corresponds exactly to the version we're looking for. When it's not found, then we need to fallback to the default. And in case both `default-for` and `latest` exist, I think `default-for` should take precedence when it matches.

fniephaus: Do we have a definition of what `latest` really means? The name suggests it marks the latest metadata for a module, but that doesn't say anything about the version of the module the metadata is for. I find `latest` a bit confusing so now I'm wondering whether `default-for` can replace it altogether?

sdeleuze: I pushed the changes in another commit.

`latest: true` translate to `"default-for": ".*"` with lower priority. I also feel there is a potential to simplify the configuration, but that could easily trigger regressions, so I think I would be for just introducing `default-for` for now, fix the Hibernate issue, and get feedback as we extend how much we use it in metadata.

melix: `latest` is exactly this: this is the latest available version of metadata, which was, until this PR, used at the fallback in case we have a version we don't know about. `default-for` is a better substitute, but it doesn't address the problem whenever a version doesn't match any `default-for`. We need a way to figure out what is the "most reasonable" version of metadata to use, and `latest` is designed for this.

sdeleuze:
@aclement Could you please give your +1 for merging this PR?

aclement:
looks good to me.

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
```diff
@@ -101,15 +103,18 @@ public Optional<DirectoryConfiguration> findLatestConfigurationFor(String groupI
...
-     * Returns the latest configuration directory for the requested artifact.
+     * Returns the matching configuration directory for the requested artifact.
      *
      * @param groupId the group ID of the artifact
      * @param artifactId the artifact ID of the artifact
+     * @param version the version of the artifact
      * @return a configuration directory, or empty if no configuration directory is available
      */
     @Override
     public Optional<DirectoryConfiguration> findLatestConfigurationFor(String groupId, String artifactId, String version) {
```
fniephaus, line 114: The name is no longer accurate. Not sure if we want to change it?

fniephaus (approved):
LGTM, thanks for the PR!

melix (approved):

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/index/artifacts/SingleModuleJsonVersionToConfigDirectoryIndex.java
sdeleuze: Could you please elaborate?

fniephaus: find**Latest**ConfigurationFor, shouldn't it be something like findMatchingConfigurationFor now?

melix:
Don't forget release notes :)

Issue #441, opened 2023-05-12
generateResourceConfig not working


blackstar-baba:
**Describe the bug**
I am clone this demo https://github.com/graalvm/native-build-tools/tree/master/samples/java-application-with-resources and change some plugin versions,  when i am running the result after build native image,  throw npe like this:
```
***** native-with-resource % ./target/example-app 
Exception in thread "main" java.lang.NullPointerException
        at java.base@17.0.5/java.io.Reader.<init>(Reader.java:168)
        at java.base@17.0.5/java.io.InputStreamReader.<init>(InputStreamReader.java:76)
        at org.graalvm.demo.Application.getMessages(Application.java:11)
        at org.graalvm.demo.Application.main(Application.java:16)
```

**To Reproduce**

my source code: https://github.com/blackstar-baba/native-with-resource

run this command: `mvn -Pnative clean package`

**System Info (please complete the following information):**
 - OS: [`MACOS AMD64`]
 - GraalVM Version [`22.0 CE`]
 - Java Version [`17`]
 - Plugin version [`native-gradle-plugin:0.9.22`]

**Additional context**

blackstar-baba:
additional context:
`target/native/generated/generatedResourceConfig/resource-config.json` includes is empty, like this:
```
{
  "resources" : {
    "includes" : [ ],
    "excludes" : [ ]
  },
  "bundles" : [ ]
}
```

blackstar-baba:
I looked this test case: JavaApplicationWithResourcesFunctionalTest and i used this command: `mvn -Pnative  -DskipTests -Dresources.autodetection.enabled=true clean package` which can working. I think we need some docs for how to use this.

blackstar-baba:
I think we can set  property `resources.autodetection.enabled` equals `true` default.

melix:
The Gradle plugin also uses `false` as a default value, which I think is a _safe_ default, especially because if you start using the agent in combination, you can easily get duplicate resources (both defined in what we infer and what is found by the agent). So I'd stick with `false`.

blackstar-baba:
How about update this demo: `java-application-with-resources` ?  it can't working now.

blackstar-baba:
It can working if add codes in `pom.xml` properties like this: 
```
 <resources.autodetection.enabled>true</resources.autodetection.enabled>
```

melix:
[...]

PR #442, opened 2023-05-12
Set resources autodetection value to true

blackstar-baba:


blackstar-baba:
Fix issue #441

graemerocher:
Don't think this is an issue, this is by design. You have to enable it explicitly

PR #445, opened 2023-06-02
Discover tests with `ClassLoader` other than `ImageClassLoader`

ziyilin:
`JUnitPlatformFeature` discovers tests before analysis. The discovery process will initialize some of the test classes. As they are loaded by the `ImageClassloader`, the initialization will cause eager class initialization error during native image building time.

This PR uses a different classloader than the `ImageClassLoader` in the `JUnitPlatformFeature` to make sure the test class initialization at discovery time won't affect the native image class initialization policy.

==updated July 7th 2023 ==
This PR launches a separated Java process to run the test discovery if `-DisolateTestDiscovery=true` is set (it is set by default), and goes to the original way otherwise. The reason is the initialization of test class may initialize some JDK classes that are loaded by null classloader. In this case, classloader isolation is not sufficient. 

The attached demo shows the issue. 
The GraalVM version is java11-22.3.1.
Simply unzip the file and run `mvn -Pnative test` and `mvn -Pnative test -DisolateTestDiscovery=false` to see the difference.

[native-build-tool-demo.zip](https://github.com/graalvm/native-build-tools/files/11675312/native-build-tool-demo.zip)

### DIFF
--- a/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
+++ b/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
@@ -42,38 +42,21 @@
 package org.graalvm.junit.platform;
 
 import org.graalvm.junit.platform.config.core.PluginConfigProvider;
+import org.graalvm.nativeimage.ImageInfo;
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.hosted.Feature;
 import org.graalvm.nativeimage.hosted.RuntimeClassInitialization;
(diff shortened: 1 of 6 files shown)


melix:
I am not sure this change is safe. Probably @sbrannen has more expertise than I do on this particular aspect. I'm in particular wondering about the use of the context classloader.

melix:
We're going to have to revert this PR as it breaks builds with build time initialization happening with JUnit classes.

ziyilin:
@melix Can you provide a test case？

melix:
You simply have to run `./gradlew :native-gradle-plugin:funTest` and every single test which uses `nativeTest` is going to fail.

PR #495, opened 2023-09-01
Relax GraalVM version check for dev versions

dnestoro:


common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
```diff
@@ -122,7 +122,10 @@ public static void checkVersion(String requiredVersion, String versionToCheck) {
         if (versionToCheck.contains("GraalVM Runtime Environment")) {
             return; // later than 22.3.1 (e.g., GraalVM for JDK 17 / GraalVM for JDK 20)
         }
-        if (versionToCheck.startsWith("GraalVM dev") || versionToCheck.startsWith("native-image dev")) {
+        if (!versionToCheck.startsWith("GraalVM") && !versionToCheck.startsWith("native-image") && !versionToCheck.startsWith("Oracle GraalVM")) {
```
fniephaus, line 125: Let's start with the latest format first (CE and Oracle GraalVM print `native-image ...`  now:

```suggestion
        if (!versionToCheck.startsWith("native-image") && !versionToCheck.startsWith("GraalVM") && !versionToCheck.startsWith("Oracle GraalVM")) {
```

fniephaus (approved):
LGTM

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
```diff
@@ -122,7 +122,10 @@ public static void checkVersion(String requiredVersion, String versionToCheck) {
...
+            throw new IllegalArgumentException("Invalid version " + versionToCheck + ", should start with GraalVM, Oracle GraalVM or native-image");
+        }
+        if (versionToCheck.contains("dev")) {
```
fniephaus: Oh, I think we can simplify this as follows:

```suggestion
        if ((versionToCheck.startsWith("native-image") && versionToCheck.contains("-dev+")) ||
                 versionToCheck.startsWith("GraalVM dev")) { /* For GraalVM 22.3 and earlier */
            return;
        }
```

fniephaus (changes requested):

melix (approved):

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
dnestoro: This is a bit stricter than it was in last commit, but as long as we don't introduce some new format, this will work.

fniephaus: I didn't like `versionToCheck.contains("dev")` in particular. What if, for some reason, a word like "developer" becomes part of the version string?

common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
```diff
@@ -122,7 +122,8 @@ public static void checkVersion(String requiredVersion, String versionToCheck) {
         if (versionToCheck.contains("GraalVM Runtime Environment")) {
             return; // later than 22.3.1 (e.g., GraalVM for JDK 17 / GraalVM for JDK 20)
         }
-        if (versionToCheck.startsWith("GraalVM dev") || versionToCheck.startsWith("native-image dev")) {
+        if ((versionToCheck.startsWith("native-image") && versionToCheck.contains("-dev+")) ||
```
christianwimmer, line 125: I would change `-dev+` to just `dev`. We should try to be future-proof, i.e., we don't want to adapt this change any time the version string of GraalVM changes.

Maybe even reduce the whole check to just `if (versionToCheck.contains("dev") {`.

fniephaus (approved):
LGTM

PR #496, opened 2023-09-01
Prepare NBT for release of GraalVM for JDK 21.

fniephaus:
This PR changes NBT so that it only uses public / stable Native Image API, which otherwise would require uses to unlock experimental options (see https://github.com/oracle/graal/issues/7105 for context) in the next release of GraalVM.

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
@@ -67,23 +67,26 @@ public class NativeImageCommandLineProvider implements CommandLineArgumentProvid
 
     private final Provider<NativeImageOptions> options;
     private final Provider<String> executableName;
-    private final Provider<String> outputDirectory;
     private final Provider<String> workingDirectory;
+    private final Provider<String> outputDirectory;
     private final Provider<RegularFile> classpathJar;
     private final Provider<Boolean> useArgFile;
+    private final Provider<Integer> majorJDKVersion;
 
     public NativeImageCommandLineProvider(Provider<NativeImageOptions> options,
                                           Provider<String> executableName,
                                           Provider<String> workingDirectory,
                                           Provider<String> outputDirectory,
                                           Provider<RegularFile> classpathJar,
-                                          Provider<Boolean> useArgFile) {
+                                          Provider<Boolean> useArgFile,
+                                          Provider<Integer> majorJDKVersion) {
         this.options = options;
         this.executableName = executableName;
         this.workingDirectory = workingDirectory;
         this.outputDirectory = outputDirectory;
         this.classpathJar = classpathJar;
         this.useArgFile = useArgFile;
+        this.majorJDKVersion = majorJDKVersion;
     }
 
     @Nested
@@ -119,13 +122,15 @@ public class NativeImageCommandLineProvider implements CommandLineArgumentProvid
         appendBooleanOption(cliArgs, options.getVerbose(), "--verbose");
         appendBooleanOption(cliArgs, options.getSharedLibrary(), "--shared");
         appendBooleanOption(cliArgs, options.getQuickBuild(), "-Ob");
-        appendBooleanOption(cliArgs, options.getRichOutput(), "-H:+BuildOutputColorful");
+        appendBooleanOption(cliArgs, options.getRichOutput(), majorJDKVersion.getOrElse(-1) >= 21 ? "--color" : "-H:+BuildOutputColorful");
         appendBooleanOption(cliArgs, options.getPgoInstrument(), "--pgo-instrument");
 
+        String targetOutputPath = getExecutableName().get();
         if (getOutputDirectory().isPresent()) {
-            cliArgs.add("-H:Path=" + getOutputDirectory().get());
+            targetOutputPath = getOutputDirectory().get() + File.separator + targetOutputPath;
         }
-        cliArgs.add("-H:Name=" + getExecutableName().get());
+        cliArgs.add("-o");
+        cliArgs.add(targetOutputPath);
 
         options.getSystemProperties().get().forEach((n, v) -> {
             if (v != null) {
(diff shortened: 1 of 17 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -189,7 +189,7 @@ private List<String> buildActualCommandLineArgs() {
                 // a mapped value before the task was called, when we are actually calling it...
                 getProviders().provider(() -> getOutputDirectory().getAsFile().get().getAbsolutePath()),
                 getClasspathJar(),
-                getUseArgFile()).asArguments();
+                getUseArgFile()).setMajorJDKVersion(majorJDKVersion).asArguments();
```
fniephaus, line 192: @melix I'm not sure how to do this right. Adding another Provider means we are changing the API. Maybe you have a better suggestion.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
```diff
@@ -119,13 +125,15 @@ public List<String> asArguments() {
         appendBooleanOption(cliArgs, options.getVerbose(), "--verbose");
         appendBooleanOption(cliArgs, options.getSharedLibrary(), "--shared");
         appendBooleanOption(cliArgs, options.getQuickBuild(), "-Ob");
-        appendBooleanOption(cliArgs, options.getRichOutput(), "-H:+BuildOutputColorful");
+        appendBooleanOption(cliArgs, options.getRichOutput(), majorJDKVersion >= 21 ? "--color" : "-H:+BuildOutputColorful");
```
fniephaus, line 128: Note: this option is only used in the Gradle plugin. Since Maven doesn't need it, it also doesn't need to check `native-image --version` unconditionally.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -203,18 +203,18 @@ public void exec() {
         NativeImageOptions options = getOptions().get();
         GraalVMLogger logger = GraalVMLogger.of(getLogger());
 
-        List<String> args = buildActualCommandLineArgs();
```
fniephaus, line 206: I'm not 100% sure this can be re-ordered. `buildActualCommandLineArgs()` finalizes options, not sure what that means.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
```diff
@@ -67,10 +67,11 @@ public class NativeImageCommandLineProvider implements CommandLineArgumentProvid
 
     private final Provider<NativeImageOptions> options;
     private final Provider<String> executableName;
-    private final Provider<String> outputDirectory;
     private final Provider<String> workingDirectory;
+    private final Provider<String> outputDirectory;
     private final Provider<RegularFile> classpathJar;
     private final Provider<Boolean> useArgFile;
+    private int majorJDKVersion = -1;
```
melix, line 74: :x: This should be a final field. Even if it's an API breaking change (the class is used in the Micronaut plugin).

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
melix: It is ok to change the API. As far as I know, the only consumer is the Micronaut plugin. We've already broken the contract several times, but it's fine, as long as figuring out what value to provide is easy.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
melix: Finalizing is a safety net. It means that the user won't be allowed to change the value of the options after this call is made.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -240,19 +240,18 @@ public void exec() {
         }
     }
 
-    private void checkRequiredVersionIfNeeded(File executablePath, NativeImageOptions options) {
-        if (!options.getRequiredVersion().isPresent()) {
-            return;
-        }
+    private int checkVersionAndGetMajorJDKVersion(File executablePath, NativeImageOptions options) {
```
melix, line 243: suggestion: Can you make this a public method in `NativeImageUtils` (you'll probably need to make `NativeImageUtils` public too)? Otherwise I will have to reimplement the very same logic in the Micronaut Gradle plugin.

melix (changes requested):
Looks good overall, except for the change to the command line provider. We should simply add another provider as a constructor argument. It's a breaking change, but I think it's fine (the only known consumer is the Micronaut Gradle plugin). This will be easier to maintain.

Also I need to have a method which computes the JDK version for the Micronaut plugin too, so having this code shared would be simpler.

Thanks!

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
fniephaus: This already is a public method of the public `NativeImageUtils` :)

melix: Sorry I'm confused. In `NativeImageUtils` I can see `getMajorJDKVersion` but not the whole `checkVersionAndGetMajorJDKVersion` which will need to be replicated in Micronaut.

melix (approved):

fniephaus:
@sschuberth is that still the case with 0.9.28? Which GraalVM distribution are you using? /cc @dnestoro

sschuberth:
I cannot reproduce this anymore with plugin version 0.9.28 and GraalVM runtime version 21.0.1+12-jvmci-23.1-b22. So I was probably just too impatient WRT the version this got fixed in...

PR #502, opened 2023-09-10
Fix Windows path escape

fuqiuluo:


fniephaus:
Thanks for the PR, we already addressed this via https://github.com/graalvm/native-build-tools/pull/515.

PR #506, opened 2023-09-12
Bug fixes and improvements for GraalVM for JDK 21.

fniephaus:
Note: We need to run with `--strict-image-heap` enabled on GraalVM dev builds / GraalVM for JDK 21 in the future. The current dev build does not have the option, but has `--strict-image-heap` enabled by default.

### DIFF
--- a/.github/workflows/test-native-gradle-plugin.yml
+++ b/.github/workflows/test-native-gradle-plugin.yml
@@ -48,6 +48,8 @@ jobs:
   functional-testing-gradle-plugin-dev:
     name: "Functional testing (GraalVM Dev Build)"
     runs-on: ${{ matrix.os }}
+    env:
+      IS_GRAALVM_DEV_BUILD: 'true'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test-native-gradle-plugin.yml
+++ b/.github/workflows/test-native-gradle-plugin.yml
@@ -45,6 +45,37 @@ jobs:
         with:
           name: unit-tests-results
           path: native-gradle-plugin/build/reports/tests/test/
+  functional-testing-gradle-plugin-dev:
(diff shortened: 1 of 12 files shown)


olpaw:
AFAIU we do not want end-users to use the version of NBTs that has this change. This is just for our own testing until time is right to switch to a GraalVM native-image where `--strict-image-heap` is the default.

fniephaus:
> AFAIU we do not want end-users to use the version of NBTs that has this change. This is just for our own testing until time is right to switch to a GraalVM native-image where --strict-image-heap is the default. Correct?

We want users of GraalVM for JDK 21 to adopt `--strict-image-heap` as soon as possible, so support for it in NBT is necessary if you ask me. On the other hand, the additional build-time init policies should still work as before without `--strict-image-heap`: they are just complementary policies that we didn't know were actually needed. But yes, we should probably test this to be on the safe side.

fniephaus:
> But yes, we should probably test this to be on the safe side.

Here's one way to test this now: https://github.com/graalvm/native-build-tools/commit/f7d1eb6d73d5a62ab1bb956725f10122307cfd3b

olpaw:
> > But yes, we should probably test this to be on the safe side.
> 
> Here's one way to test this now: [f7d1eb6](https://github.com/graalvm/native-build-tools/commit/f7d1eb6d73d5a62ab1bb956725f10122307cfd3b)

Ok, this proves that the new JUnit init-at-build-/run-time condigs have no side effect when `--strict-image-heap` is off.

Sounds safe enough for me.

olpaw (approved):

fniephaus:
It seems this is blocking the Micronaut Gradle plugin release: https://github.com/micronaut-projects/micronaut-starter/pull/2045#issuecomment-1717275538

PR #482, opened 2023-08-10
Add concurrency with test array in matrix

dnestoro:

PR #524, opened 2023-10-13
Re-enable Windows builds.

### LINKED ISSUE #520: Enable Windows Tests in GitHub Actions
Currently we are not running tests on windows.

There is a GitHub action (that already exists) which should be enabled again. 

**Additional context**
According to [this issue](https://github.com/graalvm/native-build-tools/issues/511) it is expected that action will fail. Solving this issue properly (enabling action again and fixing bugs that will be reported) will probably solve that problem as well.

fniephaus:
Fixes #520

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
@@ -190,7 +190,8 @@ public abstract class BuildNativeImageTask extends DefaultTask {
                 getProviders().provider(() -> getOutputDirectory().getAsFile().get().getAbsolutePath()),
                 getClasspathJar(),
                 getUseArgFile(),
-                getProviders().provider(() -> majorJDKVersion)).asArguments();
+                getProviders().provider(() -> majorJDKVersion),
+                getProviders().provider(() -> "plain".equals(getProject().getGradle().getStartParameter().getConsoleOutput()))).asArguments();
     }
 
     // This property provides access to the service instance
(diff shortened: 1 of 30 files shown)


dnestoro (approved):

dnestoro:
This is a good fix for the current state of workflows, but after I merge [my PR](https://github.com/graalvm/native-build-tools/pull/482) we can remove this entire job and just add `windows-latest` as a variable in [matrix](https://github.com/graalvm/native-build-tools/pull/482/files#diff-0b287022bbc4691b9caa459508d2e1c6842c38b98b7b86dc472f3508d3773629R5)

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -190,7 +190,8 @@ private List<String> buildActualCommandLineArgs(int majorJDKVersion) {
                 getProviders().provider(() -> getOutputDirectory().getAsFile().get().getAbsolutePath()),
                 getClasspathJar(),
                 getUseArgFile(),
-                getProviders().provider(() -> majorJDKVersion)).asArguments();
+                getProviders().provider(() -> majorJDKVersion),
+                getProviders().provider(() -> "plain".equals(getProject().getGradle().getStartParameter().getConsoleOutput()))).asArguments();
```
melix, line 194: :x: Calling `getProject()` at execution time is not compatible with Gradle's configuration cache. This must be changed, and the easiest is to copy the value of the `getConsoleOutput()` directly in the task constructor, so that you don't have to query that at execution time.

melix (changes requested):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
fniephaus: Thanks for catching! I've fixed this, please take another look.

melix (approved):

PR #531, opened 2023-10-23
Reduce Time Required for Each Test Execution

dnestoro:
This PR should update tests so that they run with `-Ob` flag


melix (changes requested):
I think you need to update the samples to remove `-Ob`, don't you?

fniephaus (approved):
LGTM

Issue #547, opened 2023-12-11
Execution build-native of goal org.graalvm.buildtools:native-maven-plugin:0.9.28:compile-no-fork failed.: FileSystemAlreadyExistsException


kohlschuetter:
**Describe the bug**

Execution of `compile-no-fork` fails if a `FileSystem` for a dependency jar has already been opened by another plugin.

**To Reproduce**

```
mvn clean package
```
(Sorry, can't share the offending plugin right now, but a fix will be provided)

**Expected behavior**

Execution of the `compile-no-fork` goal should succeed.

**Logs**
````
```
[ERROR] Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.9.28:compile-no-fork (build-native) on project dumbo-jacline-helloworld: Execution build-native of goal org.graalvm.buildtools:native-maven-plugin:0.9.28:compile-no-fork failed.: FileSystemAlreadyExistsException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.9.28:compile-no-fork (build-native) on project dumbo-jacline-helloworld: Execution build-native of goal org.graalvm.buildtools:native-maven-plugin:0.9.28:compile-no-fork failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:73)
[...]

PR #548, opened 2023-12-11
Fix compile-no-fork with existing jar FileSystems (#547)

kohlschuetter:
FileSystems.newFileSystem on a dependency jar fails with a FileSystemAlreadyExistsException if another Maven plugin has already created it.

Provide a wrapper that handles the exception, trying FileSystems.getFileSystem in that case.


kohlschuetter:
OCA signed.

fniephaus (approved):
LGTM, thanks for contributing!

PR #564, opened 2024-01-30
Bump GraalVM version to 23.0.0

dnestoro:


melix:
Shouldn't we use the latest, 23.1.2?

fniephaus:
> Shouldn't we use the latest, 23.1.2?

That would mean we are also dropping GraalVM for JDK 17 support, which we want to keep for a while I think.

dnestoro:
@ezzarghili suggested `23.0.2`

Issue #575, opened 2024-02-19
The junit-native does not support org.junit.rules.ExpectedException.


tzezula:
**Describe the bug**
The native junit test fails if the `org.junit.rules.ExpectedException` or `org.hamcrest.core.Every` is used by the test.
```
 MethodSource [className = 'org.me.MainTest', methodName = 'testExpectedException', methodParameterTypes = '']
    => java.lang.Error: Cannot determine correct type for matchesSafely() method.
       org.hamcrest.internal.ReflectiveTypeFinder.findExpectedType(ReflectiveTypeFinder.java:49)
       org.hamcrest.TypeSafeMatcher.<init>(TypeSafeMatcher.java:40)
       org.hamcrest.TypeSafeMatcher.<init>(TypeSafeMatcher.java:22)
```

**To Reproduce**
Uncompress the attached maven project and run `mvn -Pnative package`.

**Expected behavior**
The tests from the attached project should also pass in native mode. Currently, they are failing.

**Logs**
See comments in the `src/test/java/org/me/MainTest.java`

**System Info:**
 - OS: macos
 - GraalVM Version 23.1.2
 - Java Version 21
 - Plugin version `native-maven-plugin:0.10.0`

**Additional context**
The issue arises from `org.hamcrest.TypeSafeMatcher` and `org.hamcrest.TypeSafeDiagnosingMatcher` utilizing `ReflectiveTypeFinder` to locate the `matchesSafely` method in their subclasses during image execution. This problem can be resolved by employing a reflection configuration file. Although the `reflection.json` file is included in the test project, it is commented out in the `pom.xml`. Since the reflective access targets JUnit types, it should ideally be handled by the `JUnitPlatformFeature` rather than by the application itself.

One approach is to eliminate type safety, similar to what is implemented in the [internal GraalVM JUnit support](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.junit/src/com/oracle/svm/junit/Target_org_hamcrest_internal_ReflectiveTypeFinder.java). However, this method requires a substitution, which is not a public API.

Alternatively, we can register all used subclasses of `TypeSafeMatcher` and `TypeSafeDiagnosingMatcher` for reflective access to all declared methods.

tzezula:
[Test project](https://github.com/graalvm/native-build-tools/files/14335324/project.zip)

tzezula:
[...]

PR #576, opened 2024-02-19
[575] junit-native does not support org.junit.rules.ExpectedException.

tzezula:
Fixies issue #575 by registering all used subclasses of `TypeSafeMatcher` and `TypeSafeDiagnosingMatcher` for reflective access to all declared methods.


common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -93,6 +96,46 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
         Launcher launcher = LauncherFactory.create();
         TestPlan testplan = discoverTestsAndRegisterTestClassesForReflection(launcher, selectors);
         ImageSingletons.add(NativeImageJUnitLauncher.class, new NativeImageJUnitLauncher(launcher, testplan));
+
+        Method registerAllDeclaredMethods = findMethodOrNull(RuntimeReflection.class, "registerAllDeclaredMethods", Class.class);
```
tzezula, line 100: The required `RuntimeReflection#registerAllDeclaredMethods` method was introduced in `graalvm-23.0`. However, since the `native-build-tools` are currently utilizing `graalvm-22.3.5`, we employ the method reflectively only when it is available.

tzezula: @fniephaus Could you assist me in identifying additional reviewers?
Thank you!

fniephaus:
@tzezula I added more reviewers. Could you please add a test or test project for this? Not sure whether [this](https://github.com/graalvm/native-build-tools/issues/575#issuecomment-1952929799) will do the trick. Instead of adding it as a new [sample](https://github.com/graalvm/native-build-tools/tree/master/samples), maybe you could extend an existing one?

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
dnestoro: @tzezula actually, `native-build-tools` are currently using graalvm `23.0.2`. This change was introduced in [this PR](https://github.com/graalvm/native-build-tools/pull/565) and it is a part of [our latest release](https://github.com/graalvm/native-build-tools/releases/tag/0.10.1) .

tzezula: Perfect, I will rebase the PR and remove the reflection.
Thanks!

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -93,6 +96,46 @@ public void beforeAnalysis(BeforeAnalysisAccess access) {
...
+            if (typeSafeMatcher != null || typeSafeDiagnosingMatcher != null) {
+                BiConsumer<DuringAnalysisAccess, Class<?>> registerMatcherForReflection = (a, c) -> {
+                    try {
+                        registerAllDeclaredMethods.invoke(null, c);
+                    } catch (ReflectiveOperationException e) {
+                        throw new RuntimeException(e);
+                    }
+                };
+                if (typeSafeMatcher != null) {
+                    access.registerSubtypeReachabilityHandler(registerMatcherForReflection, typeSafeMatcher);
```
vjovanov, line 115: How do we guarantee this does not affect user code by registering their classes?

tzezula: @vjovanov: Good question. I believe it's acceptable to register even the user classes. This is because subclasses of `org.hamcrest.TypeSafeMatcher` and `org.hamcrest.TypeSafeDiagnosingMatcher` that utilize the default constructor rely on `Class.getDeclaredMethods()` to function properly. You can refer to [TypeSafeMatcher#TypeSafeMatcher()](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest/src/main/java/org/hamcrest/TypeSafeMatcher.java#L25), which indirectly invokes [ReflectiveTypeFinder#findExpectedType()](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest/src/main/java/org/hamcrest/internal/ReflectiveTypeFinder.java#L43). Ideally, we would only mark the `matchesSafely` method for `getDeclaredMethods`, but unfortunately, this is not possible.

tzezula:
@fniephaus I've added the [test](https://github.com/graalvm/native-build-tools/pull/576/commits/14a84eab815f2143db14620ee558cc519ae77d3c)

dnestoro (approved):

dnestoro:
The content of this PR will be included together with main JUnit refactoring PR ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R205)) to avoid additional (rebase and refactoring) work on this PR. Once we merge that PR, we can close this one

Issue #579, opened 2024-03-06
`metadata-copy` should follow recommendation


fniephaus:
[On the GraalVM website](https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildConfiguration/), we recommend to put metadata into directory following this pattern:

> foo.jar has its configurations in META-INF/native-image/foo_groupID/foo_artifactID
> bar.jar has its configurations in META-INF/native-image/bar_groupID/bar_artifactID

It seems that the `metadata-copy` copies metadata into just [`/META-INF/native-image`](https://github.com/graalvm/native-build-tools/blob/c791bf6d32067664d87d58938f771f8a1cf2b42d/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/MetadataCopyMojo.java#L70).
Would be great if we could adjust it so that it follows our recommendation.

vjovanov:
@dnestoro is this issue fixed?

dnestoro:
Kind of... for gradle we are just suggesting (in our docs) where should user copy metadata files (and we won't have default value for it in the plugin implementation), while maven has actual default directory for metadataCopy task. 

So this issue can be closed

PR #580, opened 2024-03-06
Update Default Target Directory for MetadataCopy Task

dnestoro:
As described in [this issue](https://github.com/graalvm/native-build-tools/issues/579) we should specify default output directory for metadataCopy task according to the recommendation from [GraalVM website](https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildConfiguration/) 

Note: currently we don't have default value for the Gradle plugin, while we do have one for the Maven

### DIFF
--- a/docs/src/docs/snippets/gradle/groovy/build.gradle
+++ b/docs/src/docs/snippets/gradle/groovy/build.gradle
@@ -99,7 +99,7 @@ graalvmNative {
         // Copies metadata collected from tasks into the specified directories.
         metadataCopy {
             inputTaskNames.add("test") // Tasks previously executed with the agent attached.
-            outputDirectories.add("src/main/resources/META-INF/native-image")
+            outputDirectories.add("src/main/resources/META-INF/native-image/<groupId>/<artifactId>/") // Replace <groupId> and <artifactId> with GAV coordinates of your project
             mergeWithExisting = true // Instead of copying, merge with existing metadata in the output directories.
         }
 
--- a/docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
+++ b/docs/src/docs/snippets/gradle/kotlin/build.gradle.kts
@@ -100,7 +100,7 @@ graalvmNative {
         // Copies metadata collected from tasks into the specified directories.
         metadataCopy {
             inputTaskNames.add("test") // Tasks previously executed with the agent attached.
-            outputDirectories.add("src/main/resources/META-INF/native-image")
+            outputDirectories.add("src/main/resources/META-INF/native-image/<groupId>/<artifactId>/") // Replace <groupId> and <artifactId> with GAV coordinates of your project
             mergeWithExisting.set(true) // Instead of copying, merge with existing metadata in the output directories.
         }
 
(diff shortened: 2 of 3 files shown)


docs/src/docs/snippets/gradle/groovy/build.gradle
```diff
@@ -99,7 +99,7 @@ graalvmNative {
         // Copies metadata collected from tasks into the specified directories.
         metadataCopy {
             inputTaskNames.add("test") // Tasks previously executed with the agent attached.
-            outputDirectories.add("src/main/resources/META-INF/native-image")
+            outputDirectories.add("src/main/resources/META-INF/native-image/<groupId>/<artifactId>/") // Replace <groupId> and <artifactId> with GAV coordinates of your project
```
fniephaus, line 102: Is the trailing forward slash correct? We used to not have it.

melix (changes requested):
The original issue as well as this PR is confusing. The issue says that `metadata-copy` copies to `META-INF/native-image`, but there's no such default in Gradle. I see that there's one in Maven, but I'm wondering if it's a good idea.

There was a reason for not adding a default in the Gradle version (which I think predates the Maven one), which is to avoid having overlapping outputs (basically preventing the build from being up-to-date), but also because we don't really know what was used by the agent to collect metadata, so adding this explicitly could overwrite legit metadata.

So I think we should either:
- remove the default from the Maven plugin, then document how to set the target directory for metadata copy
- or change the Gradle plugin so that it also introduces a default value (which again I'm not sure is a good idea)

Currently the PR mixes a bit the 2, it doesn't really fix the consistency issue between both plugins or the documentation.

docs/src/docs/snippets/gradle/groovy/build.gradle
dnestoro: Since `src/main/resources/META-INF/native-image/<groupId>/<artifactId>/` is directory, I don't see any difference if we use the trailing forward slash or not. It should work in any case, and I think it doesn't change anything. We can remove it if it is necessary

dnestoro:
> so adding this explicitly could overwrite legit metadata

But it will override metadata only if we explicitly call `metadataCopy` task? Otherwise the metadata will be stored as usual into directory specified in `agent-output-dir`. The point here is that we should make `metadataCopy` task coping metadata to the directory we are suggesting in [our docs](https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildConfiguration/). If I am not missing something.

> So I think we should either:
> - remove the default from the Maven plugin, then document how to set the target directory for metadata copy
> - or change the Gradle plugin so that it also introduces a default value (which again I'm not sure is a good idea)

Yes, I agree that we should sync Maven and Gradle solutions.

fniephaus:
> The original issue as well as this PR is confusing

When I opened the ticket, I was not aware that this is another part where our Maven and Gradle plugins diverge. Yes, we should fix that and align those. This PR is still an improvement: while it doesn't fix the misalignment, it improves the recommendation. Does it make sense to merge this and align the plugins in a follow up (after a discussion it seems)?

melix (approved):

PR #589, opened 2024-04-01
Delete old stale args file

n0tl3ss:
Problem:
Copying args file into Docker with something similar to 
```COPY *.args /home/app/graalvm-native-image.args``` causes problem because there might be more then one generated args file. This is required because `convertToArgsFile` creates temporary file that contains contains random integer in the name.

Solution:
- Maven write-args-file task should delete old stale args file before generating a new one.

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/WriteArgsFileMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/WriteArgsFileMojo.java
@@ -47,7 +47,11 @@ import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.graalvm.buildtools.utils.NativeImageUtils;
 
 import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
+import java.util.stream.Stream;
 
 /**
  * Persists the arguments file to be used by the native-image command. This can be useful in situations where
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/WriteArgsFileMojo.java [hunk 1]
```diff
@@ -67,6 +71,24 @@ public class WriteArgsFileMojo extends NativeCompileNoForkMojo {
     @Override
     public void execute() throws MojoExecutionException {
         List<String> args = getBuildArgs();
+
+        getLog().info("Cleaning old native image build args");
```
alvarosanchez, line 75: This should probably be debug

alvarosanchez (approved):

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/WriteArgsFileMojo.java
```diff
@@ -67,6 +71,24 @@ public class WriteArgsFileMojo extends NativeCompileNoForkMojo {
...
+
+        try (Stream<Path> listStream = Files.list(outputDirectory.toPath())) {
+            listStream.map(path -> path.getFileName().toString())
+                    .filter(f -> f.startsWith("native-image") && f.endsWith("args"))
+                    .map(outputDirectory.toPath()::resolve)
+                    .forEach(file -> {
+                        try {
+                            Files.delete(file);
+                        } catch (IOException e) {
+                            throw new RuntimeException(e);
```
alvarosanchez, line 85: Also, you should probably throw `MojoExecutionException` here

alvarosanchez (changes requested):

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/WriteArgsFileMojo.java
n0tl3ss: in forEach it only allows `RuntimeException` to be thrown

alvarosanchez (approved):

PR #590, opened 2024-04-10
fix class path directroy analyzer

n0tl3ss:
- visitFile method should also consider `ignoreExistingResourcesConfig` before it `TERMINATE`s walk.


alvarosanchez (approved):

fniephaus (approved):
lgtm

Issue #334, opened 2022-10-13
Maven Native Image Plugin: Support for building java modules for native image, including support for junit


Rottinator:
Hello there,

**Is your feature request related to a problem? Please describe.**
I've converted an application of mine into java modules and tried to build it via native maven plugin. The problem: Native maven plugin seams not support to build native image automatically via the module parameters. So i have to add all project modules and maven dependencies manually to the --module-path parameter which isn't very comfortable in large projects.
For executing junit tests via native image profile i did not find the correct parameters on my own for getting the tests running with modules in native image.

**Describe the solution you'd like**
I would suggest implement support for automatically detecting that the current project, that is built, is a java module and automatically feed the native-image command with the module parameters, like other maven plugins do.
For example running unit tests with surefire automatically runs them in "module mode" without any additional parameters for the plugin, it's enough the have a module-info.java present in the project.

I've already checked the documentation here [https://graalvm.github.io/native-build-tools/latest/maven-plugin.html](https://graalvm.github.io/native-build-tools/latest/maven-plugin.html) but didn't found any information for building module projects with native maven plugin.

Hopefully it's understandable what my point is, if you need additional information or a sample application i can provide it.

Regards
Christoph

fniephaus:
We'd like to properly support the module path, but as far as I know, it is poorly supported by Maven and Gradle and causes a lot of problems. Can you build and run your modularized Java application with Maven just fine when not using the native build tools?

Rottinator:
> We'd like to properly support the module path, but as far as I know, it is poorly supported by Maven and Gradle and causes a lot of problems. Can you build and run your modularized Java application with Maven just fine when not using the native build tools?

Yes, it was also possible to build with native-build-tools, but i needed to feed in all paths to maven dependencies manually, which isn't that simple, when you have different environments like Windows Dev-Machines and Linux-Build-Servers, cause of different paths to the maven repo and path separators etc.

As a final result i wrote my own maven plugin, that obtains all dependencies etc. and feeding native-image with the required parameters to build it on the module path.

fniephaus:
> As a final result i wrote my own maven plugin, that obtains all dependencies etc. and feeding native-image with the required parameters to build it on the module path.

Please consider upstreaming the work that you have done. And if you have any other suggestions how the NBTs could support the module path, we are all ears.

Maran23:
[...]

PR #607, opened 2024-06-20
334 - Fix module path and image classpath for native image generation

loiclefevre:
This PR should fix the issue related in #334.

This allows to simplify native image generation via Maven with such build args inside the pom.xml file:
```xml
                    <plugin>
                        <version>${native.maven.plugin.version}</version>
                        <executions>
                            <execution>
                                <id>build-native</id>
                                <goals>
                                    <goal>compile-no-fork</goal>
                                </goals>
                                <phase>package</phase>
                            </execution>
                            <execution>
                                <id>test-native</id>
                                <goals>
                                    <goal>test</goal>
                                </goals>
                                <phase>test</phase>
                            </execution>
                        </executions>
                        <configuration>
                            <verbose>true</verbose>
                            <skip>false</skip>
                            <imageName>${imageName}</imageName>
                            <fallback>false</fallback>
                            <agent>
                                <enabled>false</enabled>
                            </agent>
                            <buildArgs>
                                --module-path target/connectme-1.0.0.jar
                                -Ob
                                -march=native
                                --initialize-at-build-time=com.oracle.connect/com.oracle.connect.Main
                                --add-modules javafx.controls
                                --add-modules javafx.fxml
                                --add-modules com.oracle.connect
                                --module com.oracle.connect/com.oracle.connect.Main
                            </buildArgs>
                        </configuration>
                    </plugin>
```

The modification includes: 
- detecting properly if a dependency is a module (analyze jar file and look for the presence of `/module-info.class`
- if it is a module, then populate `--module-path` arg list else populate `-cp` arg list

### DIFF
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -160,6 +160,14 @@ Build Configuration]. It is also possible to customize the plugin within a
 ----
 <verbose>true</verbose>
 ----
+`<experimentalInferModulePath>`::
+If you want to enable module path inference during native-image building supply the following in the configuration of the plugin:
+[source,xml]
+----
+<experimentalInferModulePath>true</experimentalInferModulePath>
+----
+Module path inference will detect which dependencies are modules and configure the module path accordingly. The identified modules will then be removed from the usual classpath.
+NOTE: This is an experimental feature.
 `<sharedLibrary>`::
    If you want to build image as a shared library supply the following in the configuration of the plugin:
 [source,xml]
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
@@ -126,6 +126,9 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     protected final List<Path> imageClasspath;
     protected final List<Path> imageModulepath;
 
+    @Parameter(property = "inferModulePath", defaultValue = "false")
+    protected boolean inferModulePath;
+
     @Parameter(property = "debug", defaultValue = "false")
     protected boolean debug;
 
@@ -216,8 +219,11 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
         cliArgs.add("-cp");
         cliArgs.add(getClasspath());
 
-        cliArgs.add("--module-path");
-        cliArgs.add(getModulepath());
+        if(inferModulePath) {
+            logger.info("Module path inference: enabled");
+            cliArgs.add("--module-path");
+            cliArgs.add(getModulepath());
+        }
 
         if (debug) {
             cliArgs.add("-g");
@@ -347,7 +353,7 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
         try (FileSystem jarFS = openFileSystem(jarFileURI)) {
             // check if this is a module!
             Path moduleInfoClass = jarFS.getPath("/" + MODULE_INFO_CLASS);
-            module = Files.exists(moduleInfoClass);
+            module = inferModulePath && Files.exists(moduleInfoClass);
             Path nativeImageMetaInfBase = jarFS.getPath("/" + NATIVE_IMAGE_META_INF);
             if (Files.isDirectory(nativeImageMetaInfBase)) {
                 try (Stream<Path> stream = Files.walk(nativeImageMetaInfBase)) {
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
@@ -126,8 +126,9 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     protected final List<Path> imageClasspath;
     protected final List<Path> imageModulepath;
 
-    @Parameter(property = "inferModulePath", defaultValue = "false")
-    protected boolean inferModulePath;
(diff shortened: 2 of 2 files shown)


melix:
This looks like a rather dangerous change to me. There are numerous examples of projects which simply break when things are added on the module path. Also this changes the classpath order, which can be a problem. At the bare minimum I would add a property to make this opt-in.

loiclefevre:
> This looks like a rather dangerous change to me. There are numerous examples of projects which simply break when things are added on the module path. Also this changes the classpath order, which can be a problem. At the bare minimum I would add a property to make this opt-in.

Would that help if this is enabled if and only if the main artifact (not the dependencies) appears to be a module?

loiclefevre:
Ok, I've added a new plugin parameter: `inferModulePath`

```xml
                    <plugin>
                        <version>${native.maven.plugin.version}</version>
                        <executions>
                            <execution>
                                <id>build-native</id>
                                <goals>
                                    <goal>compile-no-fork</goal>
                                </goals>
                                <phase>package</phase>
                            </execution>
                            <execution>
                                <id>test-native</id>
                                <goals>
                                    <goal>test</goal>
                                </goals>
                                <phase>test</phase>
                            </execution>
                        </executions>
                        <configuration>
                            <verbose>true</verbose>
                            <skip>false</skip>
                            <imageName>${imageName}</imageName>
                            <fallback>false</fallback>
                            <agent>
                                <enabled>false</enabled>
                            </agent>
                            <inferModulePath>true</inferModulePath>
                            <buildArgs>
                                --module-path target/connectme-1.0.0.jar
                                -Ob
                                -march=native
                                --initialize-at-build-time=com.oracle.connect/com.oracle.connect.Main
                                --add-modules javafx.controls
                                --add-modules javafx.fxml
                                --add-modules com.oracle.connect
                                --module com.oracle.connect/com.oracle.connect.Main
                            </buildArgs>
                        </configuration>
                    </plugin>
```

fniephaus:
In Native Image, we have:

> [Added ability to promote jars from the class-path to the module-path in the native image driver](https://github.com/oracle/graal/commit/cdf0cb91799c461c4cf3ff4ce15a40585f0fe945#diff-355b70b1eca5ebd6e224292ba670c16371c32a5cf4e00c9b013fb4a7fc5f2c16R50). Use `ForceOnModulePath = ${module-name}`. Promoting a module to the module-path is equivalent to specifying it on the module-path in combination with exporting the module using `--add-modules ${module-name}` to the unnamed module.

Is that something that could be used instead here?

loiclefevre:
Sorry to say that's not solving the issue.
Thanks

fniephaus:
That's ok, thanks for checking whether that mechanism could be used to solve your problem.

We've discussed this PR in the NBT meeting and the biggest obstacle seems to be that this is not easily reproducible on the Gradle side. And if we only support module inference in Maven, we introduce asymmetry in our plugins which we like to avoid as much as possible.
The other concern is how useful module inference is in practice. Often, users seems to put modules on the class path because that's what they are used to. According to the Spring team, the demand for module support is still too low to be prioritized at this point.

Nonetheless, we'd like to continue this discussion. One alternative to the module inference approach would be a simple switch that puts the full class path on the module path. While that is potentially easier to implement, it's unclear whether a) this would solve your JavaFX and b) this would be useful for other use cases.

Let's keep this PR open until we have reached a consensus on how to proceed with module support in the NBT.

Again, thanks for your work so far! Much appreciated :)

JaroslavTulach:
> a simple switch that puts the full class path on the module path

That's basically what I did in https://github.com/jtulach/NativeImageModularDemo/commit/f61ebfd677ec77e9c19104028bc0690b4cc376a2#diff-abb4e0d98dcc6e2742e9b46d0f75f831b0e0dbe1feac6a85ebf38167130528a9:
- Use `-m` and `<classpath/>` to put [all JARs on classpath](https://github.com/jtulach/NativeImageModularDemo/commit/f61ebfd677ec77e9c19104028bc0690b4cc376a2#diff-abb4e0d98dcc6e2742e9b46d0f75f831b0e0dbe1feac6a85ebf38167130528a9R41)
- Use `-add-modules ALL-MODULE-PATH`

I am using `exec:exec` Maven plugin to achieve this as I am not sure how to configure NBT to achieve the same. Simple switch `<useModulePath>true</useModulePath>` for doing similar change (e.g. using `-p` instead of `-cp`) might be good enough solution for a lot of users.

fniephaus:
@loiclefevre would @JaroslavTulach's proposal work for your JavaFX application?

dnestoro:
@loiclefevre is there a plan to progress with this? Have you tried [suggestion](https://github.com/graalvm/native-build-tools/pull/607#issuecomment-2429319648) from @JaroslavTulach if it works?

Maran23:
I came here 'from JavaFX' as well and for me, the following is what I want:

> Simple switch true for doing similar change (e.g. using -p instead of -cp) might be good enough solution for a lot of users.

If your whole application is completely modularized, the classpath is not needed at all. So a switch to enable the module path instead of the classpath would be enough.

You usually want to use the module path completely as done in e.g. JLink.

As an example, this is what the official `maven-jlink-plugin` is doing automatically with the dependencies:
```
jlink [
--strip-debug, 
--compress, 
zip-9, 
--module-path, 
~\\.m2\\repository\\org\\openjfx\\javafx-base\\custom\\javafx-base-custom-win.jar;
~\\.m2\\repository\\org\\openjfx\\javafx-controls\\custom\\javafx-controls-custom-win.jar;
~\\.m2\\repository\\org\\openjfx\\javafx-graphics\\custom\\javafx-graphics-custom-win.jar;
~\\.m2\\repository\\org\\openjfx\\javafx-fxml\\custom\\javafx-fxml-custom-win.jar;
project\\target\\classes, 
--no-header-files, 
--no-man-pages, 
--add-modules, javafx.base,javafx.controls,javafx.graphics,javafx.fxml,project, 
--output, project\target\classifiers\runtime
]
```

PR #609, opened 2024-07-09
Fix args parsing on Windows

houcine7:
- When your home folder contains white spaces the write-file-args generates incorrect results as this example 
```
...
\\QC:\\Users\\Lahoucine
EL
ADDALI\\.m2\\repository\\io\\netty\\netty-transport\\4.1.108.Final\\netty-transport-4.1.108.Final.jar\\E
^/META-INF/native-image/
--exclude-config
\\QC:\\Users\\Lahoucine
EL
ADDALI\\.m2\\repository\\io\\netty\\netty-codec-http\\4.1.108.Final\\netty-codec-http-4.1.108.Final.jar\\E
^/META-INF/native-image/
-cp
"C:\\Users\\Lahoucine 
EL 
ADDALI\\Desktop\\outdir\\target/java-application-with-custom-packaging-0.1.jar"
```

- this pull request is aimed to address this issue.

Fixes https://github.com/micronaut-projects/micronaut-maven-plugin#1098


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -243,7 +243,29 @@ protected List<String> getBuildArgs() throws MojoExecutionException {
...
+                if(buildArg.startsWith("\\Q") ||
+                        buildArg.startsWith("-H:ConfigurationFileDirectories")) {
+                    cliArgs.add(buildArg);
+                    continue;
+                }
+                String[] args = buildArg.split("\\s+");
+                int i=0;
+                while(i < args.length) {
+                    String a =args[i];
+                    if (a.charAt(0) == System.getProperty("user.home").charAt(0)) {
```
alvarosanchez, line 255: Is this an attempt to detect `/` (or `\\`)? If so, please use `File.separatorChar`

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -243,7 +243,29 @@ protected List<String> getBuildArgs() throws MojoExecutionException {
...
+                        StringBuilder path = new StringBuilder(a);
+                        i++;
+                        while( i< args.length && args[i].toLowerCase().charAt(0) <= 'z' &&
```
alvarosanchez, line 258: Is the intent of this to consider only args with letters? Why? Also, what about numbers.

A better option is `Character.isLetterOrDigit()`

alvarosanchez (changes requested):
In general, the code you added to `getBuildArgs()` needs to be unit tested.

Please refactor it to a separate method that you can varify in a Spock unit test.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
houcine7: This is to detect if the argument is a path (e.g., /home/... or C:/...). If it is, then it might be split with the space regex buildArg.split("\\s+"). For example, if you have the argument `/home/My user/target/app-name.jar`, it will be split into `/home/My` and `user/target/app-name.jar` and added as separate CLI arguments, which will result in parsing errors

example in windows: 
```
buildArgs = ["-cp C:\\Users\\My User\\my-app\\target\\app-name.jar"]
// after splitting with spaces
with get cliArgs = ["-cp", "C:\\Users\\My", "User\\my-app\\target\\app-name.jar"]
// resulting in parsing issues 
```
so I think this is an issue in linux and also windows.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
houcine7: This is to construct the path that is being split, so I assumed that paths may contain only letters

alvarosanchez: In fact you should check for anything but `File.separatorChar`. Paths may contain numbers, dashes, underscores, dots... almost everything but `File.separatorChar`

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
```diff
@@ -49,12 +49,15 @@ import org.eclipse.jetty.server.handler.ResourceHandler
...
 
+import java.nio.file.FileVisitResult
 import java.nio.file.Files
 import java.nio.file.Path
+import java.nio.file.SimpleFileVisitor
 import java.nio.file.StandardCopyOption
+import java.nio.file.attribute.BasicFileAttributes
 
 abstract class AbstractGraalVMMavenFunctionalTest extends Specification {
-    @TempDir
```
dnestoro, line 57: Why you have removed @TempDir annotation and then proceeded with deleting files manually in the cleanup function?

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
```diff
@@ -158,7 +181,7 @@ abstract class AbstractGraalVMMavenFunctionalTest extends Specification {
         var resultingSystemProperties = [
                 "common.repo.uri": System.getProperty("common.repo.uri"),
                 "seed.repo.uri": System.getProperty("seed.repo.uri"),
-                "maven.repo.local": testDirectory.resolve("local-repo").toFile().absolutePath
+                "maven.repo.local": testDirectory.resolve("local repo").toFile().absolutePath
```
dnestoro, line 184: Why is this necessary?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -243,7 +243,11 @@ protected List<String> getBuildArgs() throws MojoExecutionException {
 
         if (buildArgs != null && !buildArgs.isEmpty()) {
             for (String buildArg : buildArgs) {
-                cliArgs.addAll(Arrays.asList(buildArg.split("\\s+")));
+                if(buildArg.startsWith("\\Q") || buildArg.startsWith("-H")) {
```
dnestoro, line 246: Are we sure that this is the right condition? I mean, is it sufficient that we only check start of the buildArg? Did you check all possible args starting with `\\Q` and `-H`, whether all of them can be added without `split` or not?

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
dnestoro: Should you add new functional test, and not just change the existing one?

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
houcine7: In my local when I work with Tempdir No alla tests pass, because the on windows the tempDir avoids white spaces in path with a certain encoding (C:\Users\Lahoucine~\AppData\Temp) so no parsing errors will show up if I have withe spaces in path.

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
houcine7: I added this for reproducing the issue.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
houcine7: the arguments that starts with \\Q and end with \\E should not be splitted by spaces.

And for also -H I see only the argument -H:ConfigurationDirectories and they should be on one line not splitted by spaces (which could be in path )

native-maven-plugin/src/testFixtures/groovy/org/graalvm/buildtools/maven/AbstractGraalVMMavenFunctionalTest.groovy
dnestoro: Okay, I asked because you changed the existing test, and now we are checking only the case with the space, right? My concern was that you removed existing check (when dir name contains `-`), and that you should maybe add a separate test for this case

alvarosanchez (approved):

PR #614, opened 2024-08-12
Adjust native-build tools to reachability-metadata.json

vjovanov:


native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithAgentFunctionalTest.groovy
```diff
@@ -42,10 +42,35 @@
...
+        if (metadataInSingleConfigFile()) {
+            return file("${path}/reachability-metadata.json").exists()
+        }
+
+        boolean allFilesExist = true
+        ['jni', 'proxy', 'reflect', 'resource', 'serialization'].each { name ->
+            if (!file("${path}/${name}-config.json").exists()) {
+                allFilesExist = false
+            }
+        }
```
melix, line 69:
```suggestion
        boolean allFilesExist = ['jni', 'proxy', 'reflect', 'resource', 'serialization'].every { name ->
           file("${path}/${name}-config.json").exists()
        }
```

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/DefaultGraalVmExtension.java
```diff
@@ -79,7 +78,8 @@ public DefaultGraalVmExtension(NamedDomainObjectContainer<NativeImageOptions> na
         agentOpts.getEnabled().convention(false);
         agentOpts.getModes().getConditional().getParallel().convention(true);
         agentOpts.getMetadataCopy().getMergeWithExisting().convention(false);
-        agentOpts.getFilterableEntries().convention(Arrays.asList("org.gradle.", "org.junit."));
+        // TODO: replace with agent access filter
+        // agentOpts.getFilterableEntries().convention(Arrays.asList("org.gradle.", "org.junit."));
```
melix, line 82: Is this supposed to stay commented out?

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/actions/MergeAgentFilesAction.java
```diff
@@ -94,8 +95,11 @@ public MergeAgentFilesAction(Provider<Boolean> isMergingEnabled,
         this.noLauncherProperty = objectFactory.property(JavaLauncher.class);
     }
 
+    private static final Set<String> metadataFiles = Set.of("reflect-config.json", "jni-config.json", "proxy-config.json", "resource-config.json", "reachability-metadata.json");
```
melix, line 98:
```suggestion
    private static final Set<String> METADATA_FILES = Set.of("reflect-config.json", "jni-config.json", "proxy-config.json", "resource-config.json", "reachability-metadata.json");
```

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/actions/MergeAgentFilesAction.java
```diff
@@ -94,8 +95,11 @@ public MergeAgentFilesAction(Provider<Boolean> isMergingEnabled,
...
+
     private static boolean isConfigDir(String dir) {
-        return Arrays.stream(new File(dir).listFiles()).anyMatch(file -> file.getName().equals("reflect-config.json"));
+        return Arrays.stream(new File(dir).listFiles())
+           .anyMatch(file -> metadataFiles.contains(file.getName()));
```
melix, line 102:
```suggestion
           .anyMatch(file -> METADATA_FILES.contains(file.getName()));
```

melix (changes requested):
Would also be good to add release notes and context about this change

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/DefaultGraalVmExtension.java
dnestoro: We were considering how to properly filter `org.gradle` and `org.junit` through `access-filter` file. Do you have any suggestion what is the best way to apply filter file for this purpose?

melix: I'm not sure what you mean by that. We already have `getAccessFilterFiles` where you can add more access filter files. Is the problem that you don't want to use `getFilterableEntries`, or that it hardcodes junit? If it's about hardcoding, I'm not sure how we can work around.

In any case, it's suspicious that this code is commented out, and that no tests are failing.

dnestoro: The idea is to use `access-filter.json` for filtering entries in the config files. Therefore we can replace custom [manual post processing](https://github.com/graalvm/native-build-tools/pull/614/files#diff-8d46943d72fca2ee4ffb4ecafae60ef189596e0a95f4b2879ac495a4abf7f758) and use existing Native Image feature that does the same.

melix (approved):
Note to self: the section which creates a temporary file for the access filter file is good enough for now, but it should probably be implemented differently.

Basically, instead of having the agent configuration create the file and mutate itself, there should be a task which does this, and that task should be an implicit dependency of the agent configuration. It requires changes to both the Gradle and Maven plugins.

However, currently the agent options are "serialized" as strings, including paths, which prevents Gradle from understanding that the `Collection<String>` actually refer to file contents. This means that the up-to-date checking is incorrect, since it will only consider the file paths and not their contents. That is already the situation today, so nothing to change in this PR, it's just notes so that I remember later.

PR #616, opened 2024-08-30
Fix error message when the JDK is not a GraalVM

eregon:
Currently it still mentions `gu` but that's gone and a misleading error.

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
@@ -91,29 +91,29 @@ public class NativeImageExecutableLocator {
             executablePath = metadata.getInstallationPath().file("bin/" + NATIVE_IMAGE_EXE).getAsFile();
         }
 
-        try {
-            if (!executablePath.exists()) {
-                logger.log("Native Image executable wasn't found. We will now try to download it. ");
-                File graalVmHomeGuess = executablePath.getParentFile();
+        File graalVmHomeGuess = executablePath.getParentFile();
+        File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
+        if (guPath.exists() && !executablePath.exists()) {
+            logger.log("Native Image executable wasn't found. We will now try to download it. ");
 
-                File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
-                if (!guPath.exists()) {
-                    throw new GradleException("'" + GU_EXE + "' at '" + guPath + "' tool wasn't found. This probably means that JDK at isn't a GraalVM distribution.");
-                }
-                ExecResult res = execOperations.exec(spec -> {
-                    spec.args("install", "native-image");
-                    spec.setExecutable(Paths.get(graalVmHomeGuess.getAbsolutePath(), GU_EXE));
-                });
-                if (res.getExitValue() != 0) {
-                    throw new GradleException("Native Image executable wasn't found, and '" + GU_EXE + "' tool failed to install it.");
-                }
-                diagnostics.withGuInstall();
+            ExecResult res = execOperations.exec(spec -> {
+                spec.args("install", "native-image");
+                spec.setExecutable(Paths.get(graalVmHomeGuess.getAbsolutePath(), GU_EXE));
+            });
+            if (res.getExitValue() != 0) {
+                throw new GradleException("Native Image executable wasn't found, and '" + GU_EXE + "' tool failed to install it.\n" +
+                        "Make sure to declare the GRAALVM_HOME or JAVA_HOME environment variable or install GraalVM with " +
+                        "native-image in a standard location recognized by Gradle Java toolchain support");
             }
-        } catch (GradleException e) {
-            throw new GradleException("Determining GraalVM installation failed with message: " + e.getMessage() + "\n\n"
-                                      + "Make sure to declare the GRAALVM_HOME environment variable or install GraalVM with " +
-                                      "native-image in a standard location recognized by Gradle Java toolchain support");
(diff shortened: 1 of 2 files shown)


eregon:
I tried to fix the Gradle plugin in a similar manner.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
```diff
@@ -91,29 +91,32 @@ public static File findNativeImageExecutable(Property<JavaLauncher> javaLauncher
...
-                File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
-                if (!guPath.exists()) {
-                    throw new GradleException("'" + GU_EXE + "' at '" + guPath + "' tool wasn't found. This probably means that JDK at isn't a GraalVM distribution.");
-                }
+        File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
+        if (guPath.exists() && !executablePath.exists()) {
+            logger.log("Native Image executable wasn't found. We will now try to download it. ");
+            File graalVmHomeGuess = executablePath.getParentFile();
+
+            File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
```
fniephaus, line 99: Unless something is wrong with the diff, it seems you are re-declaring `guPath` here,and then also check for it's existence again.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
```diff
@@ -91,29 +91,32 @@ public static File findNativeImageExecutable(Property<JavaLauncher> javaLauncher
...
             }
-        } catch (GradleException e) {
-            throw new GradleException("Determining GraalVM installation failed with message: " + e.getMessage() + "\n\n"
-                                      + "Make sure to declare the GRAALVM_HOME environment variable or install GraalVM with " +
-                                      "native-image in a standard location recognized by Gradle Java toolchain support");
         }
+
+        if (!executablePath.exists()) {
+            throw new GradleException(executablePath + " wasn't found. This probably means that JDK isn't a GraalVM distribution.\n" +
+                    "Make sure to declare the GRAALVM_HOME environment variable or install GraalVM with " +
```
fniephaus, line 116:
```suggestion
                    "Make sure to declare the GRAALVM_HOME or JAVA_HOME environment variables or install GraalVM with " +
```

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
```diff
@@ -91,29 +91,32 @@ public static File findNativeImageExecutable(Property<JavaLauncher> javaLauncher
...
+            File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
+            if (guPath.exists()) {
                 ExecResult res = execOperations.exec(spec -> {
                     spec.args("install", "native-image");
                     spec.setExecutable(Paths.get(graalVmHomeGuess.getAbsolutePath(), GU_EXE));
                 });
                 if (res.getExitValue() != 0) {
-                    throw new GradleException("Native Image executable wasn't found, and '" + GU_EXE + "' tool failed to install it.");
+                    throw new GradleException("Native Image executable wasn't found, and '" + GU_EXE + "' tool failed to install it.\n" +
+                            "Make sure to declare the GRAALVM_HOME environment variable or install GraalVM with " +
```
fniephaus, line 107:
```suggestion
                            "Make sure to declare the GRAALVM_HOME or JAVA_HOME environment variables or install GraalVM with " +
```

fniephaus (changes requested):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
eregon: I added the `or JAVA_HOME`, I don't know if it's respected by the Gradle plugin though.
I kept `variable` singular since the user should set either of them but no need to set both.

fniephaus (approved):

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageExecutableLocator.java
```diff
@@ -91,29 +91,29 @@ public static File findNativeImageExecutable(Property<JavaLauncher> javaLauncher
...
-                ExecResult res = execOperations.exec(spec -> {
-                    spec.args("install", "native-image");
-                    spec.setExecutable(Paths.get(graalVmHomeGuess.getAbsolutePath(), GU_EXE));
-                });
-                if (res.getExitValue() != 0) {
-                    throw new GradleException("Native Image executable wasn't found, and '" + GU_EXE + "' tool failed to install it.");
-                }
-                diagnostics.withGuInstall();
+        File graalVmHomeGuess = executablePath.getParentFile();
+        File guPath = graalVmHomeGuess.toPath().resolve(GU_EXE).toFile();
```
dnestoro, line 95: What if `executablePath` doesn't exist at this point? Then `getParentFile` returns null, right? Therefor, `graalVmHomeGuess` will be null and this can cause an exception?

eregon: `File#getParentFile` does not care whether the file exists.
It only returns null if `executablePath` is `/`, which is not possible given how `executablePath` is constructed (it ends with `.../bin/native-image*` so the getParentFile is `.../bin`).

dnestoro (approved):

Issue #357, opened 2022-11-02
Add Gradle Toolchain Support


jamesward:
I'm using:
```
id("org.graalvm.buildtools.native") version "0.9.16"
```

And have a Gradle toolchain configured:
```
java {
	toolchain {
		languageVersion.set(JavaLanguageVersion.of(17))
		vendor.set(JvmVendorSpec.GRAAL_VM)
	}
}
```

And I get:
```
./gradlew nativeRun

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':nativeCompile'.
> Failed to calculate the value of task ':compileJava' property 'javaCompiler'.
   > No compatible toolchains found for request filter: {languageVersion=17, vendor=GRAAL_VM, implementation=vendor-specific} (auto-detect true, auto-download true)
```

It'd be awesome if the Gradle Native plugin worked together with Gradle Toolchains.

graemerocher:
We would like to, but we need the Gradle team's help. See https://github.com/gradle/gradle/pull/18028

sschuberth:
> We would like to, but we need the Gradle team's help. See [gradle/gradle#18028](https://github.com/gradle/gradle/pull/18028)

Note that according to @melix's [comment](https://github.com/gradle/gradle/pull/18028#issuecomment-1266657213) "The PR, in itself, is irrelevant though" now that Gradle 7.6 added support for Java Toolchain downloads from arbitrary repositories, also see this [comment](https://github.com/graalvm/native-build-tools/issues/100#issuecomment-1298381682) of mine.

sschuberth:
In conjunction with https://github.com/gradle/foojay-toolchains this now basically works as outlined e.g. at https://github.com/oss-review-toolkit/ort/compare/main...graalvm-upgrade.

jamesward:
Thanks @sschuberth but I couldn't get that working. Test project:
https://github.com/jamesward/hello-graalvm

For me:
```
$ ./gradlew nativeRun

Invalid Java installation found at '/home/jw/.gradle/jdks/graalvm-ce-java8-linux-amd64-21/graalvm-ce-java8-21.3.1' (Auto-provisioned by Gradle). It will be re-checked in the next build. This might have performance impact if it keeps failing. Run the 'javaToolchains' task for more details.
[...]

PR #623, opened 2024-09-13
Improve Native Image SBOM Generation

rudsberg:
### Overview 
This PR extends the `native-maven-plugin` to improve the accuracy of the Software Bill of Material (SBOM) that is generated as part of Native Image builds (only available in Oracle GraalVM).

The high-level approach is the following:
* The `native-maven-plugin` invokes `cyclonedx-maven-plugin` to create a base SBOM. 
* The components of the base SBOM are augmented with new fields, foremost `packageNames` which lists the package names of that components.
* The SBOM is stored at a known location and is handed over to Native Image.
* **Only in Native Image**: Components and dependencies are pruned whenever possible as determined by the Native Image static analysis. Native Image relies on the augmented fields to map an identified type to its component.
* **Only in Native Image**: The final SBOM is saved according to the arguments passed to `--enable-sbom`.   

By using a COTS SBOM generator to get a base SBOM and passing extra information to Native Image which prunes it to make it more accurate, we at worst get an SBOM that conforms to industry standards and at best an SBOM that is significantly more accurate.

### Future Work
Future work includes:
* Adding similar support for the `native-gradle-plugin`.
* Attempting to resolve the packageNames for shaded dependencies similar as is currently done for the main artifact. Currently, shaded dependencies are marked as `prunable=false` which instructs Native Image to not prune them or any of their transitive dependencies.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -19,6 +19,8 @@ groovy = "3.0.11"
 jetty = "11.0.11"
 plexusUtils = "4.0.0"
 plexusXml = "4.0.2"
+cyclonedxMaven = "2.8.1"
+pluginExecutorMaven = "2.4.0"
 
 [libraries]
 # Local projects
@@ -61,3 +63,6 @@ jetty-server = { module = "org.eclipse.jetty:jetty-server", version.ref = "jetty
 plexus-utils = { module = "org.codehaus.plexus:plexus-utils", version.ref = "plexusUtils" }
 
 plexus-xml = { module = "org.codehaus.plexus:plexus-xml", version.ref = "plexusXml" }
+
+cyclonedx-maven-plugin = { module = "org.cyclonedx:cyclonedx-maven-plugin", version.ref="cyclonedxMaven" }
+plugin-executor-maven = { module = "org.twdata.maven:mojo-executor", version.ref="pluginExecutorMaven" }
--- a/native-maven-plugin/build.gradle.kts
+++ b/native-maven-plugin/build.gradle.kts
@@ -65,6 +65,8 @@ dependencies {
     implementation(libs.jvmReachabilityMetadata)
     implementation(libs.plexus.utils)
     implementation(libs.plexus.xml)
+    implementation(libs.cyclonedx.maven.plugin)
+    implementation(libs.plugin.executor.maven)
 
     compileOnly(libs.maven.pluginApi)
     compileOnly(libs.maven.core)
@@ -178,3 +180,4 @@ tasks.withType<Checkstyle>().configureEach {
     // generated code
     exclude("**/RuntimeMetadata*")
 }
+
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
@@ -49,36 +49,24 @@ import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.toolchain.ToolchainManager;
+import org.codehaus.plexus.logging.Logger;
 import org.graalvm.buildtools.maven.config.ExcludeConfigConfiguration;
 import org.graalvm.buildtools.utils.NativeImageConfigurationUtils;
 import org.graalvm.buildtools.utils.NativeImageUtils;
 import org.graalvm.buildtools.utils.SharedConstants;
 
 import javax.inject.Inject;
-import java.io.File;
-import java.io.InputStream;
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
-import java.io.IOException;
+import java.io.*;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystem;
-import java.nio.file.FileSystems;
-import java.nio.file.FileSystemAlreadyExistsException;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.Files;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Collections;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.nio.file.*;
+import java.util.*;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+
+import static org.graalvm.buildtools.utils.NativeImageUtils.ORACLE_GRAALVM_IDENTIFIER;
 
 /**
  * @author Sebastien Deleuze
@@ -87,6 +75,7 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     protected static final String NATIVE_IMAGE_META_INF = "META-INF/native-image";
     protected static final String NATIVE_IMAGE_PROPERTIES_FILENAME = "native-image.properties";
     protected static final String NATIVE_IMAGE_DRY_RUN = "nativeDryRun";
+    private static String nativeImageVersionInformation = null;
 
     @Parameter(defaultValue = "${plugin}", readonly = true) // Maven 3 only
     protected PluginDescriptor plugin;
@@ -447,6 +436,24 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
         if (requiredVersion == null) {
             return;
         }
+        NativeImageUtils.checkVersion(requiredVersion, getVersionInformation(logger));
+    }
+
+    static protected boolean isOracleGraalVM(Logger logger) throws MojoExecutionException {
+        return getVersionInformation(logger).contains(ORACLE_GRAALVM_IDENTIFIER);
+    }
+
+    /**
+     * Returns the output of calling "native-image --version".
+     * @param logger a logger, that may be null, to print warnings or useful information.
+     * @return the output as a string joined by "\n".
+     * @throws MojoExecutionException when any errors occurred.
+     */
+    static protected String getVersionInformation(Logger logger) throws MojoExecutionException {
+        if (nativeImageVersionInformation != null) {
+            return nativeImageVersionInformation;
+        }
+
         Path nativeImageExecutable = NativeImageConfigurationUtils.getNativeImage(logger);
         try {
             ProcessBuilder processBuilder = new ProcessBuilder(nativeImageExecutable.toString());
@@ -457,12 +464,11 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
                 throw new MojoExecutionException("Execution of " + commandString + " returned non-zero result");
             }
             InputStream inputStream = versionCheckProcess.getInputStream();
-            String versionToCheck = new BufferedReader(
+            nativeImageVersionInformation = new BufferedReader(
                     new InputStreamReader(inputStream, StandardCharsets.UTF_8))
                     .lines()
                     .collect(Collectors.joining("\n"));
-            NativeImageUtils.checkVersion(requiredVersion, versionToCheck);
-
+            return nativeImageVersionInformation;
         } catch (IOException | InterruptedException e) {
             throw new MojoExecutionException("Checking GraalVM version with " + nativeImageExecutable + " failed", e);
         }
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java
@@ -44,6 +44,7 @@ package org.graalvm.buildtools.maven;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.BuildPluginManager;
 import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Parameter;
@@ -121,6 +122,12 @@ public abstract class AbstractNativeMojo extends AbstractMojo {
 
     @Component
     protected MavenSession mavenSession;
+
+    @Component
+    protected MavenProject mavenProject;
+
+    @Component
+    protected BuildPluginManager pluginManager;
 
     @Component
     protected RepositorySystem repositorySystem;
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
@@ -53,11 +53,14 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.codehaus.plexus.component.configurator.expression.ExpressionEvaluationException;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
+import org.graalvm.buildtools.maven.sbom.SBOMGenerator;
+import org.graalvm.buildtools.utils.NativeImageUtils;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.function.BiFunction;
 
+import static org.graalvm.buildtools.utils.NativeImageUtils.ORACLE_GRAALVM_IDENTIFIER;
 
 /**
  * This goal runs native builds. It functions the same as the native:compile goal, but it
@@ -73,6 +76,14 @@ public class NativeCompileNoForkMojo extends AbstractNativeImageMojo {
 
     @Parameter(property = "skipNativeBuildForPom", defaultValue = "false")
     private boolean skipNativeBuildForPom;
+
+    /**
+     * Used in {@link NativeCompileNoForkMojo#generateAugmentedSBOMIfNeeded} to determine if an augmented SBOM should
+     * be produced by {@link SBOMGenerator}.
+     */
+    @Parameter
+    private Boolean augmentedSBOM;
+    public static final String AUGMENTED_SBOM_PARAM_NAME = "augmentedSBOM";
 
     private PluginParameterExpressionEvaluator evaluator;
 
@@ -101,7 +112,72 @@ public class NativeCompileNoForkMojo extends AbstractNativeImageMojo {
         maybeSetMainClassFromPlugin(this::consumeConfigurationNodeValue, "org.apache.maven.plugins:maven-assembly-plugin", "archive", "manifest", "mainClass");
         maybeSetMainClassFromPlugin(this::consumeConfigurationNodeValue, "org.apache.maven.plugins:maven-jar-plugin", "archive", "manifest", "mainClass");
         maybeAddGeneratedResourcesConfig(buildArgs);
+
+        generateAugmentedSBOMIfNeeded();
+
         buildImage();
+    }
+
+    /**
+     * Generates an augmented SBOM using the {@link SBOMGenerator} based on specific conditions:
+     *
+     * 1. If {@link NativeCompileNoForkMojo#augmentedSBOM} is explicitly set to false: No SBOM is generated.
+     * 2. If {@link NativeCompileNoForkMojo#augmentedSBOM} is explicitly set to true: An augmented SBOM is
+     *    generated if the required conditions are met.
+     * 3. If {@link NativeCompileNoForkMojo#augmentedSBOM} is not set: An augmented SBOM is generated only if
+     *    SBOM generation is configured for Native Image via a build argument.
+     *
+     * Note: Augmented SBOMs are only supported in Oracle GraalVM for JDK {@link SBOMGenerator#requiredNativeImageVersion}
+     * or later.
+     *
+     * @throws IllegalArgumentException if {@link NativeCompileNoForkMojo#augmentedSBOM} is explicitly set to true
+     *         but required conditions are not met (e.g., using community edition or JDK version 23 or earlier).
+     * @throws MojoExecutionException if augmented SBOM generation was attempted but failed.
+     */
+    private void generateAugmentedSBOMIfNeeded() throws IllegalArgumentException, MojoExecutionException {
+        boolean optionWasSet = augmentedSBOM != null;
+        augmentedSBOM = optionWasSet ? augmentedSBOM : true;
+
+        int detectedJDKVersion = NativeImageUtils.getMajorJDKVersion(getVersionInformation(logger));
+        String sbomNativeImageFlag = "--enable-sbom";
+        boolean sbomEnabledForNativeImage = getBuildArgs().stream().anyMatch(v -> v.contains(sbomNativeImageFlag));
+        if (optionWasSet) {
+            if (!augmentedSBOM) {
+                /* User explicitly opted out. */
+                return;
+            }
+
+            if (!isOracleGraalVM(logger)) {
+                throw new IllegalArgumentException(
+                        String.format("Configuration option %s is only supported in %s.", AUGMENTED_SBOM_PARAM_NAME, ORACLE_GRAALVM_IDENTIFIER));
+            }
+
+            SBOMGenerator.checkAugmentedSBOMSupportedByJDKVersion(detectedJDKVersion, true);
+
+            if (!sbomEnabledForNativeImage) {
+                buildArgs.add(sbomNativeImageFlag);
+                logger.info(String.format("Automatically added build argument %s to Native Image because configuration option %s was set to true. " +
+                        "An SBOM will be embedded in the image.", sbomNativeImageFlag, AUGMENTED_SBOM_PARAM_NAME));
+            }
+
+            /* Continue to generate augmented SBOM because parameter option explicitly set and all conditions are met. */
+        } else {
+            if (!isOracleGraalVM(logger) || !sbomEnabledForNativeImage) {
+                return;
+            }
+
+            if (!SBOMGenerator.checkAugmentedSBOMSupportedByJDKVersion(detectedJDKVersion, false)) {
+               return;
+            }
+
+            /*
+             * Continue to generate augmented SBOM because although the parameter option was not set, SBOM is used for
+             * Native Image and all conditions are met.
+             */
+        }
+
+        var sbomGenerator = new SBOMGenerator(mavenProject, mavenSession, pluginManager, repositorySystem, mainClass, logger);
+        sbomGenerator.generate();
     }
 
     private String consumeConfigurationNodeValue(String pluginKey, String... nodeNames) {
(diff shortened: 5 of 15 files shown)


native-maven-plugin/build.gradle.kts
```diff
@@ -45,7 +45,8 @@ import org.gradle.util.GFileUtils
 plugins {
     `java-library`
     groovy
-    checkstyle
+    // TODO: forced to remove this to allow building on my machine. Remove this before merging.
+//    checkstyle
```
rudsberg, line 49: There are a few instances where I had to uncomment code to make it build on my machine. If anyone has resolutions for avoiding this, please let me know. In either case, they will of course be removed before merging.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
```diff
@@ -101,6 +105,12 @@ public void execute() throws MojoExecutionException {
         maybeSetMainClassFromPlugin(this::consumeConfigurationNodeValue, "org.apache.maven.plugins:maven-assembly-plugin", "archive", "manifest", "mainClass");
         maybeSetMainClassFromPlugin(this::consumeConfigurationNodeValue, "org.apache.maven.plugins:maven-jar-plugin", "archive", "manifest", "mainClass");
         maybeAddGeneratedResourcesConfig(buildArgs);
+
+        if (enableSBOM) {
```
rudsberg, line 109: This should also guard that the EE version of Native Image is used. What's an easy and clean way to differentiate between EE and CE from this context?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -0,0 +1,220 @@
...
+ */
+final public class SBOMGenerator {
+    private final MavenProject mavenProject;
+    private final MavenSession mavenSession;
+    private final BuildPluginManager pluginManager;
+    private final RepositorySystem repositorySystem;
+    private final String mainClass;
+    private final Logger logger;
+
+    private static final String SBOM_NAME = "WIP_SBOM";
```
matneu, line 97: I think this deserves a better name than "WIP". It's actually a proper SBOM based on build tooling information, which may be overapproximated, but useful on its own. How about "InitialSBOM" (to indicate that it'll be refined later on) or "BaselineSBOM"?

rudsberg: I totally agree. I will update it to be called `base_sbom.json`.

fniephaus:
@melix do you think this can be reasonably well ported to Gradle, or do you see any issues?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
rudsberg: Implemented this by calling `native-image --version` and checking if the output contains "Oracle GraalVM".

melix:
So I think the biggest issue here is that we are bounding ourselves to an external plugin (cyclonedx). This could be a problem is a user is already using it, or if they are using another SBOM generator. While I'm not sure if we can work around for Maven, for Gradle this would certainly be an issue. An option would be to _react_ on the application of a known plugin like cyclonedx (e.g, if you apply it, then we do more work), but not apply it transparently.

rudsberg:
Hey @melix, thanks for your comment. I am happy to discuss this further. I will do my best to answer this and give the different pros/cons of the possible approaches. 

> [...] or if they are using another SBOM generator

I don't see the problem with users using other SBOM generators in conjunction with our augmented Native Image based SBOM generator. Would you mind to elaborate how this would pose an issue?

To give some more context, this augmented SBOM feature is only used if the user:
* Explicitly use the `--enable-sbom` build argument for Native Image, i.e. opting in to using the Native Image derived SBOM. In these cases, we automatically use this augmented approach.  
* Explicitly set the configuration parameter `augmentedSBOM` of the plugin to `true` (then we automatically add `--enable-sbom` to the build argument if not already used).

The user can always opt-out from this feature by setting `augmentedSBOM` to false. 

---- 

For maven it shouldn't pose any issues even if the user application uses a different version of CycloneDX since maven allows plugins to have a different version from other plugins or the application itself. The user could use an older version of CycloneDX while we transparently use 2.8.1, for example. 

Gradle, however, applies version resolution globally, so with the current approach, that would automatically force the latest version of CycloneDX to be used, thus potentially changing the version of the user's CycloneDX version. This could be an issue if CycloneDX is _not_ backwards compatible. If it is not, and a user application depends on a specific behavior of an earlier version of CycloneDX that is not present in a later version, then that could break the user application. 

> An option would be to react on the application of a known plugin like cyclonedx (e.g, if you apply it, then we do more work), but not apply it transparently.

That's an alternative approach. I see the following benefits with that approach compared to the current:
* **No versioning issues in Gradle.** If a user uses Gradle and already uses CycloneDX but with an earlier version, we avoid forcing them to use our newer version of CycloneDX. 
* **More flexibility.** Users can choose what baseline CycloneDX SBOM generator they want to use. For example, using `quarkus-cyclonedx-generator` instead of `cyclonedx-maven-plugin`. 
* **More transparency.** We don't transparently introduce a dependency on `cyclonedx-maven-plugin` and instead let the users decide if they want to have such a dependency.

However, there are downsides:
* **Harder to adopt this feature.** This augmented mode would not work out-of-the-box for people already generating SBOMs with Native Image, instead users must explicitly opt-in by adding a dependency on a _supported_ SBOM generator. We would, for example, give a hint to users like "add the cyclonedx-maven-plugin to your pom.xml to enable augmented SBOMs" (possibly also requiring them to use a specific file name so we can locate it). 
* **Possibly harder to maintain.** The increased flexibility of allowing users to choose their CycloneDX generator and version might be harder to maintain if there are differences between generators or if they use different CycloneDX versions. The Native Image feature assumes the CycloneDX 1.5 version and that certain fields are used. Of course, one could specify that only a specific version(s) are supported. 

Here's my take on this. I am worried that adoption would suffer if we require users to themselves add a dependency on some CycloneDX generator. Instead, I see great benefits of transparently improving the SBOMs for all existing users that use the `--enable-sbom` build argument and making it very easy for users that start adopting SBOMs. To me it seems the only reason _not to_ use my suggested approach would be the automatic upgrade of CycloneDX format for Gradle users using an older version of the CycloneDX plugin. However, I don't think this is a problem. This is taken from the CycloneDX website:

> All CycloneDX releases are production quality standards that do not change. When drafting revisions, backwards compatibility with prior versions is of upmost priority.

I believe it's fair to assume the same guiding principle holds for their CycloneDX Gradle plugin. Even in the case for users on the new Native Image build plugin version with this augmented SBOM feature that use an older version of the Gradle CycloneDX plugin, and that gets automatically updated which possibly breaks their application, resolving such incompatibility between CycloneDX versions should be straightforward. Thus, I don't think this should hinder us from moving forward with this approach.

To increase transparency, it should be clear from the documentation that we are using the `cyclonedx-{maven,gradle}-plugin` under the hood for this feature. Lastly, if for some reason that this approach introduces issues, we can promptly revisit this and go for the other approach.  

Disclaimer: I am far from being a senior plugin developer (this is the first plugin-development work) and I recently the GraalVM team. I look forward hearing your thoughts on this! 

 CC: @matneu @fniephaus.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/FileWalkerUtility.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Consumer;
+
+final class FileWalkerUtility {
```
matneu, line 52: Please add javadoc

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -0,0 +1,283 @@
...
+ * <p>
+ * Creating the package-name-to-component mapping in the context of Native Image, without the knowledge known at the
+ * plugin build-time is difficult, which was the primary motivation for realizing this approach.
+ * <p>
+ * Benefits:
+ * * Great Baseline: Produces an industry-standard SBOM at minimum.
+ * * Enhanced Accuracy: Native Image augments and refines the SBOM, potentially significantly improving its accuracy.
+ */
+final public class SBOMGenerator {
+    public static final int requiredNativeImageVersion = 24;
```
matneu, line 87: No minor version? Wouldn't this be 24.2?

matneu: ah, never mind, GraalVM for JDK 24 = GraalVM 24.2

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/ShadedPackageNameResolver.java
```diff
@@ -0,0 +1,501 @@
...
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+class ShadedPackageNameResolver {
```
matneu, line 58: Please add class-level javadoc

matneu (approved):

gradle/libs.versions.toml
```diff
@@ -61,3 +63,6 @@ jetty-server = { module = "org.eclipse.jetty:jetty-server", version.ref = "jetty
 plexus-utils = { module = "org.codehaus.plexus:plexus-utils", version.ref = "plexusUtils" }
 
 plexus-xml = { module = "org.codehaus.plexus:plexus-xml", version.ref = "plexusXml" }
+
+cyclonedx-maven-plugin = { module = "org.cyclonedx:cyclonedx-maven-plugin", version.ref="cyclonedxMaven" }
+plugin-executor-maven = { module = "org.twdata.maven:mojo-executor", version.ref="pluginExecutorMaven" }
```
melix, line 68: Can we avoid introducing too many external dependencies, especially plugins which are not under our control?

native-maven-plugin/build.gradle.kts
```diff
@@ -91,7 +94,8 @@ dependencies {
     functionalTestCommonRepository(libs.utils)
     functionalTestCommonRepository(libs.junitPlatformNative)
     functionalTestCommonRepository(libs.jvmReachabilityMetadata)
-    functionalTestCommonRepository("org.graalvm.internal:library-with-reflection")
+    // TODO: forced to remove this to allow building on my machine. Remove this before merging.
+//    functionalTestCommonRepository("org.graalvm.internal:library-with-reflection")
```
melix, line 98: :x: This needs to be removed. What is the problem with local build?

native-maven-plugin/build.gradle.kts
```diff
@@ -173,8 +177,10 @@ tasks {
     }
 }
 
-tasks.withType<Checkstyle>().configureEach {
-    configFile = layout.projectDirectory.dir("../config/checkstyle.xml").asFile
-    // generated code
-    exclude("**/RuntimeMetadata*")
-}
+// TODO: forced to remove this to allow building on my machine. Remove this before merging.
```
melix, line 180: :x: This needs to be fixed. Can you tell what's the problem with local builds?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -437,6 +440,21 @@ protected void checkRequiredVersionIfNeeded() throws MojoExecutionException {
         if (requiredVersion == null) {
             return;
         }
+        NativeImageUtils.checkVersion(requiredVersion, getVersionInformation());
+    }
+
+    protected boolean isOracleGraalVM() throws MojoExecutionException {
+        return getVersionInformation().contains(ORACLE_GRAALVM_IDENTIFIER);
```
melix, line 447: FYI: This is related to a very long time discussion we have on identifying GraalVM properly, and more specifically editions. This has changed several times in the past, this is likely not accurate.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java
```diff
@@ -175,7 +182,8 @@ private Path getDefaultRepo(Path destinationRoot) {
             try {
                 targetUrl = new URI(metadataUrl).toURL();
                 // TODO investigate if the following line is necessary (Issue: https://github.com/graalvm/native-build-tools/issues/560)
-                metadataRepositoryConfiguration.setUrl(targetUrl);
+                // TODO: forced to remove this to allow building on my machine. Remove this before merging.
+//                metadataRepositoryConfiguration.setUrl(targetUrl);
```
melix, line 186: :x: This needs to be uncommented. There shouldn't be any need for this hacks for local builds either.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java
```diff
@@ -208,7 +216,8 @@ private Path getRepo(Path destinationRoot) {
                     try {
                         targetUrl = new URI(metadataUrl).toURL();
                         // TODO investigate if the following line is necessary (Issue: https://github.com/graalvm/native-build-tools/issues/560)
-                        metadataRepositoryConfiguration.setUrl(targetUrl);
+                        // TODO: forced to remove this to allow building on my machine. Remove this before merging.
+//                        metadataRepositoryConfiguration.setUrl(targetUrl);
```
melix, line 220: :x: Same here, this needs to be reintroduced.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
```diff
@@ -74,6 +77,17 @@ public class NativeCompileNoForkMojo extends AbstractNativeImageMojo {
     @Parameter(property = "skipNativeBuildForPom", defaultValue = "false")
     private boolean skipNativeBuildForPom;
 
+    /**
+     * Used in {@link NativeCompileNoForkMojo#generateAugmentedSBOMIfNeeded} to determine if an augmented SBOM should
+     * be produced by {@link SBOMGenerator}.
+     */
+    @Parameter
+    private Boolean augmentedSBOM;
+    public static final String augmentedSBOMParamName = "augmentedSBOM";
```
melix, line 86: :x: This needs to be capitalized: `AUGMENTED_SBOM_PARAM_NAME`

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
```diff
@@ -101,9 +115,77 @@ public void execute() throws MojoExecutionException {
...
+     * @throws MojoExecutionException if augmented SBOM generation was attempted but failed.
+     */
+    private void generateAugmentedSBOMIfNeeded() throws IllegalArgumentException, MojoExecutionException {
+        boolean optionWasSet = augmentedSBOM != null;
+        augmentedSBOM = optionWasSet ? augmentedSBOM : true;
+
+        int detectedJDKVersion = NativeImageUtils.getMajorJDKVersion(getVersionInformation());
+        String sbomNativeImageFlag = "--enable-sbom";
+        boolean sbomEnabledForNativeImage = getBuildArgs().stream().anyMatch(v -> v.contains(sbomNativeImageFlag));
+        if (optionWasSet) {
```
melix, line 147: :question: Can you clarify the intent here? Why would we have to set both `--enable-sbom` explicitly and augmented sbom flag?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -0,0 +1,283 @@
...
+     * @throws MojoExecutionException if SBOM creation fails.
+     */
+    public void generate() throws MojoExecutionException {
+        String outputDirectory = mavenProject.getBuild().getDirectory();
+        Path sbomPath = Paths.get(outputDirectory, SBOM_NAME + "." + FILE_FORMAT);
+        try {
+            /* Suppress the output from the cyclonedx-maven-plugin. */
+            int loggingLevel = logger.getThreshold();
+            logger.setThreshold(Logger.LEVEL_DISABLED);
+            executeMojo(
```
melix, line 164: :x: This should ideally be done without using an external plugin to invoke the mojo. We do this already in several plugins.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -0,0 +1,283 @@
...
+                    executionEnvironment(mavenProject, mavenSession, pluginManager)
+            );
+            logger.setThreshold(loggingLevel);
+
+
+            if (!Files.exists(sbomPath)) {
+                return;
+            }
+
+            // TODO: debugging, remove before merge
```
melix, line 186: :x: Make sure all TODOS are resolved before merging

melix (changes requested):
This is a rather involved PR and great work already. I think I understand the goal now, which should be clearer. In particular, in the beginning, I thought the goal was to provide information _to_ native image so that it generates a better SBOM, but it's actually the other way around. The idea is to generate a native image, which, in turn, may help getting a better SBOM. I think we miss integration tests though, and I'd like to see this working on a real world project (e.g a Micronaut one) to make sure that it doesn't break anything.

I'm not sure that we should _augment_ existing SBOMs, in the sense of overwriting. Maybe it's my Gradle's mindset here, but in general, we should avoid overwriting things and better generate _alongside_. Also, this feature is very costly (requires building native image, then scanning packages), so this should probably be documented as as such. Which also means that maybe this shouldn't be part of the default lifecycle.

Regarding a Gradle implementation: the Gradle version should make sure to isolate the cyclonedx plugin, by resolving it at runtime using the worker API. This will avoid conflicts with existing configuration. In addition, this raises the question of how SBOMs are generated as artifacts to be published. It's unclear to me where they are going to be published (Maven Repository) and probably we don't want to make publishing of regular Java artifacts dependent on something as slow as building a native image. Or maybe not, this needs discussion. In any case, I'd be in general worried that this feature makes builds slower because of implicit dependencies (e.g publishing a jar would systematically require building a native image). Maybe what we need is a design document which clarifies both the intent and process. Note that for scanning packages, since it's fairly expensive, I have implemented something similar for Gradle on the layered images branch as a [shared library between Maven and Gradle](https://github.com/graalvm/native-build-tools/blob/cc/layered-images/common/utils/src/main/java/org/graalvm/buildtools/utils/JarScanner.java), which then Gradle uses in an [artifact transform](https://github.com/graalvm/native-build-tools/blob/cc/layered-images/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/scanner/JarAnalyzerTransform.java) which makes it much faster because cacheable.

native-maven-plugin/build.gradle.kts
melix: Please let me know the errors you are seeing, there shouldn't be any need for commenting these out.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
melix: This is directly related to the [long standing discussion](https://github.com/graalvm/native-build-tools/issues/357) about toolchain support. We need a _reliable_ way to make a difference between vendors, editions and features included in a release.

rudsberg:
@melix Thank you for your review and comments. 

>  I think I understand the goal now, which should be clearer. In particular, in the beginning, I thought the goal was to provide information to native image so that it generates a better SBOM, but it's actually the other way around. The idea is to generate a native image, which, in turn, may help getting a better SBOM.

Yes, the idea is to provide native image with more information so that it can generate a better SBOM. This feature simply hooks into the the regular `native-maven-plugin` build process that creates native images by creating an SBOM that contains more information _before_ the native image builder is invoked. This is reflected in `NativeCompileNoForkMojo.execute` where the call to `generateAugmentedSBOMIfNeeded` is made before the call to `buildImage`. 

The bullet list of the high-level approach in the PR description _sequentially_ describes the steps. 

> I think we miss integration tests though, and I'd like to see this working on a real world project (e.g a Micronaut one) to make sure that it doesn't break anything.

I can definitely add integration tests. We have integration tests already on the Native Image side, but adding it here too would be beneficial. 

> I'm not sure that we should augment existing SBOMs, in the sense of overwriting. Maybe it's my Gradle's mindset here, but in general, we should avoid overwriting things and better generate alongside.

The changes that are made to the baseline SBOM (the one generated by `cyclonedx-maven-plugin`) is 
* the pruning of components and dependencies that are not part of the image, as deduced from the static analysis.
* the inclusion of native-image specific components that is not found by the baseline SBOM, such as `svm` and `nativeimage`.

Thus, we remove unreachable components/dependencies and augment it with components to make the SBOM reflect the true bill-of-materials as seen from the native image. In other words, it's more _augmenting_ and _deleting_ of unnecessary data than it is overwriting. If you see any other approach that would be better, please elaborate.

> Also, this feature is very costly (requires building native image, then scanning packages), so this should probably be documented as as such. Which also means that maybe this shouldn't be part of the default lifecycle.

This should be covered by my earlier comments. The overhead of this feature is only the call to `generateAugmentedSBOMIfNeeded`.  Executing `./mvnw package -Dpackaging=native-image` (creating an image and an augmented SBOM by using `--enable-sbom=export,embed` as a buildArg) in the [micronaut-hello-rest-maven](https://github.com/graalvm/graalvm-demos/tree/master/micronaut-hello-rest-maven) takes around 28s on my machine (M3 Max), out of which the call to `generateAugmentedSBOMIfNeeded` takes between 550ms and 1300ms. This overhead is not very costly. 

> Regarding a Gradle implementation: the Gradle version should make sure to isolate the cyclonedx plugin, by resolving it at runtime using the worker API. This will avoid conflicts with existing configuration. 

Great to hear we can isolate the cyclonedx plugin on Gradle using the worker API. 

> In addition, this raises the question of how SBOMs are generated as artifacts to be published. It's unclear to me where they are going to be published (Maven Repository) and probably we don't want to make publishing of regular Java artifacts dependent on something as slow as building a native image. Or maybe not, this needs discussion. In any case, I'd be in general worried that this feature makes builds slower because of implicit dependencies (e.g publishing a jar would systematically require building a native image).

I'm afraid I'm not following here. SBOMs are either embedded in the image, exported as a JSON in the build directory, or added as a resource to the classpath. Thereafter, users likely run it through some vulnerability scanner. 

Maybe I'm misunderstanding your point, but I'm afraid I don't understand what you mean by publishing regular Java artifacts being dependent on building native images and how this feature causes such issues. To me that's two independent processes, and only when user opts in to building images are images built.

> Note that for scanning packages, since it's fairly expensive, I have implemented something similar for Gradle on the layered images branch as a [shared library between Maven and Gradle](https://github.com/graalvm/native-build-tools/blob/cc/layered-images/common/utils/src/main/java/org/graalvm/buildtools/utils/JarScanner.java), which then Gradle uses in an [artifact transform](https://github.com/graalvm/native-build-tools/blob/cc/layered-images/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/scanner/JarAnalyzerTransform.java) which makes it much faster because cacheable.

OK, great. I'll take a look at that.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
rudsberg: @fniephaus Maybe you can chime in on this. We need a (hopefully reliable) way to know if the user is using the EE version of native image. Currently, the `isOracleGraalVM` function does this by calling `native-image --version` (that's the `getVersionInformation()` function) and checking the presence of the string `"Oracle GraalVM"`.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeMojo.java
rudsberg: Totally agree! Hopefully we can resolve this.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
rudsberg: Ah, great! Could you give me some pointers?

gradle/libs.versions.toml
rudsberg: Update: saw your [reply here](https://github.com/graalvm/native-build-tools/pull/623#discussion_r1787338677) that invoking plugins is done elsewhere in the codebase.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
rudsberg: Setting both `--enable-sbom` and the augmented sbom flag is not required. We check `optionWasSet` since the behavior is dependent on if the user explicitly set the augmented flag option. I tried to make the intent clear via the javadoc, variable names, and comments. What exactly was unclear?

native-maven-plugin/build.gradle.kts
rudsberg: Using Java 11 installation for compilation and GraalVM 17 for testing resolved all these issues ✅

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
melix: Actually made a mistake, the code I had in mind (Micronaut Maven plugin) is using the same plugin :man_facepalming: So consider my comment resolved!

gradle/libs.versions.toml
olpaw: Would it be an option to not require `org.cyclonedx:cyclonedx-maven-plugin` here and only make your SBOM magic happen if the user explicitly adds the `org.cyclonedx:cyclonedx-maven-plugin` in his pom-file. This would then also serve as a way to enable/disable sbom specific processing in this plugin based on the users pom-file. Or is this just not the way to do things in maven?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
rudsberg: Superseded by [this thread](https://github.com/graalvm/native-build-tools/pull/623#discussion_r1791567680).

rudsberg:
[Integration tests](https://github.com/graalvm/native-build-tools/pull/623/files#diff-b95eb0f8538a4247d14c37bce95d138af32c444fc08d8f8ad8140c15693cd63d) have now been added. These tests focus on
* error handling (using CE or non-supported EE versions), and
* that the basic behavior is observed using the 'java-application'.

melix (dismissed):

melix (approved):

PR #624, opened 2024-09-18
Add Getting Started guide

dnestoro:
This PR adds a guide that shows how newcomers can start using Native Gradle Plugin


jeanbisutti (commented):
Great! I would have been happy to have this getting started in the past.

Perhaps it would be worth to also mention theses JUnit 5 configurations: 
 https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f0c25aecd006262929668f04d836309e3b3b071b/smoke-tests-otel-starter/spring-boot-reactive-3/build.gradle.kts#L65

It would be great if you could create a Github repository with the getting started code.

GradleGettingStarted.md
```diff
@@ -0,0 +1,414 @@
...
+
+You can learn more about `Reachability metadata` support and see the examples, [here](https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html#_configuring_the_metadata_repository).
+
+
+## Additional explanations
+
+### Agent filter file
+
+Filter files that agent consumes in the `conditional` mode have the following structure:
+```
```
vjovanov, line 280: Use ```json

GradleGettingStarted.md
```diff
@@ -0,0 +1,414 @@
...
+    {"includeClasses": "**"}
+  ]
+}
+```
+This filter file will instruct the agent to include everything and therefore, you will get a massive config files.
+For example this is how `resource-config.json` looks like for a very simple test:
+```
+{{
+  "resources":{
+  "includes":[{
```
vjovanov, line 311: We should use the new format here.

melix:
I am extremely confused. We already have a getting started guide. It's even mentioned in the README and is published alongside the release: https://graalvm.github.io/native-build-tools/latest/gradle-plugin-quickstart.html

Why are we adding _another one_ instead, if needed, to fix the other one?

jeanbisutti:
The getting started of this PR provides useful additional information, such an agent example with the conditional mode.

melix:
But why creating a new guide? The old one should be fixed. Now we have 2, and this one is inconsistently published (.md file instead of generated, versioned docs).

jeanbisutti:
> But why creating a new guide? The old one should be fixed. Now we have 2, and this one is inconsistently published (.md file instead of generated, versioned docs).

Agree

vjovanov:
This is a simplified guide where one comes and sees the exact steps they need to do to support a library.

What other guide do we have? I am aware of our docs, but I feel they are more of reference manual.

melix:
> What other guide do we have?

On the GitHub home page, we have:

image

This links to https://graalvm.github.io/native-build-tools/latest/gradle-plugin-quickstart.html

which is also crossed referenced from the reference documentation: https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html#quickstart

Both quickstart guide and reference docs are generated by asciidoctor: https://github.com/graalvm/native-build-tools/tree/db65d975c6e05ccd1f5d9f0bf7ddb163e7866691/docs/src/docs/asciidoc

dnestoro:
I think that we can fuse this guide with the existing one. Existing one misses few points explained in this guide (like agent modes, how to pass runtime/buildtime options, slightly more details about `metadataCopy` task, how to run tests in the CI...). 

Another important point is accessibility of the guide. I think it should be one of the first items on the [main docs page](https://graalvm.github.io/native-build-tools/latest/). At the moment it looks a bit hidden.
cc @melix

melix:
Yes, I think the main page deserves a revamp. Especially now that there are so many releases, having the changelog as the TOC doesn't quite make sense anymore.

dnestoro:
Changes included in: https://github.com/graalvm/native-build-tools/pull/645

Issue #626, opened 2024-09-30
Sporadic Gradle failure:  access-filter.json already exists


jeanbisutti:
The Native Build Tools Gradle plugin is used in the [OpenTelemetry Java instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation) Github repository. 

Sometimes the plugin fails. A failure report example: https://ge.opentelemetry.io/s/2rq2vt7d7bsng/failure?expanded-stacktrace=WyIwLTEtMi0zLTQiXQ#1

The root case:

```
Caused by: java.nio.file.FileAlreadyExistsException: /tmp/agent-config
at org.graalvm.buildtools.agent.AgentConfiguration.addDefaultAccessFilter(AgentConfiguration.java:156)	
at org.graalvm.buildtools.agent.AgentConfiguration.getAgentCommandLine(AgentConfiguration.java:99)	
at org.gradle.api.internal.provider.ValueSupplier$Present.transform(ValueSupplier.java:519) 
```

The problem seems to come from the recent changes in the [AgentConfiguration](https://github.com/graalvm/native-build-tools/blob/c4305294bdf3519a2db2af1916c85689a3842467/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java#L55) class, and released in 0.10.3.

The native agent is not enabled from our [build.gradle.kts](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/f0c25aecd006262929668f04d836309e3b3b071b/smoke-tests-otel-starter/spring-boot-reactive-3/build.gradle.kts#L62) file or a command line (`./gradlew -Pagent`). Perhaps some code related to the agent could only be executed if the agent is enabled.

Would you have any workaround to recommend?

Thanks.

dnestoro:
Please use `0.10.2` version util we fix the problem. Thanks for reporting!

ghilainm:
I have the same kind of issue with the version 0.10.4. The processAot tasks is run in the context of a multi project Gradle build wih parallel feature enabled. It seems to be not supported. Can you confirm @dnestoro ?

```
Caused by: org.gradle.api.internal.provider.AbstractProperty$PropertyQueryException: Failed to calculate the value of property 'agentOptions'.
[...]

PR #635, opened 2024-10-31
Make accessFilter file registration thread safe

dnestoro:
This PR should fix https://github.com/graalvm/native-build-tools/issues/626

Described issue probably occurs in multi-threaded environment because the reported error throws `FileAlreadyExistsException`  even though [we checked first that file doesn't exist](https://github.com/graalvm/native-build-tools/blob/master/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java#L155). 

As per documentation of `java.nio.file.Files#createDirectory`:
> Creates a new directory. The check for the existence of the file and the creation of the directory if it does not exist are a single operation that is atomic with respect to all other filesystem activities that might affect the directory

Since we only care that directory with the given name exists, we can just catch and ignore `FileAlreadyExistsException` in this case.

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
@@ -40,21 +40,23 @@
  */
 package org.graalvm.buildtools.agent;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
-import java.nio.file.CopyOption;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.StandardCopyOption;
+import java.nio.file.*;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.logging.Logger;
 
 public class AgentConfiguration implements Serializable {
 
-    private static final String DEFAULT_ACCESS_FILTER_FILE = "/access-filter.json";
+    private static final String ACCESS_FILTER_PREFIX = "access-filter";
+    private static final String ACCESS_FILTER_SUFFIX = ".json";
+    private static final String DEFAULT_ACCESS_FILTER_FILE_LOCATION = "/" + ACCESS_FILTER_PREFIX + ACCESS_FILTER_SUFFIX;
+
+    private static final Logger logger = Logger.getGlobal();
+
     private final Collection<String> callerFilterFiles;
     private final Collection<String> accessFilterFiles;
     private final Boolean builtinCallerFilter;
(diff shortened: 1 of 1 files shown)


common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -142,25 +138,39 @@ private void addDefaultAccessFilter() {
             return;
         }
 
+        String accessFilterPrefix = "access-filter";
```
melix, line 141: :x: Should these be constants?

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -142,25 +138,39 @@ private void addDefaultAccessFilter() {
...
+            }
+
+            Path tmpAccessFilter = Files.createTempFile(agentDir, accessFilterPrefix, accessFilterSuffix);
+            Files.copy(accessFilterData, tmpAccessFilter);
+
+            try {
+                Files.move(tmpAccessFilter, accessFilterFile, StandardCopyOption.ATOMIC_MOVE);
+            } catch (FileAlreadyExistsException e) {
+                Files.delete(tmpAccessFilter);
+                System.out.println("Access-filter file already exists. Delete temporary one.");
```
melix, line 170: :x: Either use a logger or remove the debug statement.

melix (changes requested):

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -144,23 +146,43 @@ private void addDefaultAccessFilter() {
...
+                logger.info("Ignore directory creation because " + agentDir + " directory is already created.");
+            }
 
-                Files.copy(accessFilter, accessFilterFile, StandardCopyOption.REPLACE_EXISTING);
-                accessFilterFiles.add(accessFilterFile.toString());
-            } else {
-                throw new IOException("Cannot find access-filter.json on default location: " + DEFAULT_ACCESS_FILTER_FILE);
+            Path tmpAccessFilter = null;
+            try {
+                 tmpAccessFilter = Files.createTempFile(agentDir, ACCESS_FILTER_PREFIX, ACCESS_FILTER_SUFFIX);
```
olpaw, line 168: indentation is not aligned

olpaw, line 168: You mentioned that his gave you:
```
Caused by: java.nio.file.FileAlreadyExistsException: /tmp/agent-config/access-filter21830177911714255.json
```
Why is this not an issue anymore?

olpaw: Maybe it's better to create the temp name differently. A combination of PID + UTC timestamp would be good.

olpaw (changes requested):

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -163,17 +163,10 @@ private void addDefaultAccessFilter() {
...
-            } catch (FileAlreadyExistsException e) {
-                logger.info("Temporary access-filter file already exists and we should just rename it.");
-            }
-
-            if (tmpAccessFilter == null) {
-                throw new IOException("Cannot create temporary access-filter file.");
-            }
+            long pid = ProcessHandle.current().pid();
+            long time = System.currentTimeMillis();
+            Path tmpAccessFilter = agentDir.resolve(ACCESS_FILTER_PREFIX + pid + time + ACCESS_FILTER_SUFFIX);
```
olpaw, line 168:
```suggestion
            Path tmpAccessFilter = agentDir.resolve(ACCESS_FILTER_PREFIX + '_' + pid  + '_' + time  + '_' + ACCESS_FILTER_SUFFIX);
```

olpaw (dismissed):
- [x]  Adjust the logging and error messages before merging.

melix (approved):

PR #645, opened 2024-12-02
Refactor Documentation

dnestoro:
This PR:

- restyles existing docs pages
- adds end-to-end guides for both maven and gradle
- revisits existing options in the documentation


docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+    }
+}
+----
+
+[[collect-metadata]]
+== Collecting metadata
+
+When your test/application starts to be a bit more complex things like **reflection**, **resources**, **serialization**, **proxies** or **jni** may be required.
+Since the Native Image has closed world assumption, all of these things must be known in advance during the image build.
+The easiest way how this information can be passed to the Native Image is through metadata config file(s) - depending on the GraalVM version you are using, there could be
```
melix, line 258:
```suggestion
The easiest way to pass this information to the Native Image is through configuration file(s) - depending on the GraalVM version you are using, there could be
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+a single `reachability-metadata.json` file (for newer GraalVM versions) or multiple json files (`reflect-config.json`, `resource-config.json`, `proxy-config.json`, `serialization-config.json`, `jni-config.json`).
+To learn more about metadata that Native Image consumes, https://www.graalvm.org/latest/reference-manual/native-image/metadata/[see this].
+
+For example, if you run the test that uses the reflection, and you don't have entry for that reflection call in the metadata config file, your test will fail.
```
melix, line 262:
```suggestion
For example, if you run tests which make use of reflection, and you don't have entry for reflective calls in the metadata config file, your native tests will fail.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+
+To make your test/application work while using reflection (like in this example) or other metadata, you should either generate metadata configurations or write them manually.
```
melix, line 264:
```suggestion
In order to make your test/application work natively while using reflection (like in this example) or other features which require additional metadata, you should either generate configuration files or write them manually.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+To generate metadata automatically, you can run your tests (or the main application) with the Native Image Agent, that will collect all the metadata your test/application require.
```
melix, line 265:
```suggestion
The easiest is to generate metadata automatically.
For this you can run your tests (or the main application) with the Native Image Agent, which will collect all the metadata your test/application require and generate metadata for you.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+To enable the agent (through Native Gradle Plugin) you should either:
+
+- add `-Pagent` flag to the command you are executing
```
melix, line 268:
```suggestion
- add `-Pagent` flag to the command line (for example `./gradlew test -Pagent`)
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+
+
+[.underline]#To generate the metadata file(s) for your `application` just run:#
+
+- `./gradlew run` if you added the agent block to the configuration or `./gradlew -Pagent run` if you didn't. This command runs on JVM with native-image-agent and collects the metadata.
+
+[WARNING]
+====
+Unless you specify the following block in your `build.gradle` file, Gradle will pick up the Agent from the `JAVA_HOME` environment variable.
+This may cause problems if you set `GRAALVM_HOME` environment variable because your project will be build with one version and the Agent will generate metadata for the other one.
```
melix, line 299:
```suggestion
This may cause problems if you set `GRAALVM_HOME` environment variable because your project will be built with one version and the Agent will generate metadata for the another one.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+- `./gradlew nativeTest` if you added the agent block to the configuration or `./gradlew -Pagent nativeTest` if you didn't. This command runs on JVM with the native-image agent, collects the metadata and uses it for testing on native-image.
+
+
+[.underline]#To generate the metadata file(s) for your `application` just run:#
+
+- `./gradlew run` if you added the agent block to the configuration or `./gradlew -Pagent run` if you didn't. This command runs on JVM with native-image-agent and collects the metadata.
+
+[WARNING]
+====
+Unless you specify the following block in your `build.gradle` file, Gradle will pick up the Agent from the `JAVA_HOME` environment variable.
```
melix, line 298: :question: Are you sure about this? Isn't it the problem that the application may be _compiled_ with one Java version, and that the agent uses the version from `GRAALVM_HOME`? That would be what I expect.

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+    "$it/bin/java"
+}.get()
+----
+
+This way, generated metadata config file(s) will have format that is specified in the GraalVM version (from `GRAALVM_HOME`) you are using.
+====
+
+
+[[metadata-copy]]
+=== Move generated metadata to non-default location
```
melix, line 313:
```suggestion
=== Copy generated metadata to permanent location
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+
+By default, generated metadata will be placed inside `build/native/agent-output` directory.
+In many cases you may want to move generated metadata to some other location.
```
melix, line 316:
```suggestion
In many cases you may want to copy generated metadata to some other location.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+To do so, you can configure and run `metadataCopy` task.
+
+==== Configure metadataCopy task
+
+First, you can configure `metadataCopy` task by adding a new block, named `metadataCopy` inside `agent` block that you added in the previous step.
+Inside this block, you can specify:
+
+- `outputDirectories` - location where you want to move the metadata
```
melix, line 324:
```suggestion
- `outputDirectories` - location where you want to copy the generated metadata
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+Here is an example of a valid `metadataCopy` usage:
+
+[source,bash,subs="verbatim,attributes", role="multi-language-sample"]
+----
+./gradlew metadataCopy --task test --dir resources/META-INF/native-image/org.example
+----
+
+[WARNING]
+====
+Note that **if you store generated metadata on location other than the default one**, you will need to pass that location as a Native Image `buildArg` with `-H:ConfigurationFileDirectories` option.
```
melix, line 389:
```suggestion
Note that **if you store generated metadata outside of the standard resources directories (e.g `src/main/resources/META-INF/native-image/...`) **, you will need to pass that location as a Native Image `buildArg` with `-H:ConfigurationFileDirectories` option.
In this case Gradle will _not_ be able to detect changes to metadata properly, which is why we recommend to use standard locations.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+** `extraFilterPath` - (optional) extra filter used to further filter the collected metadata. See <<agent-filter-file, the following section>>
+* `direct` - in this mode user configures the agent completely manually by adding all options with:
+** `options.add("<option>")`
+
+Each of the described modes has its own benefits.
+For example:
+
+- `standard` mode is a **great starting point** in your project development
+- `conditional` mode is mainly aimed towards **library maintainers** with the goal of reducing overall footprint
+- `direct` mode is for **experienced users** that knows how to configure the agent manually
```
melix, line 427:
```suggestion
- `direct` mode is for **experienced users** who know how to configure the agent manually
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+Here is an example of the `agent` block with configured conditional and direct modes, where the conditional mode is set as default and will be used to generate the metadata:
+
+[source,groovy,subs="verbatim,attributes", role="multi-language-sample"]
+----
+agent {
+    enabled = true
+    defaultMode = "conditional"
+    modes {
+        conditional {
+            userCodeFilterPath = "src/test/resources/metadata.with.starting.user-code-filter/user-code-filter.json"
```
melix, line 439: :nit: Technically speaking, files found in `src/test/resources` are found on classpath at runtime (here, for tests). This doesn't quite make sense to have the user code filter at runtime, so I would personally use a different directory, for example `src/test/native-image/...`

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+- builtinCallerFilter
+- builtinHeuristicFilter
+- enableExperimentalPredefinedClasses
+- enableExperimentalUnsafeAllocationTracing
+- trackReflectionMetadata
+
+[WARNING]
+**These options are for advanced usages, and you can read more about them https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/#agent-advanced-usage[here]**.
+
+Complete example of the agent block should look like this:
```
melix, line 491:
```suggestion
A complete example of the agent block should look like this:
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+    {"excludeClasses": "regex\.example\.exclude[0-9]+"},
+  ]
+}
+----
+
+The process how you can pass the config files to the agent is described in the <<additional-agent-options,previous section>>.
+
+We can see on the example how different filter files affect generated metadata:
+
+We are starting with the simple filter file:
```
melix, line 584:
```suggestion
Let's start with the simple filter file:
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+[source,json,subs="verbatim,attributes", role="multi-language-sample"]
+----
+{
+  "rules": [
+    {"includeClasses": "**"}
+  ]
+}
+----
+
+This filter file will instruct the agent to include everything and therefore, you will get a massive config file.
```
melix, line 595:
```suggestion
This filter file will instruct the agent to include everything, which will result in a massive configuration file.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+              "java.lang.String"
+            ]
+          }
+        ]
+      }
+    ]
+  }
+----
+
+As you can see, there are lots of entries that you may don't want.
```
melix, line 671:
```suggestion
As you can see, there are lots of entries that you likely don't want.
They are present because the metadata was generated using tests, for example, and that the testing library was present at runtime during the agent execution pass, but in practice, your application will not use the testing library in production.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+    {"excludeClasses": "worker.org.gradle.**"},
+    {"excludeClasses": "org.slf4j.**"},
+    {"excludeClasses": "java.**"}
+  ]
+}
+----
+
+[WARNING]
+====
+Be careful to remove only redundant metadata and keep metadata necessary for your project.
```
melix, line 690:
```suggestion
Always be careful when removing entries in metadata, as this may result in a broken binary.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+====
+
+After we regenerate the metadata with the new filter, `reachability-metadata.json` generated on the same example as above will look like this:
```
melix, line 693:
```suggestion
After updating our filters, we can regenerate the metadata, which will result in the following `reachability-metadata.json` file:
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+}
+----
+
+As you can see there are no more entries that contain classes from `org.junit` (as their condition) for example.
+
+
+[[maintain-generated-metadata]]
+== Maintain generated metadata
+
+If you are a library maintainer, or your application became huge, you may consider covering most of your code with tests.
```
melix, line 737:
```suggestion
If you are a library maintainer, or if your application grows large, you may consider covering most of your code with tests.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+This way you can also track if your project requires updates of the existing metadata.
+Considering that you run your tests in certain CI, at some point (after updating some dependency or adding new feature/test) you may notice some test failures with missing resources errors or that something is reflectively missing...
```
melix, line 739:
```suggestion
Considering that you run your tests on CI, at some point (after updating some dependency or adding new feature/test) you may notice some test failures with missing resources errors or that something is reflectively missing...
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+In that case, your metadata requires update.
+
+[WARNING]
+Be aware that if you modified existing metadata file(s) on the default location, generating a new metadata will overwrite the existing one and your manual changes will be lost.
+
+So if you modified existing metadata file(s) on the default location, please do the following:
+
+1. Move your metadata to some other, non-default location (with the `metadataCopy` task for example)
```
melix, line 747: :question: why is this step required if `mergeWithExisting` is true?

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+3. Run your tests again to generate new metadata (as we already described in the <<collect-metadata, collect metadata section>>)
+4. Run `metadataCopy` task again (with `mergeWithExisting` set to true in step 2)
+
+This way you will keep your original metadata, and add a new one.
+
+
+[[reachability-metadata-repository]]
+== Reachability metadata repository
+
+Native Build Tools (both Gradle and Maven plugins) picks metadata from Reachability metadata repository to ensure your application works out-of-box (if all metadata required by your app is already contributed to the metadata repository).
```
melix, line 758:
```suggestion
While we have described how you can add metadata for your own code, Native Build Tools (both Gradle and Maven plugins) makes use of metadata from Reachability Metadata Repository to ensure your application works out-of-box with 3rd party libraries that you depend on.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+Furthermore, you can configure Reachability metadata support through `metadataRepository` block added to our main `graalvmNative` block inside `build.gradle`.
```
melix, line 759:
```suggestion
You can configure Reachability metadata support through `metadataRepository` block added to our main `graalvmNative` block inside `build.gradle`.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+Most common options you may want to configure in this block are:
+
+* `enabled` - determines if you want to use Reachability metadata support or not (`true` by default)
+* `version` - specifies exact Reachability metadata version you want to use
+* `uri` - specifies the url where the metadata is stored. This can be used to point to the local repository
```
melix, line 764:
```suggestion
* `uri` - specifies the url where the metadata is stored. This can be used to point to the local repository. If not set, it defaults to metadata which is shipped inside the Native Build Tools.
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+
+YYou can read more about __Reachability metadata support__ and other (advanced) configuring options, https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html#_configuring_the_metadata_repository[here].
```
melix, line 766:
```suggestion
You can read more about __Reachability metadata support__ and other (advanced) configuring options, https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html#_configuring_the_metadata_repository[here].
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+----
+./gradlew scaffold --coordinates com.example:my-library:1.0.0
+----
+
+- implement tests in test stubs that will show how you have generated metadata
+- collect metadata as described https://github.com/oracle/graalvm-reachability-metadata/blob/master/docs/CollectingMetadata.md#collecting-metadata-for-a-library[here]
+- create a pull request and fill the checklist
+
+[[track-diagnostics]]
+== Track diagnostics
```
melix, line 790:
```suggestion
== Using diagnostics
```

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+
+If you want to explore details about native images you are generating, you can add:
+
+[source,bash,subs="verbatim,attributes", role="multi-language-sample"]
+----
+buildArgs.add("--emit build-report")
```
melix, line 796: :?: if it is really `--emit build-report`, then it should be 2 build args: `--emit`, `build-report`.

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -0,0 +1,871 @@
...
+----
+For GraalVM versions starting from the GraalVM for JDK23
```
melix, line 798:
```suggestion
For GraalVM versions starting from GraalVM for JDK 23
```

docs/src/docs/asciidoc/end-to-end-maven-guide.adoc
```diff
@@ -0,0 +1,830 @@
...
+The GraalVM team
+:highlighjsdir: {maven-relative-srcdir}/highlight
+
+
+This guide leads you through the process of integrating Native Maven Plugin with your project.
+It starts from adding the plugin to your project, goes through some of the main functionalities that you may use (like collecting the metadata),
+all the way to some diagnostics tools that you can use to analyse native executables you have made.
+You can see an example of tests on which we tested whole guide <<example,here>>
+
+In case you only want to see how a simple application works in practise, you can check <<quickstart-maven-plugin.adoc#,our demo>>.
```
melix, line 11:
```suggestion
In case you only want to see how a simple application works in practise, you can check <<quickstart-maven-plugin.adoc#,our demo>>.
```

docs/src/docs/asciidoc/end-to-end-maven-guide.adoc
```diff
@@ -0,0 +1,830 @@
...
+
+[[prerequisites]]
+== Prerequisites
+
+- Before starting, make sure that you have GraalVM installed locally. You can https://www.graalvm.org/downloads/[download Oracle GraalVM from the official website].
+- You should set `JAVA_HOME` environment variable to point to your GraalVM installation.
+
+[[adding-plugin]]
+== Adding plugin
```
melix, line 20:
```suggestion
== Adding the plugin
```

melix (changes requested):
Great work David!

I have mostly made changes to the Gradle end-to-end guide. I think most comments can be copied to the Maven guide which reuses the same sentences.

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
dnestoro: Because we are executing `metadataCopy` as a separate task. When you run the agent, it will always overwrite existing metadata on the default location. By the time you want to execute `metadataCopy` (even with `mergeWithExisting` set to true) your manual changes are lost.

To avoid this:

-  you should copy metadata onto the "safe" location
- generate new metadata that will overwrite the original metadata
- then, when you execute `metadataCopy` with `mergeWithExisting` set to true, your new metadata will be merged with the old metadata placed on the "safe" location

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
dnestoro: Of course we can also separate this and it will work, but in my opinion it is more understandable if we group argument with its value. It will work both ways.

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
melix: > When you run the agent, it will always overwrite existing metadata on the default location

It shouldn't. The agent will output the metadata in a temporary directory always. It shouldn't matter how `metadataCopy` is configured. If it's the case, then it's a bug.

dnestoro: There is a misunderstanding then... When I say **default location**, in this case, I mean `temporary directory` which is the default location where agent prints metadata if you don't do anything additionally. Maybe I should assume that at this point users already moved metadata to non-temporary (or _"non-default"_) location?

dnestoro: Maybe I should also add some kind of tip in the `metadataCopy` section that it is strongly recommended to move metadata from the temporary directory to standard location

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
dnestoro: Discussed offline. Please take a look again

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
dnestoro: I guess it makes more sense now?

melix: Right, the confusion came from the fact I assumed that the metadata was already copied. I think we should recommend this indeed.

docs/src/docs/asciidoc/index.adoc
```diff
@@ -1,496 +1,37 @@
 = Native Build Tools
 The GraalVM team
 
-The {doctitle} project provides plugins for different build tools to add support for building and testing native applications written in Java (or any other language compiled to JVM bytecode).
+The {doctitle} project provides plugins for different build tools to add support for building and testing native applications written in Java
```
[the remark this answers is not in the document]
olyagpl: I didn't hear about "native projects", but "native applications"

docs/src/docs/asciidoc/index.adoc
```diff
@@ -1,496 +1,37 @@
...
-- Revamped Gradle plugin that is now a lot more idiomatic.
-- Fixes for several issues regarding JUnit testing.
-- Removal of Test Discovery mode from the Maven plugin.
-- Fix for Maven creating empty test images when no tests are present.
-- Added support for Kotlin tests in Gradle.
+[discrete]
+=== Documentation
 
-In addition to those improvements, several behind-the-scenes changes were made in order to ensure better compatibility moving forward:
+In case you are new to the Native Build Tools (either as an end user or a library author), read our:
```
olyagpl, line 16:
```suggestion
If you're new to the Native Build Tools, whether as an end user or a library author, check out our:
```

docs/src/docs/asciidoc/index.adoc
```diff
@@ -1,496 +1,37 @@
...
 
-- Test coverage has been greatly improved for all subprojects.
-- Build tooling for this repository has been improved significantly.
+- <<end-to-end-gradle-guide.adoc#,End-to-end Gradle guide>>
```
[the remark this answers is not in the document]
olyagpl: also we should change the file names accordingly

docs/src/docs/asciidoc/index.adoc
```diff
@@ -1,496 +1,37 @@
...
 
-- Fixes for most of the known issues regarding Gradle and Maven plugins
-- Massively improved automatic JUnit support as well as initial JUnit Vintage support
-- Improved JavaDoc and tests for the Gradle plugin
+[discrete]
+=== Contributing
 
-=== Release 0.9.0
+If you are interested in contributing or reporting an issue related to the plugins, please refer to the https://github.com/graalvm/native-build-tools[Native Build Tools repository].
+Also, be aware that all contributors must sign the https://oca.opensource.oracle.com/[Oracle Contributor Agreement] (OCA).
```
olyagpl, line 32:
```suggestion
```suggestion
Please note that all contributors must sign the https://oca.opensource.oracle.com/[Oracle Contributor Agreement] (OCA).
```

olyagpl (changes requested):
@dnestoro, at this point I reviewed the main entry page index.adoc and gradle-plugin.adoc. Could you please follow the comments or apply suggestions? Some of the comments in gradle-plugin.adoc apply to maven-plugin.adoc too. Then I will continue.

docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -161,8 +131,43 @@ include::../snippets/gradle/kotlin/build.gradle.kts[tags=select-toolchain]
...
 Again, be aware that the toolchain detection _cannot_ distinguish between GraalVM JDKs and standard JDKs without Native Image support: if you have both installed on the machine, Gradle may randomly pick one or the other.
 
-[[configuration-options]]
-==== Configuration options
+[[configuration]]
+== Configuration
+
+This plugin works with the `application` plugin and will register a number of tasks and extensions for you to configure.
+
+=== Available tasks
```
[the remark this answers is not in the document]
dnestoro: I am not sure if we can drop it. @melix what is your opinion?

docs/src/docs/asciidoc/quickstart-maven-plugin.adoc
olyagpl, line 1: @dnestoro, don't keep the quickstart-gradle-plugin.adoc and quickstart-gradle-maven.adoc content around. The point of this refactoring was to simplify getting started experience. Step 1: Provide demo example, a HelloWorld or very simple app. Step 2 how to add a plugin. Step 3: build and run.  Step 4, the last step how to add additional options (-Ob, -Os etc.). That's what end users need to see.  What I see in end-to-end-maven-guide.adoc belongs to the reference documentation that should be in maven-plugin.adoc.

dnestoro: But these are not getting started guides, but sequential end-to-end guides. I kept getting started guides as links at the top of end-to-end guides. 

Difference between end-to-end guides and reference docs is that I described only _necessary_ stuff that users may need, while reference documentation contains everything we support).

docs/src/docs/asciidoc/index.adoc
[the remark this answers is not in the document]
dnestoro: Addressed [here](https://github.com/graalvm/native-build-tools/pull/645#discussion_r1881964618)

docs/src/docs/asciidoc/quickstart-maven-plugin.adoc
olyagpl: After discussing the ideas over zoom with David and Vojin, here is a proposed beginning and a structure for the end-to-end Gradle guide:

```markdown
# Gradle Plugin User Guide

_The GraalVM team_ 
_version 0.10.4-SNAPSHOT_

This guide walks you through integrating the Gradle plugin for Native Image into your project. 
It starts from enabling the plugin, building the first native image, and running it.

Then it takes you to more advanced use-cases such us providing configuration.
If you are an advanced user, you can skip the getting started part and go directly to the advanced section. 

## Getting Started

To compile your application ahead of time with [GraalVM Native Image](https://www.graalvm.org/latest/reference-manual/native-image/) and Gradle, enable the Gradle plugin for Native Image building.

The plugin requires that you [install GraalVM](#link).

### Enable the Plugin

1. Add the plugin declaration into your _build.gradle_ file inside `plugins` block:
 
  id 'org.graalvm.buildtools.native' version '0.10.4-SNAPSHOT'

  Older versions of the plugin are listed [here](#).

2. In _build.gradle_, add the following block that configures the plugin:

    GroovyKotlin
    graalvmNative {
        graalvmNative {
            binaries.main {
                imageName = "yourName"
                mainClass = "org.example.Main"
                buildArgs.add('-Ob')
            }
        }
    }

  It specifies the path for the application main class (the main entry point) and what the name for a native executable file.
  Adjust this path according to your application sources.

  You can pass additional options to the Native Image tool using the `buildArgs.add('')` syntax.
  In this example, `-Ob` enables quick build mode.
  Learn more in [Providing Configuration Options](#link).

### Build and Run Your Application

> Note: A Java version between 17 and 21 is required to execute Gradle. (See the [Gradle Compatibility Matrix](https://docs.gradle.org/current/userguide/compatibility.html)).

1. Compile and build a native executable of your application:

  ./gradlew nativeCompile

  The executable is created in the _app/build/native/nativeCompile/_ directory of the project.

2. Run the application from the native executable:

    ./gradlew nativeRun

The Gradle plugin for Native Image registers more tasks that you may want to use, for example, `nativeTest` that executes tests found in the test sources in native mode. 
Continue reading below to learn more about the plugin.

## Advanced Use Cases

For advanced users, the guide explains how to collect metadata, use diagnostics tools to analyze native executables, and includes examples of tests that demonstrate the guide's concepts.

### Providing Configuration Options

### Collecting Metadata with Tracing Agent

### Collecting Configuration from Reachability Metadata Repository

### Maintaining Generated Metadata 

### Enabling Monitoring Tools

## Summary
```

olyagpl (approved):
@dnestoro, I'm done proofreading your latest version of [end-to-end-gradle-guide.adoc](https://github.com/graalvm/native-build-tools/pull/645/files#diff-a347972e886baf0c4e6689a951bd55fa2c0eca26faf0ddb140b9902d5a6316c3).

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
```diff
@@ -28,13 +33,27 @@ The plugin requires that you https://www.graalvm.org/downloads/[install GraalVM]
   id("org.graalvm.buildtools.native") version "{gradle-plugin-version}"
 ----
 
-All plugin versions are listed https://github.com/graalvm/native-build-tools/releases[here]
+Replace `{maven-plugin-version}` with the latest released version.
+All plugin versions are listed https://github.com/graalvm/native-build-tools/releases[here].
+
+
+- Provide the application main class in _build.gradle_ / _build.gradle.kts_ as shown below.
```
olyagpl, line 40: @dnestoro, is it a must-have step to enable the Gradle plugin? Also, is it still required to  declare an additional plugin repository in settings.gradle? 

```
 pluginManagement {
     repositories {
         mavenCentral()
         gradlePluginPortal()
     }
 }
```

olyagpl (approved):

docs/src/docs/asciidoc/end-to-end-gradle-guide.adoc
dnestoro: No, I was able to run few reproducers without that.

olyagpl (approved):

olyagpl (approved):

dnestoro:
@loicottet @vjovanov I have a question about detecting missing metadata: 

With GraalVM 23:

- we pass `--exact-reachability-metadata` as a build argument
- and `-XX:MissingRegistrationReportingMode=Warn` as a runtime argument

this way we get a list of missing metadata.

Is it possible to do the same with earlier GraalVM releases (like GraalVM 21)? If so, what are the options we should pass?

olyagpl (approved):

melix (approved):

PR #657, opened 2024-12-17
Fix checkstyle errors

sdeleuze:


snicoll (approved):

dnestoro:
Yes, I think we should. Lets make an issue for this as I can't work on this immediately.

Issue #671, opened 2025-01-12
Maven functional test (sbom) fails with the latest ea release


dnestoro:
**Describe the bug**
When running Maven functional tests with the [latest ea release](https://github.com/graalvm/oracle-graalvm-ea-builds/releases/tag/jdk-25.0.0-ea.04), sbom test: `sbom is created when only the augmented sbom parameter is used (but not the '--enable-sbom' buildArg)` fails with the following:

```
Condition not satisfied:

validateSbom sbom
|            |
false        /tmp/spock_sbom_is_created_whe_0_testDirectory3060895557523752692/target/example-app.sbom.json

	at org.graalvm.buildtools.maven.SBOMFunctionalTest.sbom is created when only the augmented sbom parameter is used (but not the '--enable-sbom' buildArg)(SBOMFunctionalTest.groovy:107)
```

**To Reproduce**

- download latest ea build from: https://github.com/graalvm/oracle-graalvm-ea-builds/releases/tag/jdk-25.0.0-ea.04
- set `GRAALVM_HOME` env variable to point to the downloaded build 
- set `JAVA_HOME` env variable to GraalVM 17
- from the project root, run: `./gradlew :native-maven-plugin:functionalTest` 

**Expected behavior**

The test should pass in the same way it passes with previous GraalVM releases.

**Logs**

Described failure occurred in the github actions when tried to enable maven functional tests with the latest ea build: https://github.com/graalvm/native-build-tools/actions/runs/12689070434/job/35366970949?pr=663#step:5:10653
 

**System Info (please complete the following information):**
 - OS: [`Linux`]
 - GraalVM Version [[latest ea build](https://github.com/graalvm/oracle-graalvm-ea-builds/releases/tag/jdk-25.0.0-ea.04)]
 - Plugin version [0.10.5-SNAPSHOT (current master)]

dnestoro:
@rudsberg can you please take a look on this failure since [you introduced this test](https://github.com/graalvm/native-build-tools/pull/623/files#diff-b95eb0f8538a4247d14c37bce95d138af32c444fc08d8f8ad8140c15693cd63dR95). Maybe some version check (maybe [this one](https://github.com/graalvm/native-build-tools/pull/623/files#diff-b95eb0f8538a4247d14c37bce95d138af32c444fc08d8f8ad8140c15693cd63dR60)) doesn't work well with the ea build version string (that is just my guess without deeper investigation)? Or maybe we really have some bug here.

PR #673, opened 2025-01-13
Fix SBOM integration test bug 

rudsberg:
This PR resolves a test failure identified by @dnestoro (see [here](https://github.com/graalvm/native-build-tools/issues/671#issuecomment-2585963620)). The issue was that the test incorrectly checked for an exported SBOM when only an embedded SBOM is generated. I believe the test previously did not fail since the exported SBOM was picked up from another test, but in the failed run that was not the case.

### DIFF
--- a/native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
+++ b/native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
@@ -69,7 +69,7 @@ class SBOMFunctionalTest extends AbstractGraalVMMavenFunctionalTest {
     }
 
     @Requires({ supportedAugmentedSBOMVersion() })
-    def "sbom is created when buildArg '--enable-sbom=export,embed' is used"() {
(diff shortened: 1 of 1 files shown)


native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
```diff
@@ -82,29 +82,26 @@ class SBOMFunctionalTest extends AbstractGraalVMMavenFunctionalTest {
...
         when:
         mvn '-Pnative-augmentedSBOM-only', '-DquickBuild', '-DskipTests', 'package', 'exec:exec@native'
 
-        def sbom = file("target/example-app.sbom.json")
-
         then:
         buildSucceeded
         outputContainsPattern".*CycloneDX SBOM with \\d+ component\\(s\\) is embedded in binary \\(.*?\\)."
         outputDoesNotContain "Use '--enable-sbom' to assemble a Software Bill of Materials (SBOM)"
-        validateSbom sbom
```
rudsberg, line 107: This was the culprit. The `sbom` JSON should _not_ be created in this scenario since `--enable-sbom` gets used which only embeds the SBOM in the image.

melix (approved):

PR #682, opened 2025-01-22
Fix workflow trigger

### LINKED ISSUE #670: [QUESTION] Pre-release/ snapshot repository unavailable?
Hi,

I'm trying to verify one of the latest bugfixes that is not yet part of 0.10.4. https://graalvm.github.io/native-build-tools/latest/maven-plugin.html#configuration-registering-plugin describes that SNAPSHOT builds are available via https://raw.githubusercontent.com/graalvm/native-build-tools/snapshots. However I'm unable to fetch 0.10.5-SNAPSHOT via Maven. Not sure if directory listing is enabled but when I try via browser I get `400: Invalid request`.

melix:
Fixes #670

@dnestoro I think the issue comes from the fact that we should use the names of the workflows, not their ids, in the list.

### DIFF
--- a/.github/workflows/deploy-snapshots.yml
+++ b/.github/workflows/deploy-snapshots.yml
@@ -2,7 +2,7 @@ name: "Deploy snapshots"
 
 on:
   workflow_run:
-    workflows: [ "deploy-documentation", "test-native-gradle-plugin", "test-native-maven-plugin", "test-junit-platform-native" ]
+    workflows: [ "Deploy documentation to website", "Test native-gradle-plugin", "Test native-maven-plugin", "Test junit-platform-native" ]
     branches: [ "master" ]
     types:
       - completed


dnestoro:
@melix my question is why we trigger this workflow like this? Why we don't trigger it when something is merged with master? We do the same in the metadata repository, [here](https://github.com/oracle/graalvm-reachability-metadata/blob/master/.github/workflows/test-all-metadata.yml#L6)

melix:
I think it's done so that we only publish snapshots for successful pipeline. This would be to avoid that a snapshot from a broken master is published. But that's only a guess.

dnestoro:
But we are performing checks before anything gets merged into master, right? So if we pass those checks, the master should be in the correct state? Or I am missing some edge case?

melix:
I don't think this was the case when this workflow was created.

dnestoro:
Aha, okay. So should we keep this approach with `workflows` or we want to switch to the strategy we use in metadata repo? What do you think is the better solution in this case? I am fine with both

dnestoro (approved):

PR #683, opened 2025-01-22
Add options to allow overriding some native-image parameters from com…

### LINKED ISSUE #475: Provide support to pass build args from the command line in the Gradle plugin
Add possibility to provide `buildArgs` option from the command line.

Open questions:
- Should command line argument override existing `buildArgs` from the config file, or it should append command line arguments to the config file arguments?
- What should we do in case we want to modify certain subset of arguments (provided in the config file) from the command line?

melix:
…mand line

Fixes #475

### DIFF
--- a/docs/src/docs/asciidoc/changelog.adoc
+++ b/docs/src/docs/asciidoc/changelog.adoc
@@ -1,6 +1,12 @@
 [[changelog]]
 == Changelog
 
+== Release 0.10.6
+
+=== Gradle plugin
+
+- Added options to override parameters from command line
+
 == Release 0.10.5
 
 - Add missing getters to `DirectoryConfiguration`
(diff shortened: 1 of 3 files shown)


melix:
Not sure if it's worth the effort, that would significantly increase build times for a feature which should be tested in Gradle itself.

dnestoro:
Also (just for my better understanding) how do we know that this will override existing option? Is there a defined order what option (from build files or command line) applies first? Like, are we sure that we will always load options from a build file first, and only then the options from the command line (so that the override is always achieved)?

dnestoro (approved):

melix:
We know because the convention values of the tasks are set from the extension, and that the `overrideXXX` methods are using the `set` method, so it's guaranteed to override. If it wasn't the case that would be a bug.

PR #684, opened 2025-01-23
Add exclusions to Maven plugin

### LINKED ISSUE #612: Exclude dependencies from native-image
Maven has a limited list of dependency scopes and it's not possible to create custom ones. As such, dependencies that are development only can end up on the classpath. This has caused issues in the past for things that need to build an executable based on a Maven project.

Spring Boot has had several issues such as https://github.com/spring-projects/spring-boot/issues/13289 and implemented several features to allow dependencies to be excluded:

* [Exclude by GAV](https://docs.spring.io/spring-boot/maven-plugin/packaging.html#packaging.examples.exclude-dependency)
* [Explicit flag for well-known dependencies](https://docs.spring.io/spring-boot/maven-plugin/packaging.html#packaging.repackage-goal.parameter-details.exclude-devtools)
* Add support for [an attribute in the MANIFEST](https://github.com/spring-projects/spring-boot/issues/22036) that would remove the jar

The plugin currently works in two modes: it builds the classpath itself, or it lets you define it. It would be nice if we could configure it to be able to exclude certain dependencies that shouldn't be analyzer by native-image in the first place.

A way to exclude a dependency. Either by providing the GAV in the plugin configuration or via a manifest entry. The latter looks less invasive and would work out of the box.

None as building the classpath ourselves isn't really practical with Maven. The plugin is configured in our maven parent and used in a wide range of contexts.

melix:
This commit adds an `exclusions` parameter to the Maven plugin, which can be used to exclude dependencies from the compile classpath of native image.

It can be used to remove some dependencies from compilation, such as these injected by Spring Boot for development. Note, however, that if the main jar is configured to have the manifest with `addClaspath` to true, then the manifest file will still reference the jar, and native compile will issue a warning (because the jar will be missing from classpath).

Fixes #612

### DIFF
--- a/docs/src/docs/asciidoc/changelog.adoc
+++ b/docs/src/docs/asciidoc/changelog.adoc
@@ -7,6 +7,10 @@
 
 - Added options to override parameters from command line
 
+=== Maven plugin
+
+- Add support for dependency exclusions
+
 == Release 0.10.5
 
 - Add missing getters to `DirectoryConfiguration`
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -137,6 +137,16 @@ The following configuration options are available:
     <param>path/to/classes</param>
 </classpath>
 ----
+`<exclusions>`::
+   This can be used to exclude artifacts from native-image compilation.
+----
+<exclusions>
+    <exclusion>
+       <groupId>org.example</groupId>
+       <artifactId>example</artifactId>
+    </exclusion>
+</exclusions>
+----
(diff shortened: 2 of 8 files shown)


alvarosanchez (approved):

alvarosanchez:
This needs to be documented somewhere, though. Unfortunately, Maven plugin site is not published

dnestoro (dismissed):

melix:
@dnestoro the workflow changes are probably required on all PRs, it just happened to fail suddenly yesterday, because GH actions changed behavior wrt the deprecated v3 version of upload-artifact.

dnestoro (approved):

PR #688, opened 2025-01-31
SBOM: fix issue with unresolved artifacts and empty "components" 

rudsberg:
This PR fixes two issues:
* If an artifact cannot be resolved via the [RepositorySystem](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/RepositorySystem.html) it can throw an `ArtifactResolutionException`, causing the SBOM generation to fail. This can happen for an artifact that cannot be located locally or on maven central. An example of such an artifact: `com.example:demo:jar:sources:0.0.1-SNAPSHOT`. The fix is catch the `ArtifactResolutionException` and return `Optional.empty()` from `resolvePackageNamesFromArtifact`. Such components will not be pruned by Native Image and will be included under `components`.
* The SBOM for a simple app with no dependencies will only contain the application component and it will be listed under `metadata/component` and the `components` list will be empty. Previously we incorrectly threw an exception for such cases. The fix is to simply return from the `augmentSBOM` method instead of throwing the exception.


rudsberg:
The CI failures are unrelated to this diff and will be resolved in [this PR](https://github.com/graalvm/native-build-tools/pull/684).

dnestoro (approved):

Issue #638, opened 2024-11-05
Add support for `@FieldSource` for `@ParameterizedTest` methods


sbrannen:
## Overview

[`@FieldSource`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-FieldSource) was introduced in JUnit Jupiter 5.11 for use with `@ParameterizedTest` methods.

Similar to `@MethodSource` which allows developers to supply a fully-qualified method name, `@FieldSource` allows developers to supply a fully-qualified field name.

Native Build Tools already registers reflection metadata for fully-qualified method names for `@MethodSource` here:

https://github.com/graalvm/native-build-tools/blob/979e9ecfcc8d2c4e35808022b3c6724212bc4417/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java#L144-L166

Similar support should be added for `@FieldSource`. For example, given the following parameterized test, reflection metadata should be registered for the `example.FruitUtils#tropicalFruits` field.

```java
@ParameterizedTest
@FieldSource("example.FruitUtils#tropicalFruits")
void testWithExternalFieldSource(String tropicalFruit) {
    // test with tropicalFruit
}
```

## Related Issues

- #51 
- #54 
- #636

dnestoro:
Hey @sbrannen thanks for rising this issue. As part of [JUnit support refactoring](https://github.com/graalvm/native-build-tools/pull/693), I've added support for `@FieldSource` as well

dnestoro:
Implemented here: https://github.com/graalvm/native-build-tools/pull/693/files#diff-7b2f9d1eea335da0c18259129c76170dfd6e022048a993d867e121cfe1a23eb7R89

Issue #636, opened 2024-11-01
Interfaces implemented by test classes are not registered for reflection


sbrannen:
Given the following interface:

```java
package org.example;

import java.util.List;

interface TestInterface {

	static List<String> names() {
		return List.of("Sarah", "Susan");
	}
}
```

... and the following test class:

```java
package org.example;

import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;

import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.MethodSource;

class DemoTests implements TestInterface {

	@ParameterizedTest
	@MethodSource("names")
	void test(String name) {
		assertEquals(5, name.length());
		assertTrue(name.startsWith("S"));
	}
}
```

... `DemoTests` passes on the JVM but fails within a native image as follows.

```
Failures (1):
  JUnit Jupiter:DemoTests:test(String)
    MethodSource [className = 'org.example.DemoTests', methodName = 'test', methodParameterTypes = 'java.lang.String']
    => org.junit.platform.commons.PreconditionViolationException: Could not find factory method [names] in class [org.example.DemoTests]
```

However, if `TestInterface` were a class that `DemoTests` extended the test would then pass.

The reason is that the `JUnitPlatformFeature` currently invokes `registerAllClassMembersForReflection()` for all classes with the test class hierarchy, but that recursive algorithm ignores implemented interfaces.

https://github.com/graalvm/native-build-tools/blob/979e9ecfcc8d2c4e35808022b3c6724212bc4417/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java#L150-L158

The `registerTestClassForReflection()` method in `JUnitPlatformFeature` should therefore be revised to process implemented interfaces (and super-interfaces) at each level of the class hierarchy.

vjovanov:
Thanks for reporting, this is on our short-term plan and we will look into it. `MethodSource` seems to be computed at run time and needs extra metadata.

Our future implementation of the JUnit feature will have to register this metadata based on the test annotations.

sbrannen:
Hi @vjovanov,

> `MethodSource` seems to be computed at run time and needs extra metadata.

I apologize: my use of `@MethodSource` to demonstrate the larger issue was perhaps a bit misleading.

NBT already has specific support for registering reflection metadata for fully-qualified method names configured via `@MethodSource`, which can be seen here:

[...]

PR #693, opened 2025-02-12
[GR-60258] Refactor JUnit feature

dnestoro:
Problem:
- Currently, we are running test discovery and generate test plan at buildtime.
- Then, we are storing launcher and test plan in the ImageSingleton, in order to run it at runtime.
- This approach requires many classes to be initialized at buildtime.
- The negative effect of this approach is that we have to maintain multiple lists of classes that should be initialized at buildtime ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7b2f9d1eea335da0c18259129c76170dfd6e022048a993d867e121cfe1a23eb7L71), [here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-c30f9fd7ac074fcffe02c542389081f6f3032c3ed93b30a87daa13d217086c72L51), and [here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-723561f3c316f3bf30520ed786281efd1a3ed147e3b06713b90759642b5d6770L51)).

Idea:
- Move test plan generation and tests execution to runtime.

Solution:
- _PREREQUISITE_: We are using `test-ids` generated before test compilation to discover tests (currently on master as well)
- We should run test discovery at buildtime to only collect classes needed for reflection registration
- Then at runtime, we will have everything necessary for reflection, and the only thing left is to run test discovery with `test-ids` to find tests and execute them
- `Test-ids` location should be passed to the NativeImageJUnitLauncher class, so we simply set `systemProperty` with this location in order to read it at runtime (we do this automatically for both [gradle](https://github.com/graalvm/native-build-tools/pull/693/files#diff-3ca19e7e49fe64feca3a1e88279c961b56e03450666583025c711da44cfb73bbR664) and [maven](https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java#L163))

This PR:
- removes hardcoded lists for `initializeAtBuildtime` ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7b2f9d1eea335da0c18259129c76170dfd6e022048a993d867e121cfe1a23eb7L71), [here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-c30f9fd7ac074fcffe02c542389081f6f3032c3ed93b30a87daa13d217086c72L51), and [here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-723561f3c316f3bf30520ed786281efd1a3ed147e3b06713b90759642b5d6770L51))
- removes `launcher` and `testPlan` from ImageSingletons ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0L95)) => therefore we don't need classes to be initialized at buildtime 
- [computes `testPlan` at runtime](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7c5f329e3e5f35f935274f283d925007ff37ed30f5a2bb2d3e7f4bf84304aa1eR144)

In addition:
- [registers interfaces for reflection](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R161) to solve https://github.com/graalvm/native-build-tools/issues/636
- [adds support for `@FieldSource` annotations](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7b2f9d1eea335da0c18259129c76170dfd6e022048a993d867e121cfe1a23eb7R98) to solve https://github.com/graalvm/native-build-tools/issues/638
- [Adds support for `org.junit.rules.ExpectedException`](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R197) (code originally taken from [this PR](https://github.com/graalvm/native-build-tools/pull/576) cc @tzezula)
- [Fixes JUnit tests](https://github.com/graalvm/native-build-tools/pull/693/files#diff-6e74bcea7cf033a528fecfaccafcf8a79e5faa47c83209591f457ced1d447f6e) that fail when updating JUnitJupiter from `5.10.0` to `5.11.0` (see [this](https://junit.org/junit5/docs/5.11.0/release-notes/#release-notes-5.11.0-junit-platform-deprecations-and-breaking-changes))
- Adds functional tests for both [gradle](https://github.com/graalvm/native-build-tools/pull/693/files#diff-2d38597074d1371576faa4a9afce149185252158ed0532cc4c2bad7ee0c4afc1) and [maven](https://github.com/graalvm/native-build-tools/pull/693/files#diff-792b73050a524e87925c106be5bdd6405a5ec0c20c0b33fe35df2381cfc021a6)


dnestoro:
> Thanks David. Can you give a description of the changes that this makes? For example you mention that the tests are discovered at runtime but in practice you are still using the test-ids.
> 
> There are changes which need to be reverted on the samples: build plugins versions and maven repositories.

Hey @melix the PR is not ready yet (it is in draft state). I will update PR description and revert unnecessary changes after I finish testing.

olpaw:
Please clean up the commit history: https://github.com/graalvm/native-build-tools/pull/693/commits

E.g. fold
`[Fix checkstyle errors]` commits into the commit that introduced the style error.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -74,66 +75,73 @@
 public final class JUnitPlatformFeature implements Feature {
```
olpaw, line 75: If we migrate this class and it's related classes (e.g. `NativeImageJUnitLauncher`, ...) to e.g. `com.oracle.svm.junit5` in e.g.
`graal/substratevm/src/com.oracle.svm.junit5/src/com/oracle/svm/junit5` in
https://github.com/oracle/graal
we would be able to support building junit5 tests into images even without `native-build-tools`.

This would be very handy to eventually replace `mx native-unittest` with `mx native-unittest5` in the future and to only have one remaining junit feature that is used for all situations.

cc @zapster 

(obviously this would happen via follow up PRs)

gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ mavenResolver = "1.9.22"
 graalvm = "23.0.2"
 openjson = "1.0.13"
 junitPlatform = "1.10.0"
-junitJupiter = "5.10.0"
+junitJupiter = "5.11.0"
```
dnestoro, line 15: I updated this in order to provide support for `@FieldSource` annotation introduced in `5.11.0` (see [this](https://junit.org/junit5/docs/5.11.0/release-notes/#release-notes-5.11.0))

sbrannen: Please note that we recently released JUnit 5.12.

https://github.com/junit-team/junit5/releases/tag/r5.12.0

So, it's probably a good idea to upgrade to that. I can create a separate GitHub issue for that if you like.

sbrannen: As a side note, JUnit 5.12 includes `native-image.properties` files.

> All affected JAR files now include `native-image.properties` files that contain the `--initialize-at-build-time` option to avoid breakages in GraalVM projects when updating to newer versions of JUnit.

So, that might affect your refactoring effort.

dnestoro: So, if I get it right, we can just use that as a default for GraalVM versions with previous initialization strategy, and `--execlude-config` for newer GraalVM versions, right?

fniephaus:
While not so concerning for testing, how much bigger do such test images get when launcher and testPlan are no longer ImageSingletons? Bigger or smaller? :)

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JUnitFunctionalTests.groovy
```diff
@@ -0,0 +1,85 @@
...
+import spock.lang.Ignore
+import spock.lang.IgnoreIf
+import spock.lang.Issue
+import spock.lang.Requires
+import spock.lang.Unroll
+
+import java.nio.file.Files
+
+class JUnitFunctionalTests extends AbstractFunctionalTest {
+    @Unroll("test if JUint support works with various annotations, reflection and resources")
```
melix, line 54: :nit: this is not an unrolled test. I guess you just wanted a display name, in which case you can put this directly in the test method name (also, there's a typo in JUnit)

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -662,6 +661,7 @@ public void registerTestBinary(Project project,
             // Later this will be replaced by a dedicated task not requiring execution of tests
             testList.from(testListDirectory).builtBy(testTask);
             testOptions.getClasspath().from(testList);
+            testOptions.getRuntimeArgs().add("-D" + JUNIT_PLATFORM_LISTENERS_UID_TRACKING_OUTPUT_DIR + "=" + testResultsDir.dir(testTask.getName() + "/testlist").get().getAsFile().getAbsolutePath());
```
melix, line 664:
```suggestion
            testOptions.getRuntimeArgs().add(testResultsDir.dir(testTask.getName() + "/testlist").map(d -> "-D" + JUNIT_PLATFORM_LISTENERS_UID_TRACKING_OUTPUT_DIR + "=" + d.getAsFile().getAbsolutePath()));

```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
```diff
@@ -137,4 +140,106 @@ public static void main(String... args) {
...
+            System.out.println("[junit-platform-native] WARNING: Using test-ids from default Maven project location:" + defaultMavenTestIDsLocation);
+            return defaultMavenTestIDsLocation;
+        }
+
+        throw new RuntimeException("[junit-platform-native] test-ids not provided to the NativeImageJUnitLauncher and cannot be found on default locations. " +
+                  "Searched in: " + defaultGradleTestIDsLocation + " and " + defaultMavenTestIDsLocation);
+    }
+
+    private static Path getGradleTestIdsDefaultLocation() {
+        return Path.of(getBuildDirectory(File.separator + "build" + File.separator))
```
melix, line 222: :x: This seems extremely fragile. This would fail as soon as the build directory is changed, or if the test suite is not the main `test` one. Can you clarify if this is a fallback or if this is always used. In the latter case this must be fixed or basically this will break execution of native integration tests for example (test suite name = `integTest`).

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
melix: Why is it done here and not in `TrackingDirectorySystemPropertyProvider` anymore? Actually I don't see this class changed either, is this redundant?

melix (changes requested):
I didn't finish the review but left a few comments. I'm still confused and would appreciate a few more details:

- you are saying that we should discover the test ids at runtime, however, we still run the tests in JVM mode to get the test ids?
- if the test ids are not needed anymore, maybe we should change the name of the parameters which are used

gradle/libs.versions.toml
dnestoro: Hey @sbrannen, I tried `JUnit 5.12` with both my implementation and the current master. My findings are the following:

- If I bump JUnit version from `5.10` to `5.12` image build command contains multiple lines like: 
```
-H:ClassInitialization@jar:.../.gradle/caches/modules-2/files-2.1/org.junit.platform/junit-platform-reporting/.../META-INF/native-image/org.junit.platform/junit-platform-reporting/native-image.properties+api=org.junit.platform.reporting.open.xml.OpenTestReportGeneratingListener...
``` 
- So `native-image.properties` are really provided to the native image
- But if I remove our own `initializeAtBuildtime` (either on [master](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/platform/PlatformConfigProvider.java#L51) or in [my implementation](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R100)) I am facing the following error: 
```
Fatal error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: An object of type 'org.junit.platform.commons.support.scanning.DefaultClasspathScanner' was found in the image heap. This type, however, is marked for initialization at image run time for the following reason: classes are initialized at run time by default.
This is not allowed for correctness reasons: All objects that are stored in the image heap must be initialized at build time.
...
```
- So, it looks like properties files provided by JUnit are not complete

dnestoro:
> While not so concerning for testing, how much bigger do such test images get when launcher and testPlan are no longer ImageSingletons? Bigger or smaller? :)

On a first look, the tests image size gets around 5% of increase. But I will run more tests and get provide more accurate information soon

dnestoro:
@melix 
> - you are saying that we should discover the test ids at runtime, however, we still run the tests in JVM mode to get the test ids?
> - if the test ids are not needed anymore, maybe we should change the name of the parameters which are used

No, I am just saying that we should use test ids to compute test plan (to discover them with `launcher.discover(request)`) at runtime. With that approach, we don't have to [compute testPlan at buildtime](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java#L128) and [store it in ImageSingletons](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java#L95) in order to [use it at runtime](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java#L74) (as we currently do on master)

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
dnestoro: Let me check if I can move it into `TrackingDirectorySystemPropertyProvider`. 

> is this redundant?

I would say it is not redundant. Without this line, we are not able to read the value of `UniqueIdTrackingListener.OUTPUT_DIR_PROPERTY_NAME` at runtime ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7c5f329e3e5f35f935274f283d925007ff37ed30f5a2bb2d3e7f4bf84304aa1eR155))

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
dnestoro: Yes, this should only be used in the fallback mode when we can't find the location of test ids directory from the system property ([here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7c5f329e3e5f35f935274f283d925007ff37ed30f5a2bb2d3e7f4bf84304aa1eR156)). 

If I am not wrong, this can only happen in a situation where you:
- compiled tests with `gradle nativeTestCompile`
- then tried to execute tests executable from command line (not by using `gradle nativeTest`) like `./my-tests-executable`

because we are setting the value of the system property [when configuring the test task](https://github.com/graalvm/native-build-tools/pull/693/files#diff-3ca19e7e49fe64feca3a1e88279c961b56e03450666583025c711da44cfb73bbR664). Therefore, if you use `gradle nativeTest` this code won't be executed as we would have test ids location set as systemProperty (runtime argument). Only if you run tests executable manually, test ids location won't be provided as systemProperty and we will try to find it on default locations 

Note: of course if you are running tests executable manually, you can also provide `-Djunit.platform.listeners.uid.tracking.output.dir=<pathToTestIds>` as a runtime argument, and again this code won't be executed

There is a warning describing that [here](https://github.com/graalvm/native-build-tools/pull/693/files#diff-7c5f329e3e5f35f935274f283d925007ff37ed30f5a2bb2d3e7f4bf84304aa1eR197)

dnestoro: But I agree that we might strengthen this calculation in order to cover more use-cases. One of my ideas was to walk through whole `build`/`target` directory searching for `testlists`/`test-ids` folder (but that might take some time, and I gave up on that idea. But we can try that as well)

dnestoro:
@fniephaus after deeper investigation, I came to the following results regarding image size:

- On an image created from a single test:
  - master: 25.47MB
  - branch: 27.29MB
  - diff: 1.82MB (6.6691% bigger image size)


- On an image created from 400 tests:
  - master: 26.58MB
  - branch: 27.88MB
  - diff: 1.3MB (4,6628% bigger image size)

- On an image created from 1500 tests:
  - master: 28.96MB
  - branch: 29.31MB
  - diff: 0.35MB (1.1941% bigger image size)

- On an image created from 2800 tests, the master and the branch are equal

- On an image, created from 10 000 tests:
  - master: 48.81MB
  - branch: 42.39MB
  - diff: -6.42MB (13.15% smaller image size)

The more tests, the more benefits we get!

fniephaus:
Thanks for investigating, @dnestoro!

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -74,66 +75,73 @@
...
+        registerClassesForHamcrestSupport(access);
     }
 
-    private List<? extends DiscoverySelector> getSelectors(List<Path> classpathRoots) {
+    private List<? extends DiscoverySelector> getSelectors() {
         try {
-            Path outputDir = Paths.get(System.getProperty(UniqueIdTrackingListener.OUTPUT_DIR_PROPERTY_NAME));
-            String prefix = System.getProperty(UniqueIdTrackingListener.OUTPUT_FILE_PREFIX_PROPERTY_NAME,
+            Path uniqueIdDirectory = Path.of(System.getProperty(UniqueIdTrackingListener.OUTPUT_DIR_PROPERTY_NAME));
+            String uniqueIdFilePrefix = System.getProperty(UniqueIdTrackingListener.OUTPUT_FILE_PREFIX_PROPERTY_NAME,
```
olpaw, line 115: Provide error messages if system properties OUTPUT_DIR_PROPERTY_NAME or OUTPUT_FILE_PREFIX_PROPERTY_NAME are missing (null)

olpaw (changes requested):
Looks quite good overall and almost ready to merge.
Only a few comments (see above).

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
```diff
@@ -137,4 +140,131 @@ public static void main(String... args) {
...
+        if (!Files.exists(dir)) {
+            return Stream.empty();
+        }
+        return Files.find(dir, Integer.MAX_VALUE,
+                (path, basicFileAttributes) -> (basicFileAttributes.isRegularFile()
+                        && path.getFileName().toString().startsWith(prefix)));
+    }
+
+    private static Path getTestIDsFromDefaultLocations() {
+        System.out.println("[junit-platform-native] WARNING: Trying to find test-ids on default locations. " +
```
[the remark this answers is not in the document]
dnestoro: This is required in case when someone compiles tests executable with `nativeTestCompile` and runs it from the command line (and not with `nativeTest`) as we do in some of our functional tests (for example [here](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy#L72))

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -192,4 +202,26 @@ private static Stream<Path> findFiles(Path dir, String prefix) throws IOExceptio
                     && path.getFileName().toString().startsWith(prefix)));
     }
 
+    private static void registerClassesForHamcrestSupport(BeforeAnalysisAccess access) {
```
dnestoro, line 205: Implementation originally taken from: https://github.com/graalvm/native-build-tools/pull/576

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
```diff
@@ -137,4 +140,135 @@ public static void main(String... args) {
...
+            throw new RuntimeException("Failed to read UIDs from UniqueIdTrackingListener output files: " + ex.getMessage());
+        }
+
+        throw new RuntimeException("Cannot compute test selectors from test ids.");
+    }
+
+    private static Stream<String> readAllFiles(Path dir, String prefix) throws IOException {
+        return findFiles(dir, prefix).map(outputFile -> {
+            try {
+                return Files.readAllLines(outputFile);
```
olpaw, line 184:
```suggestion
                return Files.readAllLines(outputFile).stream();
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
```diff
@@ -137,4 +140,135 @@ public static void main(String... args) {
...
+            }
+        } catch (Exception ex) {
+            throw new RuntimeException("Failed to read UIDs from UniqueIdTrackingListener output files: " + ex.getMessage());
+        }
+
+        throw new RuntimeException("Cannot compute test selectors from test ids.");
+    }
+
+    private static Stream<String> readAllFiles(Path dir, String prefix) throws IOException {
+        return findFiles(dir, prefix).map(outputFile -> {
```
olpaw, line 182:
```suggestion
        return findFiles(dir, prefix).flatmap(outputFile -> {
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/NativeImageJUnitLauncher.java
```diff
@@ -137,4 +140,135 @@ public static void main(String... args) {
...
+            try {
+                return Files.readAllLines(outputFile);
+            } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
+            }
+        }).flatMap(List::stream);
```
olpaw, line 188:
```suggestion
        });
```

olpaw (approved):

dnestoro:
@sbrannen @melix @fniephaus @vjovanov I planned to merge this PR by the end of the week, so please take a look on it if you have time this week.

tzezula:
We are using `junit-platform-native` to run Truffle unit tests in native-image.
I've tried to update to `junit-platform-native` built from your branch and the truffle gate is failing.
The native image is created but I am getting following error while executing the resulting native-image.
Truffle does not use maven nor gradle to build and run the unit tests.

```
[junit-platform-native] WARNING: Trying to find test-ids on default locations. As this is a fallback mode, it could take a while. This should only happen if you are running tests executable manually.
Exception in thread "main" java.lang.RuntimeException: Failed to read UIDs from UniqueIdTrackingListener output files: [junit-platform-native] test-ids not provided to the NativeImageJUnitLauncher and cannot be found on default locations. Searched in: null and null
	at org.graalvm.junit.platform.NativeImageJUnitLauncher.getSelectors(NativeImageJUnitLauncher.java:175)
	at org.graalvm.junit.platform.NativeImageJUnitLauncher.getTestPlan(NativeImageJUnitLauncher.java:145)
	at org.graalvm.junit.platform.NativeImageJUnitLauncher.main(NativeImageJUnitLauncher.java:119)
	at java.base@25/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
```

tzezula:
Truffle is using `junit-platform-native` to run Truffle unit tests as a native-image. I've tried to use `junit-platform-native` from your branch to verify that it does not break Truffle tests. I am passing test_ids using `junit.platform.listeners.uid.tracking.output.dir` system property to the generated image. 

I am getting the followimg waring:
```
WARNING: Could not read unique ID for Description; using display name instead: somShapeAllocatedOnContextPreInit(com.oracle.truffle.api.test.polyglot.ContextPreInitializationNativeImageTest)
java.lang.NoSuchFieldException: fUniqueId
	at java.base@25/java.lang.Class.checkField(DynamicHub.java:1246)
	at java.base@25/java.lang.Class.getDeclaredField(DynamicHub.java:1388)
	at org.junit.platform.commons.util.ReflectionUtils.lambda$tryToReadFieldValue$2(ReflectionUtils.java:640)
	at org.junit.platform.commons.function.Try.lambda$call$0(Try.java:57)
	at org.junit.platform.commons.function.Try.of(Try.java:93)
	at org.junit.platform.commons.function.Try.call(Try.java:57)
	at org.junit.platform.commons.util.ReflectionUtils.tryToReadFieldValue(ReflectionUtils.java:640)
	at org.junit.vintage.engine.support.UniqueIdReader.apply(UniqueIdReader.java:47)
	at org.junit.vintage.engine.support.UniqueIdReader.apply(UniqueIdReader.java:28)
	at java.base@25/java.util.function.Function.lambda$andThen$0(Function.java:88)
	at java.base@25/java.util.stream.Collectors.lambda$groupingBy$0(Collectors.java:1111)
	at java.base@25/java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
	at java.base@25/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1716)
	at java.base@25/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
	at java.base@25/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
	at java.base@25/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base@25/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:265)
	at java.base@25/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:727)
	at org.junit.vintage.engine.discovery.RunnerTestDescriptorPostProcessor.addChildrenRecursively(RunnerTestDescriptorPostProcessor.java:51)
	at org.junit.vintage.engine.discovery.RunnerTestDescriptorPostProcessor.applyFiltersAndCreateDescendants(RunnerTestDescriptorPostProcessor.java:41)
	at org.junit.vintage.engine.discovery.VintageDiscoverer.discover(VintageDiscoverer.java:46)
	at org.junit.vintage.engine.VintageTestEngine.discover(VintageTestEngine.java:64)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverEngineRoot(EngineDiscoveryOrchestrator.java:152)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverSafely(EngineDiscoveryOrchestrator.java:132)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:107)
	at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:78)
	at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:99)
	at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:77)
	at org.junit.platform.launcher.core.DelegatingLauncher.discover(DelegatingLauncher.java:42)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.discover(SessionPerRequestLauncher.java:56)
	at org.graalvm.junit.platform.NativeImageJUnitLauncher.getTestPlan(NativeImageJUnitLauncher.java:150)
	at org.graalvm.junit.platform.NativeImageJUnitLauncher.main(NativeImageJUnitLauncher.java:119)
	at java.base@25/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)
```

Any idea how to resolve it?

tzezula:
There are also problems with non default runners that used to work before.
For example:
```
JUnit Vintage:ContextAPITest:initializationError
    ClassSource [className = 'com.oracle.truffle.api.test.polyglot.ContextAPITest', filePosition = null]
    => java.lang.Exception: Custom runner class Theories should have a public constructor with signature Theories(Class testClass)
       java.base@25/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215)
       java.base@25/java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1693)
       java.base@25/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:147)
       java.base@25/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:588)
       java.base@25/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:574)
       java.base@25/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
       java.base@25/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
       java.base@25/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:265)
       java.base@25/java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:687)
       java.base@25/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215)
       [...]
       ```

**Comment by tzezula** [comment]
The problem with `WARNING: Could not read unique ID for Description; using display name instead: ` is now resolved.

**Comment by tzezula** [comment]
Regarding missing reflection configurations that were not needed before.
Failure example:
```
JUnit Vintage:VarargsParseSignatureTest:initializationError
    ClassSource [className = 'com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest', filePosition = null]
    => org.junit.runners.model.InvalidTestClassError: Invalid test class 'com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest':
  1. Node com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseTwoFixedArgs should have a default constructor or a constructor taking a com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest
  2. Node com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseOneVararg should have a default constructor or a constructor taking a com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest
  3. Node com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseTwoVarargs should have a default constructor or a constructor taking a com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest
  4. Node com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseFixedArgs should have a default constructor or a constructor taking a com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest
  5. Node com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseNoFixedArgs should have a default constructor or a constructor taking a com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest
       com.oracle.truffle.tck.TruffleRunner.<init>(TruffleRunner.java:342)
       java.base@25/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
       java.base@25/java.lang.reflect.Constructor.newInstance(Constructor.java:483)
       java.base@25/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:215)
       java.base@25/java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1693)
       java.base@25/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:147)
       java.base@25/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:588)
       java.base@25/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:574)
       java.base@25/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
       java.base@25/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150)
```

For us it's easy to register classes for reflection for top level test classes like `com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest` but finding that `com.oracle.truffle.nfi.test.parser.VarargsParseSignatureTest$ParseTwoFixedArgs` needs to be registered as well is non-trivial.

**Review by olpaw — DISMISSED** [review]

**Review by melix — DISMISSED** [review]

**Review by tzezula — DISMISSED** [review]
Looks good to me.
Works fine for Truffle unit tests.

File: gradle/libs.versions.toml (hunk 2 above)
Comment (sbrannen) [reply gradle/libs.versions.toml:15]: @marcphilipp, please take a look. Thanks!

Comment (marcphilipp) [reply gradle/libs.versions.toml:15]: I doubt that's the only class that's missing. I could copy the list of classes from the hard-coded lists in the feature implementation but I'd rather extend our existing tests so it exercises all relevant code.

> But if I remove our own initializeAtBuildtime (either on [master](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/platform/PlatformConfigProvider.java?rgh-link-date=2025-02-26T15%3A43%3A34Z#L51) or in [my implementation](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R100)) I am facing the following error

@dnestoro What do I need run to reproduce this locally?

Comment (dnestoro) [reply gradle/libs.versions.toml:15]: I am so confused now... I tried to reproduce this error again, but now everything works correctly on master (while on my branch it fails with different errors depending on the GraalVM version I use). I will double check what happened here (maybe something got cached) because I am 100% sure that it was failing on master as well.

@marcphilipp @sbrannen Sorry for the false alarm. I will investigate this deeper to see if something is still missing on master and give you a reproducer if something fails.

File: common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java [hunk 12]
```diff
@@ -221,4 +231,15 @@ private static Class<?> findClassOrNull(ClassLoader loader, String className) {
             return null;
         }
     }
+
+    private static void initializeClassesForJDKsBefore21() {
+        try (InputStream is = JUnitPlatformFeature.class.getResourceAsStream("/initialize-at-buildtime")) {
+            if (is != null) {
+                BufferedReader br = new BufferedReader(new InputStreamReader(is));
+                br.lines().forEach(RuntimeClassInitialization::initializeAtBuildTime);
```
olpaw, line 239: BufferedReader is not closed.

olpaw (changes requested):

common/junit-platform-native/src/main/resources/initialize-at-buildtime
```diff
@@ -0,0 +1,55 @@
+org.junit.jupiter.api.condition.OS
```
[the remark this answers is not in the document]
dnestoro: Actually, I don't know if this is really a **minimal** set, but I have run all junit platform tests we have (with `./gradlew :junit-platform-native:nativeTest`) and collected classes that were reported as problematic during the image build. Maybe removing some of the classes would still keep those tests passing, but finding which class is sufficient could only be done by guessing or with some fuzzing (which is most likely an overkill in this case since I would expect very few removals).

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
olpaw: If an exception happens between
```
BufferedReader br = ...
```
and
```
br.close();
```
close will not be called.

Use
```
try (BufferedReader br = ...) { ... }
```

common/junit-platform-native/src/main/resources/initialize-at-buildtime
```diff
@@ -0,0 +1,119 @@
+java.beans.Introspector
```
olpaw, line 1: There is
```
rci.initializeAtBuildTime("java.beans.Introspector", "JMX support");
```
in `substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/jdk/JmxCommonFeature.java`
Instead of adding `java.beans.Introspector` here, we should unconditionally set it to initializeAtBuildTime in the image builder (in `com.oracle.svm.hosted.jdk.JDKInitializationFeature#afterRegistration`).

olpaw (changes requested):

common/junit-platform-native/src/main/resources/initialize-at-buildtime
olpaw: cc @christianhaeubl

common/junit-platform-native/gradle/native-image-testing.gradle
```diff
@@ -90,6 +91,17 @@ abstract class NativeTestArgumentProvider implements CommandLineArgumentProvider
     @Input
     abstract Property<Boolean> getDiscovery()
 
+    @InputDirectory
+    abstract DirectoryProperty getExcludeConfigDir()
+
+    List<String> addExcludeConfig() {
```
marcphilipp, line 97: Shouldn't this method be `private`?

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -764,14 +765,20 @@ private static NativeImageOptions createTestOptions(GraalVMExtension graalExtens
...
         runtimeArgs.add("--xml-output-dir");
         runtimeArgs.add(project.getLayout().getBuildDirectory().dir("test-results/" + binaryName + "-native").map(d -> d.getAsFile().getAbsolutePath()));
+
         testExtension.buildArgs("--features=org.graalvm.junit.platform.JUnitPlatformFeature");
         ConfigurableFileCollection classpath = testExtension.getClasspath();
         classpath.from(configs.getImageClasspathConfiguration());
         classpath.from(sourceSet.getOutput().getClassesDirs());
         classpath.from(sourceSet.getOutput().getResourcesDir());
+
+        /* in version 5.12.0 JUnit added initialize-at-builditime properties files which we need to exclude */
```
marcphilipp, line 779:
```suggestion
        /* in version 5.12.0 JUnit added initialize-at-build-time properties files which we need to exclude */
```

common/junit-platform-native/gradle/native-image-testing.gradle
```diff
@@ -90,6 +91,17 @@ abstract class NativeTestArgumentProvider implements CommandLineArgumentProvider
...
+        String[] jarPatternPair = getExcludeConfigDir().get().file("exclude-config").getAsFile().readLines().get(0).split(",")
```
marcphilipp, line 98: Why is this loaded from an external file rather than being hard-coded? AFAICS this script is only applied once.

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -74,66 +78,85 @@
...
+            System.out.printf("[Debug] " + format + "%n", args);
+        }
+    }
+
+    @Override
+    public void afterRegistration(AfterRegistrationAccess access) {
+        extensionConfigProviders.forEach(p -> p.initialize(access.getApplicationClassLoader(), nativeImageConfigImpl));
+    }
+
+    public static boolean debug() {
```
marcphilipp, line 95: Shouldn't this method be `private`?

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -143,14 +166,18 @@ private TestPlan discoverTestsAndRegisterTestClassesForReflection(Launcher launc
                 .map(ClassSource.class::cast)
                 .map(ClassSource::getJavaClass)
                 .forEach(this::registerTestClassForReflection);
-
-        return testPlan;
     }
 
     private void registerTestClassForReflection(Class<?> clazz) {
         debug("Registering test class for reflection: %s", clazz.getName());
         nativeImageConfigImpl.registerAllClassMembersForReflection(clazz);
```
marcphilipp, line 173: I think interfaces and superclasses of member classes also need to be registered for reflection to support scenarios such as this:

```java
abstract BaseTests {
    @Test test() {}
}

@ParameterizedClass
class OuterTests {
	@Nested class InnerTests extends BaseTests {}
}
```

common/junit-platform-native/gradle/native-image-testing.gradle
dnestoro: So there is a case where users may need to add `exclude-config` manually (if they are using JUnit `5.12` and they only have dependency to `junit-platform-native` and not to `Gradle` or `Maven`). 
To help them add this exclude-config manually, I stored all necessary information in the external file, which they can just parse and use

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
dnestoro: I extracted [this code](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R176) from `registerAllClassMembersForReflection` function. That way we are registering all class members + interfaces + superclass from inner classes.

@marcphilipp Could you please add the above test [here](https://github.com/graalvm/native-build-tools/tree/master/common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter) once you release next JUnit version that supports `@ParameterizedClass`?

marcphilipp: > I extracted [this code](https://github.com/graalvm/native-build-tools/pull/693/files#diff-88ba6393b3a8379912730935514a00f87cbf552725c49effd00eeb8b3f17cbd0R176) from `registerAllClassMembersForReflection` function. That way we are registering all class members + interfaces + superclass from inner classes.

LGTM

> Could you please add the above test [here](https://github.com/graalvm/native-build-tools/tree/master/common/junit-platform-native/src/test/java/org/graalvm/junit/jupiter?rgh-link-date=2025-04-28T10%3A44%3A01Z) once you release next JUnit version that supports `@ParameterizedClass`?

Would it be okay to use a milestone version (5.13.0-M2) only for the tests in `junit-platform-native`? If so, we could add the test right now. WDYT?

common/junit-platform-native/gradle/native-image-testing.gradle
marcphilipp: I see. Thanks for the explanation!

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
dnestoro: This will require another issue to be opened, and should not be mixed into this PR. Please open an issue with this request (e.g. something like "Add testing of JUnit support for JUnit milestone releases")

olpaw (dismissed):

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
marcphilipp: I submitted https://github.com/graalvm/native-build-tools/pull/722 as a draft. I'll update it once 5.13.0 is released.

marcphilipp (commented):
LGTM :+1: 

(I lack permissions to formally approve)

tzezula (approved):

olpaw (approved):

fniephaus (approved):
Thank you, @dnestoro!

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -143,14 +168,44 @@ private TestPlan discoverTestsAndRegisterTestClassesForReflection(Launcher launc
...
         forEachProvider(p -> p.onTestClassRegistered(clazz, nativeImageConfigImpl));
+
+        Class<?>[] declaredClasses = clazz.getDeclaredClasses();
+        for (Class<?> declaredClass : declaredClasses) {
+            registerTestClassForReflection(declaredClass);
+        }
+
+        Class<?>[] interfaces = clazz.getInterfaces();
+        for (Class<?> inter : interfaces) {
+            registerTestClassForReflection(inter);
```
marcphilipp, line 206:
```suggestion
            registerTestClassForReflection(inter, registeredClasses);
```

common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java
```diff
@@ -143,14 +168,44 @@ private TestPlan discoverTestsAndRegisterTestClassesForReflection(Launcher launc
...
+        }
+
         Class<?> superClass = clazz.getSuperclass();
         if (superClass != null && superClass != Object.class) {
             registerTestClassForReflection(superClass);
```
marcphilipp, line 211:
```suggestion
            registerTestClassForReflection(superClass, registeredClasses);
```

marcphilipp (commented):
Good catch on the recursive type declarations! :+1:

dnestoro:
@marcphilipp as discussed offline, last changes you suggested will be added in the separate PR, to avoid losing approvals and re-running all the tests

Eng-Fouad:
@dnestoro Any ETA for 0.10.7 release? Thanks.

dnestoro:
> @dnestoro Any ETA for 0.10.7 release? Thanks.

Hey, we will start the release as soon as these pull requests get merged: 
- https://github.com/graalvm/native-build-tools/pull/719
- https://github.com/graalvm/native-build-tools/pull/712

PR #719, opened 2025-04-23
Experimental support for layered images

melix:
This commit introduces a Gradle DSL to support layered images creation. As of now, this is mainly aimed towards a single use case which is incremental builds. The test project demonstrates how to use the DSL to create a base layer which includes the JDK "java.base" module as well as all external dependencies used by the project.

The DSL builds on top of the binaries concept, by allowing them to declare either that they produce a layer, or that they use one or more layers. The DSL contains methods to make it easier to declare use or creation, as well as supports an easy way to declare that a layer should use only external dependencies. For example, this is how you would create a base layer and consume it from the main binary:

```
graalvmNative {
    binaries {
        libdependencies {
            createLayer {
                modules = ["java.base"]
                jars.from(externalDependenciesOf(configurations.runtimeClasspath))
            }
        }
        main {
            useLayer("libdependencies")
        }
    }
}
```

Note that there is a _binary_ named `libdependencies`, and as soon as the `createLayer` is called, it will offer additional options which are specific to layers (for example declaring the list of packages or modules).

The DSL to create a layer contains the `packages` option, which could be used with automatic extraction of package names, which is why there is code to extract packages from jars, however, this code is currently unused for a reason: these packages can contain dependencies to "optional" modules, which cannot be figured out at build time. Typically, logback will support additional modules and load them dynamically, and if the package is included in the list and that the supporting dependencies are not on classpath, then the layer creation would fail. Therefore, the only reliable option right now is to use the `jars` property to set the list of jars which should belong to the layer.

### DIFF
--- a/docs/src/docs/asciidoc/changelog.adoc
+++ b/docs/src/docs/asciidoc/changelog.adoc
@@ -5,6 +5,10 @@
 
 This version introduces a breaking change: the plugin now requires Java 17 to run.
 
+=== Gradle plugin
+
+- Added experimental support for layered images
+
 == Release 0.10.6
 
 === Gradle plugin
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JUnitFunctionalTests.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JUnitFunctionalTests.groovy
@@ -45,7 +45,7 @@ import org.graalvm.buildtools.gradle.fixtures.AbstractFunctionalTest
 
 class JUnitFunctionalTests extends AbstractFunctionalTest {
     def "test if JUint support works with various annotations, reflection and resources"() {
-
+        debug=true
         given:
         withSample("junit-tests")
 
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to any
+ * person obtaining a copy of this software, associated documentation and/or
+ * data (collectively the "Software"), free of charge and under any and all
+ * copyright rights in the Software, and any and all patent rights owned or
+ * freely licensable by each licensor hereunder covering either (i) the
+ * unmodified Software as contributed to or provided by such licensor, or (ii)
+ * the Larger Works (as defined below), to deal in both
+ *
+ * (a) the Software, and
+ *
+ * (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+ * one is included with the Software each a "Larger Work" to which the Software
+ * is contributed by such licensors),
+ *
+ * without restriction, including without limitation the rights to copy, create
+ * derivative works of, display, perform, and distribute the Software and make,
+ * use, sell, offer for sale, import, export, have made, and have sold the
+ * Software and the Larger Work(s), and to sublicense the foregoing rights on
+ * either these or other terms.
+ *
+ * This license is subject to the following condition:
+ *
+ * The above copyright notice and either this complete permission notice or at a
+ * minimum a reference to the UPL must be included in all copies or substantial
+ * portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package org.graalvm.buildtools.gradle
+
+import org.graalvm.buildtools.gradle.fixtures.AbstractFunctionalTest
+import org.graalvm.buildtools.gradle.fixtures.GraalVMSupport
+import org.graalvm.buildtools.utils.NativeImageUtils
+import spock.lang.Requires
+
+@Requires(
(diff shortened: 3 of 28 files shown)


native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy
```diff
@@ -0,0 +1,155 @@
...
+            succeeded ':nativeLibdependenciesCompile', ':nativeCompile'
+        }
+
+        when:
+        def builder = new ProcessBuilder()
+            .directory(testDirectory.toFile())
+            .inheritIO()
+            .command("build/native/nativeCompile/layered-mn-app${IS_WINDOWS?".exe":""}")
+        def env = builder.environment()
+        env["LD_LIBRARY_PATH"] = testDirectory.resolve("build/native/nativeLibdependenciesCompile").toString()
```
cstancu, line 137: Instead of setting `LD_LIBRARY_PATH` you can set `-H:LinkerRPath={layer_path}` when building the app layer to point to the directory containing the shared object(s) of underlying layer(s). The value gets passed to the linker as the `-rpath`. The path is relative to the directory where the executable resides. For example `-H:LinkerRPath=$ORIGIN/lib` points to the `lib` directory next to the executable.

melix: I had tried this without success. That said I'm not sure it's a good idea: this is only used for running the app, it is not required when building. So if we change this to use `-H:LinkerRPath`, wouldn't that mean that the binary wouldn't be portable anymore if the base layer paths change?

samples/layered-java-application/build.gradle
```diff
@@ -0,0 +1,85 @@
...
+        libdependencies {
+            createLayer {
+                modules = ["java.base"]
+                jars.from(externalDependenciesOf(configurations.runtimeClasspath))
+            }
+        }
+        main {
+            useLayer("libdependencies")
+        }
+    }
```
cstancu, line 84: Can you please add an alternative layered configuration where only the `java.base` is in the base layer, e.g., something like:
```
        libjavabase {
            createLayer {
                modules = ["java.base"]
            }
        }
        main {
            useLayer("libjavabase")
        }
```
That would be useful for testing when something goes wrong by adding all the dependencies in the base layer. Similarly for the Micronaut test.

cstancu, line 84: Can you please also add a configuration that could switch off the layered build? In your early demo if I recall correctly you had an `-incremental` flag. It would be useful to be able to easily switch between these 3 scenarios (and maybe other in the future):
 - a regular build (no layers)
 - a layered build that has just `java.base` in the base layer
 - a layered build that has `java.base` and all dependencies in the base layer

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy
cstancu: That's correct, if you set `-rpath` then that gets hardcoded in the binary, but it's common for apps to depend on dirs relative to the main binary. I remember you had issues with `LD_LIBRARY_PATH` before, that's why I mentioned it. Also, you need to always be sure to not overwrite `LD_LIBRARY_PATH` in case other apps depend on it.

samples/layered-java-application/build.gradle
melix: Not sure what you mean, since the DSL already allows that. Or do you want to have an additional test? We try not to add too many redundant tests, since the build is already taking ages.

melix: Again it's unclear to me what you want. These are tests of the plugin itself, it doesn't quite make sense to add these options, since they wouldn't be used by our tests.

cstancu: Moreover, for demo purposes, it would be nice if we could run the scenarios without the functional tests infrastructure, as standalone apps. Can you please add some minimal documentation with the bare bone commands one needs to do that?

cstancu: I see, then what would be the best way to showcase this plugin functionality to an end user?

cstancu: Maybe after we merge the plugin changes we can add a demo project to the Micronaut repo?

melix: Yes. Although once this is merged and NBT released, I could implement this directly in the Micronaut Gradle plugins, where the options would make more sense, and the configuration could be hidden.

cstancu (approved):
The PR looks good to me with regards to the layer specific options/configs and the tests. @vjovanov @dnestoro please do a review of the plugin specific parts.

docs/src/docs/asciidoc/changelog.adoc
```diff
@@ -1,6 +1,12 @@
 [[changelog]]
 == Changelog
 
+== Release 0.10.7
```
dnestoro, line 4: This should be `0.11.0`, right?

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JUnitFunctionalTests.groovy
```diff
@@ -45,7 +45,7 @@ import org.graalvm.buildtools.gradle.fixtures.AbstractFunctionalTest
 
 class JUnitFunctionalTests extends AbstractFunctionalTest {
     def "test if JUint support works with various annotations, reflection and resources"() {
-
+        debug=true
```
dnestoro, line 48: Is this a leftover?

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -332,26 +345,46 @@ private void configureAutomaticTaskCreation(Project project,
                 task.setDescription("Executes the " + options.getName() + " native binary");
                 task.getImage().convention(imageBuilder.flatMap(BuildNativeImageTask::getOutputFile));
                 task.getRuntimeArgs().convention(options.getRuntimeArgs());
+                var useLayers = options.getLayers()
+                    .stream()
+                    .filter(layer -> layer instanceof UseLayerOptions)
+                    .collect(Collectors.toList());
```
dnestoro, line 351: Since we updated the minimal project version to `17`, can't we use just `toList()` here?

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
```diff
@@ -116,10 +124,68 @@ public Provider<RegularFile> getClasspathJar() {
...
+                    var create = (CreateLayerOptions) layer;
+                    layerCreateName = layer.getLayerName().get();
+                    arg.append("-H:LayerCreate=");
+                    arg.append(layerCreateName).append(".nil");
+                    var modules = create.getModules().get();
+                    jarsClasspath = create.getJars();
+                    boolean hasModules = !modules.isEmpty();
+                    boolean hasPackage = create.getPackages().isPresent() && !create.getPackages().get().isEmpty();
+                    var createLayerJars = jarsClasspath.getFiles();
+                    boolean hasJars = !createLayerJars.isEmpty();
```
dnestoro, line 149: Minor: maybe merge this line with the one above like:
`boolean hasJars = !jarsClasspath.getFiles().isEmpty();`
since `createLayerJars` is only used to calculate `hasJars` value.

dnestoro (approved):
It looks good to me. I just left few suggestions

PR #740, opened 2025-06-09
Add support for running integration tests with Maven

dnestoro:
In order to execute integration tests with Native Build Tools Maven plugin, we should support Maven `failsafe` plugin. Previously this only worked because we had fallback mode for JUnit tests which unintentionally discovered integration tests as well. Since we removed that fallback mode, we have to support integration tests properly.

This PR adds JUnit listeners configuration when using Maven `failsafe` plugin and collect `environmentVariables` and `systemPropertyVariables` provided through it.


melix (approved):

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/IntegrationTest.groovy
```diff
@@ -0,0 +1,30 @@
+package org.graalvm.buildtools.maven;
+
+class IntegrationTest extends AbstractGraalVMMavenFunctionalTest {
+    def "run integration tests with failsafe plugin"() {
+        withSample("integration-test")
+
+        when:
+        mvn'-Pnative', 'verify'
```
marcphilipp, line 8:
```suggestion
        mvn '-Pnative', 'verify'
```

marcphilipp (commented):
LGTM :+1:

melix (approved):

alvarosanchez (approved):

PR #726, opened 2025-05-17
[gradle] Added JavaLauncher capability to agent run

MartelliEnrico:
This is a first implementation of the javaLauncher support for agent run and metadataCopy task.

I'm still having issues with local testing, since it's not really explained how to handle the multiple version of java required by the plugin. Also the agent run tests are failing when run from the functionalTest project, but if replicated by hand they are working fine (if setting the default java sdk to a graalvm one).

I'm more than thankful if someone can help me with the tests situation 😄 

Fixes #724

PR #751, opened 2025-07-23
[gradle] Added JavaLauncher capability to agent run and metadataCopy task

### LINKED ISSUE #724: [gradle] Fix agent toolchain detection
With the latest improvements in Gradle around nativeImageCapable toolchain detection and configuration, it's a shame that the native-agent-run tasks are still not configured to read the javaLauncher (or a more generic executable) property, like the nativeCompile plugin is already doing. This would enable a fully hands-free experience, where no GRAALVM_HOME variable should be defined, and where everything is handled by gradle toolchains.

https://github.com/graalvm/native-build-tools/blob/8a7b6af17fe8f1dc511627533a104a387752a0e2/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/actions/MergeAgentFilesAction.java#L95

MartelliEnrico:
This change adds support for javaLauncher for the agent run and metadataCopy task.
The agent run will use the javaLauncher set for the run task that's instumenting.
The metadataCopy task will have a new javaLauncher property (maybe not 100% correctly named since it's not really a Java task, we could change it with something else).

I opened a new PR since #726 was automatically closed while I was fixing my branch.

Fixes #724

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 # Project versions
-nativeBuildTools = "0.11.0"
+nativeBuildTools = "0.11.1-SNAPSHOT"
 metadataRepository = "0.3.22"
 
 # External dependencies
--- a/native-maven-plugin/reproducers/issue-144/pom.xml
+++ b/native-maven-plugin/reproducers/issue-144/pom.xml
@@ -56,8 +56,8 @@
     <properties>
         <java.version>1.8</java.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <native.maven.plugin.version>0.11.0</native.maven.plugin.version>
-        <junit.platform.native.version>0.11.0</junit.platform.native.version>
+        <native.maven.plugin.version>0.11.1-SNAPSHOT</native.maven.plugin.version>
+        <junit.platform.native.version>0.11.1-SNAPSHOT</junit.platform.native.version>
         <imageName>example-app</imageName>
         <mainClass>org.graalvm.demo.Application</mainClass>
     </properties>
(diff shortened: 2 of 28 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/MetadataCopyTask.java
```diff
@@ -110,6 +117,12 @@ public void overrideOutputDirectories(List<String> outputDirectories) {
         getOutputDirectories().set(outputDirectories);
     }
 
+    private void initJavaLauncher() {
+        JavaToolchainSpec toolchain = getProject().getExtensions().getByType(JavaPluginExtension.class).getToolchain();
+        Provider<JavaLauncher> defaultJavaLauncher = javaToolchainService.launcherFor(toolchain);
+        getJavaLauncher().convention(defaultJavaLauncher);
```
melix, line 123: Does this work if _no_ toolchain is defined in the project?

melix (approved):
Looks good to me, thanks! I only have a doubt regarding the conventional value in the task: I wonder if it's not going to fail if no toolchain is declared in build scripts.

MartelliEnrico:
When I comment out the definition of the default java toolchain from my project, this is what I see in the debugger for the value of line 121: `{languageVersion=unspecified, vendor=any vendor, implementation=vendor-specific, nativeImageCapable=false}`

Even when nothing is set manually, Gradle returns a generic JavaToolchainSpec, that will match any installed and recognized JVM. This is somewhat important since in the `findNativeImageExecutable()` method the javaLauncher property will have an actual value, so if the toolchainDetection is disabled it will get used instead of the alternative graalvmHomeProvider.

MartelliEnrico:
Hi @melix
When I try and run locally the failed tests, it fails aswell, but if I try and directly run the failing sample with this command it works without any issues:

```bat
.\gradlew.bat -p .\samples\java-application-with-reflection run -Pagent=standard
```

Do you have any suggestions on how I could try and fix this error?

PR #755, opened 2025-07-28
Add scheduled job that checks and updates version of reachability metadata repository

dnestoro:
Check if the a version of [Reachability metadata repository](https://github.com/oracle/graalvm-reachability-metadata) is released, and update the version we use in [libs.versions.toml](https://github.com/graalvm/native-build-tools/blob/master/gradle/libs.versions.toml#L4)

### DIFF
--- a/.github/workflows/update-reachability-metadata-version.yml
+++ b/.github/workflows/update-reachability-metadata-version.yml
@@ -1,0 +1,67 @@
+name: "Create scheduled reachability metadata version updates"
+
+on:
+  # Reachability metadata releases happen on 1st and 15th of each month -> we should look for updates the day after
+  schedule:
+    - cron: "0 0 2 * *"
+    - cron: "0 0 16 * *"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  actions: write
+
+jobs:
+  get-latest-release-version:
+    runs-on: ubuntu-latest
+    outputs:
+      new_version: ${{ steps.get_release.outputs.new_version }}
+    steps:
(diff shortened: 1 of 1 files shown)


dnestoro:
The problem could be that for Native Build Tools pull requests, user needs to sign OCA. Since this process is automatic I am not sure if we can merge PR that was created by "actions@github.com".

.github/workflows/update-reachability-metadata-version.yml
```diff
@@ -0,0 +1,67 @@
...
+      - uses: actions/checkout@v4
+      - name: Get current metadataRepository version
+        id: get_current_version
+        run: |
+          value=$(grep '^metadataRepository\s*=' gradle/libs.versions.toml | sed -E 's/.*=\s*"(.*)"/\1/')
+          echo "old_version=$value" >> $GITHUB_OUTPUT
+
+  update-version-and-create-pull-request:
+    runs-on: ubuntu-latest
+    needs: [get-latest-release-version, get-currently-used-version]
```
fniephaus, line 41: Do you really need three jobs for this? Can't this be done in just one?

dnestoro: Personally, this looks more readable to me. I split the code into three logically independent chunks, with clear responsibilities for each of them. I could put everything in one job, but in my opinion, that would make the code harder to read without providing significant benefits (unless I am missing some advantages of having all the steps in one job).

dnestoro:
Closing this PR as it will be implemented internally. cc @ezzarghili

PR #757, opened 2025-07-30
Allow the test classes directory to be configured for the NativeTestMojo

thegridman:
The `NativeTestMojo` hard codes the location of the test classes to be the `project.build.testOutputDirectory`. Some Maven projects have test classes in other locations, for example under `src/main/java` which compiles them to the `project.build.outputDirectory`. 

This PR allows the test classes location to be configured for the `NativeTestMojo` in the same way that plugins such as Surefire and Failsafe do by setting the `<testClassesDirectory>` element in the plugin configuration. The default is the `project.build.testOutputDirectory`.


samples/java-application-with-tests-in-src-main/pom.xml
```diff
@@ -0,0 +1,340 @@
...
+                                    <goal>compile-no-fork</goal>
+                                </goals>
+                                <phase>package</phase>
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <skip>false</skip>
+                            <imageName>${imageName}</imageName>
+                            <fallback>false</fallback>
+                            <testClassesDirectory>${project.build.outputDirectory}</testClassesDirectory>
```
vjovanov, line 113: I am wondering if we can go without an extra parameter. How does the `surefire` plugin know where are the tests?

dnestoro (commented):
This looks okay to me (with few questions that I left). I will wait for @alvarosanchez or @melix approval because I don't know if this is potentially against some Maven policies or conventions (other than that, this looks like a reasonable feature).

thegridman:
I have been doing some more work on this based on @vjovanov comment about getting the test classes directory from the Surefire plugin. The way this plugin works currently, it merges in the system properties and environment variables from both the surefire and failsafe plugins if they are configured in the project. So it sounds like it should be simple to also pull in the test classes directory from those plugins but it starts to become a can of worms.

If my pom.xml only contained is a simple surefire config like the one below then it is easy.
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.5.3</version>
    <configuration>
        <testClassesDirectory>${project.build.outputDirectory}</testClassesDirectory>
    </configuration>
</plugin>
```

But we use executions in plugin configs like below, which your plugin does not support
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.5.3</version>
    <executions>
        <execution>
            <goals>
                <goal>test</goal>
            </goals>
            <phase>test</phase>
            <configuration>
                <testClassesDirectory>${project.build.outputDirectory}</testClassesDirectory>
            </configuration>
        </execution>
    </executions>
</plugin>
```

I do not think this is too big an issue, you just need to search through any executions defined for the Surefire plugin to find the one with the `test` goal. I cannot see a reason why you'd have more than one execution for the Surefire plugin so that should work.

The complications start when we have use both Surefire and Failsafe. As I said above, the plugin is currently merging system properties and environment variables from both Surefire and Failsafe and this might not be correct. Typically Surefire is for unit tests that execute all in a single JVM, which is how the native test plugin runs tests. Failsafe is typically used for integration tests where each test class runs in a forked JVM so tests are more isolated. 

I could have a pom.xml file like this:
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <systemPropertyVariables>
            <my.property>one</my.property>
        </systemPropertyVariables>
    </configuration>
</plugin>

<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-failsafe-plugin</artifactId>
    <configuration>
        <systemPropertyVariables>
            <my.property>two</my.property>
        </systemPropertyVariables>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>integration-test</goal>
                <goal>verify</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

Both Surefire and Failsafe set a value for the `my.property` system property, but they set different values. Which will be used? Currently the Failsafe value just because that is the order the native plugin processes things.

It would be reasonably simple if a mojo knew which phase was executing when `execute()` was called, but Maven does not supply this information to the mojo. 

I believe the way to fix this is to have two native test mojos, the current `NativeTestMojo` that has the `test` goal and by default binds to the `test` phase, then a second mojo `NativeIntegrationTestMojo` that has an `integration-test` goal and by default binds to the `integration-test` phase. 

The bulk of the code would remain in the `NativeTestMojo` (or could all be extracted to an abstract test mojo). When either of the mojos is executed we then know which goal is running, either `test` or `integration-test` and then the correct values can be pulled from Surefire and Failsafe.

I had a go at writing the code for this, but I did  not want to do too much in case there are big objections to it.

thegridman:
OK, so maybe I shouldn't believe everything said on the web before looking at all the Maven code. You can determine things like the goal and phase in a mojo. One `NativeTestMojo` would be fine as it can work out whether it is in running the `test` or `integration-test` goal

thegridman:
I have pushed new commits to this PR to fix the following issues.

- The native plugin merged configuration from both Surefire and Failsafe into the native plugin test configuration regardless of whether the goal being executed is `test` or `integration-test`. This is fixed by having two goals for the native plugin, `test` and `integration-test` and then the plugin knows how to correctly configure itself.
- The native plugin only obtains configuration from the Surefire or Failsafe `<configuration>` section and ignored any configuration in the `<executions>`. This is fixed by making the native plugin properly look for executions in the Surefire and Failsafe plugins.
- The native plugin added a JUnit listener to the Surefire and Failsafe configuration to create a file of test identifiers which is how the native plugin identifies the tests to run. The problem here is that if a project uses both Surefire and Failsafe to run unit and integration tests. First Surefire runs and creates a test-id file, then native unit tests are run with that file, then Failsafe runs and creates another test-id file for the integration tests, then the native intergration test runs and sees both test id files so runs unit and integration tests. This is fixed by putting the test id files in different directories.

With the current code in this PR it looks like the native plugin can now be used to run both unit and integration tests using both Surefire and Failsafe. Which is how I would think most project use these plugins.

Even with these fixes though, we could not run integration tests in our project because we rely on Failsafe fork mode where every test class runs in a fork. I can see no way the native plugin can do this as it just uses Unit to run the tests and Unit does not support forks. You would end up duplicating a lot of what Falsafe/Surefire do to support forks, which is a massive job, or somehow being able to extend the functionality of Surefire/Failsafe so that when they execute a fork, they actually run a native image. Either way a big amount of work.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java
```diff
@@ -88,21 +89,39 @@
...
+     * This field will be set at execution time from either the Surefire or
+     * Failsafe plugin configurations.
+     */
+    private String testClassesDirectory;
+
     @Override
     protected void populateApplicationClasspath() throws MojoExecutionException {
         super.populateApplicationClasspath();
-        imageClasspath.add(Paths.get(project.getBuild().getTestOutputDirectory()));
+        imageClasspath.add(Path.of(testClassesDirectory));
```
vjovanov, line 124: @melix are we handling this in Gradle properly?

samples/java-application-with-tests-in-src-main/pom.xml
```diff
@@ -0,0 +1,341 @@
...
+            </build>
+        </profile>
+    </profiles>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M5</version>
```
vjovanov, line 285: Should we use `3.5.3` consistently?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java
```diff
@@ -181,25 +201,112 @@ public void execute() throws MojoExecutionException {
...
+
+        if (!testPluginProcessed) {
+            // neither Surefire nor Failsafe has an execution for the current goal,
+            // so use the configuration for whichever plugin's default goal matches
+            // the current goal
+            if (mojoExecution.getGoal().equals(TEST_GOAL) && surefire != null) {
+                // the current goal is "test" which is Surefire's default goal
+                processTestPluginConfig(surefire, null);
+                getLog().info("Using configuration from " + surefire.getArtifactId());
+            } else if (mojoExecution.getGoal().equals(INTEGRATION_TEST_GOAL) && failsafe != null) {
```
vjovanov, line 230: Could we put this check into the `NativeIntegrationTestMojo`?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java
```diff
@@ -181,25 +201,112 @@ public void execute() throws MojoExecutionException {
...
+                getLog().info("Using configuration from " + failsafe.getArtifactId());
+            }
+        }
+    }
+
+    /**
+     * Process the configuration from the Surefire or Failsafe plugins.
+     * <p>
+     * This method will check to see whether the plugin has an execution
+     * matching the goa for this plugin. If it does the plugin's configuration
```
vjovanov, line 242: the goal?

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeTestMojo.java
thegridman: This part of the code has been removed after I found out that plugins always have a default execution and Maven combines the global and execution configurations together in the model. So the Surefire and Failsafe plugins always  their default executions matching the `test` or `integration-test` goals. The only reason the code still looks in both plugins for the current goal is because someone _might_ bind Surefire to the `integration-test` goal or Failsafe to the `test` goal.

alvarosanchez (approved):
LGTM, but this should be documented somewhere (the fact that we read test classes dir from Failsafe/Surefire), eg in the release notes.

vjovanov (approved):

vjovanov:
@jormundur00 can we rebase this and spec it up?

PR #759, opened 2025-07-31
Prepare SBOM integration for GraalVM 25

rudsberg:
This PR updates the SBOM integration in the `native-maven-plugin` to handle SBOMs being embedded by default in GraalVM 25. SBOMs will be embedded even if `--enable-sbom` is not passed to `native-image`. We have updated our checks accordingly to ensure the base SBOM is generated only when the SBOM feature is enabled.

Additionally, the option `augmentedSBOM` has been renamed to `skipBaseSBOM`. We are discontinuing the use of the term "augmented" and will instead refer to the SBOM produced by `native-maven-plugin` (and consumed by `native-image`) as the "base" SBOM. The new "skip" prefix aligns better with other configuration options. `skipBaseSBOM` defaults to `false` and when set to `true` the base SBOM is not generated. 

With regards to backwards compatibility, I believe the deprecation of `augmentedSBOM` should be OK given that it was just recently introduced for 24 and it's not documented in the public [native-maven-plugin documentation](https://graalvm.github.io/native-build-tools/latest/maven-plugin.html).

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
@@ -132,7 +132,7 @@ public class NativeCompileNoForkMojo extends AbstractNativeImageMojo {
         var config = new SBOMGenerator.Config(isOracleGraalVM(logger), getBuildArgs(), NativeImageUtils.getMajorJDKVersion(getVersionInformation(logger)));
         try {
             sbomGenerator.generateIfSupportedAndEnabled(config);
-        } catch (Exception e) {
+        } catch (Throwable e) {
             logger.warn(String.format("Could not generate base SBOM: %s. Fallback to generating an SBOM without knowledge from 'native-build-tools'.",
                     e.getCause().getMessage()));
         }


rudsberg:
The failing [functional tests](https://github.com/graalvm/native-build-tools/actions/runs/16643748673/job/47099334674?pr=759) seems unrelated to this PR. The failing test is `JavaApplicationFunctionalTest > can build and execute a native image with the Maven plugin when the application has a custom packaging type FAILED` and fails with:
```
org.graalvm.demo.ApplicationTest > message is hello native FAILED
Failures (1):
JUnit Jupiter:ApplicationTest:message is hello native
MethodSource [className = 'org.graalvm.demo.ApplicationTest', methodName = 'usesReflection', methodParameterTypes = '']
=> org.opentest4j.AssertionFailedError: expected: <Hello, native!> but was: <null>
```

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
```diff
@@ -113,78 +111,29 @@ public void execute() throws MojoExecutionException {
...
         try {
-            sbomGenerator.generate();
-        } catch (MojoExecutionException e) {
-            /* Only throw exception for users that explicitly opt-in to using augmented SBOMs. */
-            if (optionWasSet) {
-                throw e;
-            }
-            logger.warn(String.format("Could not generate an augmented SBOM: %s. Fallback to generating a non-augmented SBOM.",
+            sbomGenerator.generateIfSupportedAndEnabled(config);
+        } catch (Exception e) {
```
rudsberg, line 135: We now catch all exceptions to avoid the risk of a runtime exception bypassing this fallback mechanism.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -145,48 +149,38 @@ public SBOMGenerator(
...
-     * Generates an SBOM that will be further augmented by Native Image. The SBOM is stored in the build directory.
-     *
-     * @throws MojoExecutionException if SBOM creation fails.
-     */
-    public void generate() throws MojoExecutionException {
         Path sbomPath = Paths.get(outputDirectory, SBOM_FILENAME);
         try {
-            /* Suppress the output from the plugin. */
-            int loggingLevel = logger.getThreshold();
-            logger.setThreshold(Logger.LEVEL_DISABLED);
```
rudsberg, line 189: The logging suppression was removed because it was not working. Setting the logging level like so doesn't restrict the logging in the `executeMojo` call. Another PR could ensure that the output from `cyclonedx-maven-plugin` is suppressed.

rudsberg:
> The failing [functional tests](https://github.com/graalvm/native-build-tools/actions/runs/16643748673/job/47099334674?pr=759) seem unrelated to this PR, but I'm looking into it.
>
> The failing test is `JavaApplicationFunctionalTest > can build and execute a native image with the Maven plugin when the application has a custom packaging type FAILED` and it fails with:
>
> ```
> org.graalvm.demo.ApplicationTest > message is hello native FAILED
> Failures (1):
> JUnit Jupiter:ApplicationTest:message is hello native
> MethodSource [className = 'org.graalvm.demo.ApplicationTest', methodName = 'usesReflection', methodParameterTypes = '']
> => org.opentest4j.AssertionFailedError: expected: <Hello, native!> but was: <null>
> ```
I've identified and resolved the issue. Here’s what went wrong:
* This PR correctly generates a base SBOM for GraalVM 25 builds that embed SBOMs by default. In other words, previously all functional tests running on GraalVM 25 and that did not use `--enable-sbom` did not exercise the `SBOMGenerator` code path, but now they do.
* That caused [this line to fail](https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/JavaApplicationFunctionalTest.groovy#L107). The reason was that it uses `mvnDebug`, causing excessive printing by the `cyclonedx-maven-plugin` on version `2.8.1` that we were using, leading to a `java.lang.OutOfMemoryError`.

The excessive printing seemed like a bug, so I tried updating `cyclonedx-maven-plugin` to the latest version (`2.9.1`), which indeed solved it. It now logs in a reasonable manner.

To me it looks like an oversight that `mvnDebug` is used [here](https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/JavaApplicationFunctionalTest.groovy#L107), so I reverted it to use `mvn` like the other tests. The only test now using `mvnDebug` is [this one](https://github.com/graalvm/native-build-tools/blob/master/native-maven-plugin/src/functionalTest/groovy/org/graal).

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/sbom/SBOMGenerator.java
```diff
@@ -277,8 +299,10 @@ private void augmentComponentNode(JSONObject componentNode, Set<ArtifactAdapter>
             if (optionalArtifact.isPresent()) {
                 ArtifactAdapter artifact = optionalArtifact.get();
                 JSONArray packageNamesArray = new JSONArray();
-                List<String> sortedPackageNames = artifact.packageNames.stream().sorted().collect(Collectors.toList());
-                sortedPackageNames.forEach(packageNamesArray::put);
+                if (artifact.prunable) {
```
rudsberg, line 302: Minor orthogonal improvement. 

(Shaded dependencies are correctly marked as non-prunable since we cannot be sure the `packageNames` are accurate. Therefore, leaving the `packageNames` array empty increases clarity.)

dnestoro (dismissed):

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
```diff
@@ -44,88 +44,67 @@ package org.graalvm.buildtools.maven
...
-    /**
-     * If user sets {@link NativeCompileNoForkMojo#AUGMENTED_SBOM_PARAM_NAME} to true then Native Image should be
-     * invoked with '--enable-sbom' and an SBOM should be embedded in the image.
-     */
-    @Requires({ supportedAugmentedSBOMVersion() })
-    def "sbom is embedded when only the augmented sbom parameter is used (but not the '--enable-sbom' buildArg)"() {
+    @Requires({ supportsBaseSBOM() })
+    def "base sbom generation is skipped when skipBaseSBOM is true"() {
         withSample 'java-application'
+        deleteTargetDir()
```
melix, line 72: :x: This shouldn't be needed. Tests are executed in isolation, the samples are copied in a fresh directory, so if this is needed then it means we have a bigger issue to solve.

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
```diff
@@ -44,88 +44,67 @@ package org.graalvm.buildtools.maven
...
     }
 
-    @Requires({ EE() && unsupportedJDKVersion() })
-    def "error is thrown when augmented sbom parameter is used with EE but with an unsupported JDK version"() {
-        withSample 'java-application'
-
-        when:
-        mvn '-Pnative-augmentedSBOM-only', '-DquickBuild', '-DskipTests', 'package'
+    private def deleteTargetDir() {
+        file("target").deleteDir()
```
melix, line 101: :x: Very fishy call

melix (changes requested):
Looks mostly ok. I cannot tell if the removal of the previous property is acceptable or not, that's probably for @vjovanov to tell.

native-maven-plugin/src/functionalTest/groovy/org/graalvm/buildtools/maven/SBOMFunctionalTest.groovy
melix: Are you sure you didn't run a test directly from the `samples` directory, and that it accidentally added a `target` directory that would be copied when the test starts? Saying this because you wouldn't be the first one to try to run tests directly from `samples` instead of copying.

rudsberg: Ah, yes – that's it. Thanks for spotting this. I have pushed a diff that removes the usage of `deleteTargetDir`.

melix (approved):

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeCompileNoForkMojo.java
rudsberg: I agree, that's better. I'll update accordingly.

dnestoro:
>The 1 failing test is the same test that is failing on master ([link](https://github.com/graalvm/native-build-tools/actions/runs/17290969652/job/49078022110)), so it's unrelated to this PR.
>
>CC: @dnestoro

It looks like a regression in `latest-ea` GraalVM release. I will check

dnestoro (approved):

PR #765, opened 2025-08-22
Run native tests without running tests on JVM first

### LINKED ISSUE #764: The native JUnit tests cannot be executed without the prior run of JVM tests

In projects where test suites run for hours, or days, it is unrealistic to demand that native testing is preceded with the execution on the JVM. 

Additionally, the developer experience is not ideal: when testing locally, one needs to wait for both tests to see the result of their work.

We need to modify JUnit in such a way that test discovery can be executed without executing the actual tests. That way we could execute the discovery at build time from the Native Build Tools plugin.

dnestoro:
Problem: At the moment, every time we run native tests (with both Maven or Gradle), we must run tests on JVM first, in order to collect `test-ids` required for native run. With more complex tests, this additional step increases time required for tests execution.

Idea: We should use JUnit's [dry-run mode](https://docs.junit.org/current/user-guide/#launcher-api-dry-run-mode) to just collect `test-ids` without executing tests on JVM.

Solution: we can set `junit.platform.execution.dryRun.enabled` system property to true when executing tests on JVM ([maven change](https://github.com/graalvm/native-build-tools/pull/765/files#diff-1eff98b60d5a4c9aad6024aa570fc33d1241d14f9d01bac8c26c052ea38512f5R228), [gradle change](https://github.com/graalvm/native-build-tools/pull/765/files#diff-3ca19e7e49fe64feca3a1e88279c961b56e03450666583025c711da44cfb73bbR683)).

NOTE: Since this approach changes the default behavior, we should bump major version for the next release.

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/utils/SharedConstants.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/utils/SharedConstants.java
@@ -78,6 +78,7 @@ public interface SharedConstants {
     String AGENT_OUTPUT_DIRECTORY_MARKER = "{output_dir}";
     String AGENT_OUTPUT_DIRECTORY_OPTION = "config-output-dir=";
     String METADATA_REPO_URL_TEMPLATE = "https://github.com/oracle/graalvm-reachability-metadata/releases/download/%1$s/graalvm-reachability-metadata-%1$s.zip";
+    String SKIP_JVM_TESTS = "skipJVMTests";
     /**
      * The default metadata repository version. Maintained for backwards
      * compatibility.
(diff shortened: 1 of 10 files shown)


melix (changes requested):

dnestoro:
> See [junit-team/junit-framework#3164 (comment)](https://github.com/junit-team/junit-framework/pull/3164#issuecomment-1464967048).

@sbrannen can we somehow extend `dryRun` mode to collect ids for all tests? If we are going to break some existing support, can we add a new mode (like `dryRunAll`) to collect all test ids? 

I see that `dryRun` mode [is completely separated,](https://github.com/junit-team/junit-framework/blob/main/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineExecutionOrchestrator.java#L104) but can we follow [standard execution pipeline](https://github.com/junit-team/junit-framework/blob/main/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineExecutionOrchestrator.java#L108) and do something like:
```
// do all test preparations as in the normal execution
if (!dryRunAll()) {
    executeTest()
}
// post processing where we register test-id
```

If that idea is not possible to achieve for some reason, do you have some other proposal?

cc @marcphilipp @vjovanov @sdeleuze @melix

vjovanov:
@jormundur00 will take this ticket over.

Issue #760, opened 2025-08-05
Plugin incompatible with Gradle 9.0.0 for multi-module projects


rkudryashov:
**Describe the bug**
The GraalVM native build tools plugin works correctly with Gradle 8.14.3 but fails with Gradle 9.0.0 in multi-module projects. The first module builds successfully, but the second module fails with configuration resolution errors.

**To Reproduce**

Clone https://github.com/rkudryashov/multi-module-native-repro
Build second module: `./gradlew :service2:bootBuildImage`

**Logs**

````
```
./gradlew :service2:bootBuildImage                                                                                                                                                                                                                                       
> Task :service2:collectReachabilityMetadata FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':service2:collectReachabilityMetadata'.
> Failed to query the value of task ':service2:collectReachabilityMetadata' property 'metadataService'.
   > Could not isolate value org.graalvm.buildtools.gradle.internal.GraalVMReachabilityMetadataService$Params_Decorated@70dbf065 of type GraalVMReachabilityMetadataService.Params
      > Resolution of the configuration ':service1:detachedConfiguration9' was attempted without an exclusive lock. This is unsafe and not allowed.

* Try:
> For more information, please refer to https://docs.gradle.org/9.0.0/userguide/viewing_debugging_dependencies.html#sub:resolving-unsafe-configuration-resolution-errors in the Gradle documentation.
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (Powered by Develocity).
> Get more help at https://help.gradle.org.

BUILD FAILED in 597ms
6 actionable tasks: 1 executed, 5 up-to-date

```
````

**System Info (please complete the following information):**
 - OS: Windows
 - Java Version: 24
 - Plugin version: 0.11.0

heruan:
> At this stage I suspect the issue is in fact in the Spring Boot plugin, I cannot reproduce without it.

@melix have you been able to reproduce this with Spring Boot? I can provide a sample if that helps.

wilkinsona:
That's not what I described in https://github.com/spring-projects/spring-boot/issues/47308. The issue occurs with Spring Boot 4.0.0-M3 and, perhaps more importantly, [without Spring Boot at all](https://github.com/spring-projects/spring-boot/issues/47308#issuecomment-3327689839).

rkudryashov:
[...]

PR #766, opened 2025-08-22
Add testing of Gradle 9

melix:
While investigating #760 , I added testing of Gradle 9 to our build. This caused a few cascading issues:

1. we have to upgrade the Kotlin plugin version we use in tests to the latest 1.9.x branch
2. doing so breaks the `KotlinApplicationWithTestsFunctionalTest` with a class initialized at build time error

This error only happens when testing with GraalVM 17, not with 23. The workaround I have found is to add:

```
kotlin.annotation.AnnotationTarget
kotlin.annotation.AnnotationRetention
```

to the list of classes initialized at build time, but I'm not completely sure this is the right thing to do. /cc @sbrannen @marcphilipp 

In any case it seems that the original issue doesn't occur outside of the context of Spring.


marcphilipp:
> The workaround I have found is to add [...] to the list of classes initialized at build time

Which versions of JUnit did you test this with?

marcphilipp:
Sounds ok to me

dnestoro (approved):

PR #787, opened 2025-10-16
add more junit classes when stict heap is enabled

n0tl3ss:
Add more junit classes when

`--strict-image-heap` is enabled


vjovanov:
@n0tl3ss is there a reproducer for this issue? Can we add a test that exercises this?

jvukicev:
Per author, this is stale and most likely unneeded.

PR #791, opened 2025-10-27
Enhance Project Documentation and Testing

vjovanov:


build-logic/common-plugins/src/main/groovy/org.graalvm.build.github-actions-helper.gradle
```diff
@@ -2,7 +2,7 @@ import groovy.json.JsonOutput
 
 def matrixDefault = [
         "java-version": [ "17" ],
-        "os": [ "ubuntu-22.04", "windows-latest" ]
+        "os": [ "ubuntu-22.04", "windows-latest", "macos-latest" ]
```
vjovanov: I added `macos` as I don't see how we can ship without testing that.

build.gradle.kts
```diff
@@ -40,5 +40,6 @@
  */
 
 plugins {
+    // Aggregates build configurations for all project modules. Required for multi-module builds and reporting.
```
melix: Looks like a redundant comment. AI generated?

PR #794, opened 2025-11-04
Add class introduced in 5.14.1/6.0.1

marcphilipp:
`MethodSegmentResolver` is used during discovery and therefore also needs to be initialized at build time on JDK <= 21.


phipag:
Hi all, I would love a review here since it affects my native tests. See failing CI here: https://github.com/aws-powertools/powertools-lambda-java/pull/2267

PR #795, opened 2025-11-10
Generate dynamic access metadata and provide it to the classpath when passing "--emit build-report" as a build argument

jormundur00:
In this PR we add the Maven mojo and Gradle task for generating and providing to the native image build classpath a `dynamic-access-metadata.json` file, when emitting a Build Report (when the `--emit build-report` build argument is present) and the used release of the `graalvm-reachability-metadata` repository contains the `library-and-framework-list.json` file.

The generated JSON file maps all of the artifacts present on the classpath that exist in the `library-and-framework-list.json` to the list of their transitive dependencies (as they are marked as "metadata provided-for" by the parent artifact). It is later used by the dynamic access tab of the native image Build Report to mark these "provided-for" dependencies as being covered by native image and not "needing investigation". The JSON file follows the schema at: https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/assets/dynamic-access-metadata-schema-v1.0.0.json.

Both the Maven mojo and the Gradle task can be ran standalone, but will automatically run when emitting a Build Report and running the native image build mojo/task.

Fixes: https://github.com/graalvm/native-build-tools/issues/801.

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/GenerateDynamicAccessMetadata.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/GenerateDynamicAccessMetadata.java
@@ -51,8 +51,9 @@ import org.gradle.api.artifacts.result.ResolvedArtifactResult;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.artifacts.result.ResolvedDependencyResult;
 import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.provider.MapProperty;
 import org.gradle.api.provider.Property;
-import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.OutputFile;
 import org.gradle.api.tasks.TaskAction;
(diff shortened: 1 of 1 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/GenerateDynamicAccessMetadata.java
```diff
@@ -0,0 +1,202 @@
...
+ * <a href="https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/assets/dynamic-access-metadata-schema-v1.0.0.json">schema</a>.
+ */
+public abstract class GenerateDynamicAccessMetadata extends DefaultTask {
+    private static final String LIBRARY_AND_FRAMEWORK_LIST = "library-and-framework-list.json";
+    private static final String ARTIFACT = "artifact";
+    private static final String METADATA_PROVIDER = "metadataProvider";
+    private static final String PROVIDES_FOR = "providesFor";
+
+    @Internal
+    public abstract Property<Configuration> getRuntimeClasspath();
```
melix: :x: This is not compatible with Gradle's configuration cache. We shouldn't use `Configuration` as input (probably a `Property<ResolvedComponentResult>`).

PR #811, opened 2026-01-13
Support compatibility mode in GraalVM

vjovanov:
And the `debug` method in the feature was pulling in hosted code into runtime code which was crashing native image. So we made it a separate class and functionality.

Fixes: #812


jormundur00 (dismissed):

melix:
Isn't it doing exactly what we always said that library authors shouldn't do? (add flags in `native-image.properties`)?

vjovanov:
@melix this is a good point. The matter is that `-H:+CompatibilityMode` spawns two different behaviors for Native Image.

The alternative is to simply ignore all of the `--initialize-at-build-time` and `--features` flags in Native Image when `-H:+CompatibilityMode` is enabled.

melix:
> The alternative is to simply ignore all of the --initialize-at-build-time and --features flags in Native Image when -H:+CompatibilityMode is enabled.

This sounds like a better fix to me.

Issue #818, opened 2026-01-19
native-maven-plugin does not provide environment variables to the image builder


jormundur00:
**Describe the bug**
The `environment` tag of the `native-maven-plugin` only provides environment variables to the produced binary, but doesn't provide these variables to the image builder.

**To Reproduce**

Add environment variables to the `pom.xml` with:
```
<configuration>
    <environment>
        <variable1>value1</variable1>
        <envVariable2>value2</envVariable2>
    </environment>
</configuration>
```
then proceed to create an image where `System.getenv()` is printed/logged in a static initializer. The environment variables provided through the tag are not visible and don't get printed during the image build.

**Expected behavior**
The environment variables provided through the `<environment>` tag SHOULD be visible during the image build.

**System Info (please complete the following information):**
 - OS: [e.g. `Linux Ubuntu-22.04`]
 - GraalVM Version [e.g. `25.0.1 EE`]
 - Java Version [e.g. `17`]
 - Plugin version [e.g. `native-maven-plugin:0.11.4`]

PR #819, opened 2026-01-19
Make environment variables set in the pom.xml of the native-maven-plugin be visible in the image builder

jormundur00:
In this PR, we make environment variables set in the `pom.xml` `<environmentVariables>` tag visible to the native image static analysis by providing them to the `native-image` builder using the `-E<variable>=<value>` flag (consistent with how `systemProperties` are currently handled).

Also in this PR, to avoid documentation mismatches and breaking older builds, we add aliases for both `environmentVariables` and `systemPropertyVariables` properties. 

Current documentation and samples frequently use these names even though the plugin only recognized the field names `environment` and `systemProperties`. By using aliases, we ensure:

* **Consistency**: The naming now matches the `maven-surefire-plugin` and `maven-failsafe-plugin` conventions.
* **Backwards Compatibility**: Existing builds using the `<environment>` and `<systemProperties>` tags will continue to work without modification.

Alternatively we could just rename the fields to `environmentVariables` and `systemPropertyVariables` (and avoid aliases), but that could potentially break older builds that used the old tag names.

Fixes: https://github.com/graalvm/native-build-tools/issues/818

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
@@ -156,10 +156,10 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     @Parameter(property = "excludeConfig")
     protected List<ExcludeConfigConfiguration> excludeConfig;
 
-    @Parameter(property = "environmentVariables")
+    @Parameter(alias = "environmentVariables", property = "environmentVariables")
     protected Map<String, String> environment;
 
-    @Parameter(property = "systemPropertyVariables")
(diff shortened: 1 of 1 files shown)


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -156,10 +156,10 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     @Parameter(property = "excludeConfig")
     protected List<ExcludeConfigConfiguration> excludeConfig;
 
-    @Parameter(property = "environmentVariables")
+    @Parameter(alias = "environmentVariables", property = "environmentVariables")
```
vjovanov: We need this for bundles?

PR #822, opened 2026-01-21
Support `-H:+CompatibilityMode` in build tools

vjovanov:
When the Native Image [Compatibility Mode](https://github.com/oracle/graal/pull/12756) flag is enabled (`-H:+CompatibilityMode`), the Maven native test goal is automatically short-circuited: native-image based JUnit tests are skipped and tests run on the JVM via Surefire/Failsafe instead.

- What happens
  - If `-H:+CompatibilityMode` is detected, the `native:test` goal is skipped early and JVM tests execute (Surefire/Failsafe).

- Detection sources
  - Effective native-image build arguments configured for the plugin/goal (e.g. `-DbuildArgs=...`).
  - Environment variable `NATIVE_IMAGE_OPTIONS`:
    - From the process environment.
    - From Surefire/Failsafe environment passed via plugin configuration (`<environmentVariables>NATIVE_IMAGE_OPTIONS</environmentVariables>`).

- Behavior details
  - The native test goal returns early right after environment setup when Compatibility Mode is enabled.
  - The plugin logs once at INFO level with the exact text:
    - "Compatibility Mode detected (-H:+CompatibilityMode); skipping native-image test goal, JVM tests will run instead."
  - JVM tests still run normally; no extra configuration needed.
  - Default behavior is unchanged when the flag is absent; native-image tests run as before.

- How to trigger and verify
  - CLI build args example:
    ```bash
    mvn -Pnative -DbuildArgs=-H:+CompatibilityMode test
    ```
  - Using `NATIVE_IMAGE_OPTIONS`:
    - Shell export:
      ```bash
      NATIVE_IMAGE_OPTIONS="-H:+CompatibilityMode" mvn -Pnative test
      ```
    - Or configure `NATIVE_IMAGE_OPTIONS` in the Surefire/Failsafe plugin environment within the POM.
  - Expected output:
    - The INFO log appears once with the exact text above.
    - No native-image build/run steps execute for the test goal.
    - JVM test results are reported by Surefire/Failsafe.

- Resetting back to native tests
  - Remove `-H:+CompatibilityMode` from build args and ensure `NATIVE_IMAGE_OPTIONS` does not contain it.


native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/CompatibilityModeGatingTest.groovy
```diff
@@ -0,0 +1,103 @@
...
+import org.graalvm.buildtools.gradle.dsl.GraalVMExtension
+import org.graalvm.buildtools.gradle.dsl.NativeImageOptions
+import org.gradle.api.Project
+import org.gradle.api.plugins.JavaPlugin
+import org.gradle.testfixtures.ProjectBuilder
+import org.gradle.api.JavaVersion
+import spock.lang.Requires
+import spock.lang.Specification
+
+@Requires({ JavaVersion.current().isCompatibleWith(JavaVersion.toVersion(25)) })
```
jormundur00: This points to the `JAVA_HOME` used by the gradle wrapper, and not the `GRAALVM_HOME`, so the tests get skipped when attempting to run with `JAVA_HOME` set as JDK 17 (for the gradle), and `GRAALVM_HOME` set as JDK 25 (for the image build). I think we should use `@Requires({ NativeImageUtils.getMajorJDKVersion(GraalVMSupport.getGraalVMHomeVersionString()) >= 25 })` like it's done in `LayeredApplicationFunctionalTest` here (also to stay consistent with the Maven test. that runs properly with this setup).

PR #833, opened 2026-02-05
Prefer GraalVM's Java when running with agent

### LINKED ISSUE #581: Native Image Agent Uses `JAVA_HOME` Instead of `GRAALVM_HOME`

When running `gradle -Pagent clean test` agent picks `JAVA_HOME` even though `GRAALVM_HOME` is set. The problem is that we can set `JAVA_HOME` to some earlier version (lets say JDK17) and `GRAALVM_HOME` to the later one (e.g. JDK21), and therefor agent will generate metadata with one JDK version while the Native Image will be created with the different one ([example](https://github.com/oracle/graalvm-reachability-metadata/pull/382) where different metadata can be generated with different Java versions). 

Make `native-image-agent` use `GRAALVM_HOME` as default if possible. Only if `GRAALVM_HOME` is not configured, use `JAVA_HOME`.

At least mention in the documentation that agent uses `JAVA_HOME`

vjovanov:
Fixes #581.

## Summary

- Prefer the GraalVM Java executable from `GRAALVM_HOME` or `JAVA_HOME` when running tasks instrumented with the native-image-agent.
- Set a matching `javaLauncher` for `Test` and `JavaExec` tasks so Gradle 9 does not reject a mismatch between `executable` and `javaLauncher`.
- Keep the executable override as a fallback for other `JavaForkOptions` implementations.
- Add functional coverage for the agent-instrumented `run` task using GraalVM Java.

## Validation

- `./gradlew :native-gradle-plugin:functionalTest -DgradleVersion=9.0.0 --tests org.graalvm.buildtools.gradle.JavaApplicationWithAgentFunctionalTest --fail-fast`
- `./gradlew :native-gradle-plugin:configCacheFunctionalTest -DgradleVersion=9.0.0 --tests org.graalvm.buildtools.gradle.JavaApplicationWithAgentFunctionalTest --fail-fast`
- `./gradlew :native-gradle-plugin:test :native-gradle-plugin:inspections`
- `git diff --check`

### DIFF
--- a/native-gradle-plugin/docs/functional/tracing-agent.md
+++ b/native-gradle-plugin/docs/functional/tracing-agent.md
@@ -16,6 +16,14 @@ Every task that implements `JavaForkOptions` is eligible for instrumentation. Th
 `tasksToInstrumentPredicate` setting may narrow that set. Non-matching tasks must be skipped
 without failing the build.
 
+### 2.1 Agent Java runtime
+
+When agent instrumentation is enabled, the task must prefer the Java executable from a valid
+`GRAALVM_HOME`. If that is unavailable, the task may use `JAVA_HOME` only when the installation
+identifies itself as GraalVM. A regular JDK in `JAVA_HOME` must not replace the task's configured
+launcher or executable. When instrumentation is disabled, the plugin must leave the task's Java
+runtime selection unchanged.
+
 ## 3. Agent modes
 
 The Gradle DSL must expose standard, conditional, direct, and disabled agent modes using the
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithAgentFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithAgentFunctionalTest.groovy
@@ -212,8 +212,9 @@ class JavaApplicationWithAgentFunctionalTest extends AbstractFunctionalTest {
         withSample("java-application-with-reflection")
         buildFile << """
             tasks.named('run') {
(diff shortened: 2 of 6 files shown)


jvukicev:
samples/java-application-with-extra-sourceset/build.gradle now passes org.graalvm.demo.ApplicationFeature twice: the new binaries.main block adds -H:Features=..., and the existing binaries.all block below still adds --features=....

The sample still builds, but Native Image reports the duplicate option and the Groovy/Kotlin versions now diverge. Can we remove one of the two entries?

.github/workflows/test-native-gradle-plugin.yml
```diff
@@ -160,9 +160,9 @@ jobs:
       - name: "🔧 Install GraalVM (dev)"
         uses: graalvm/setup-graalvm@60c26726de13f8b90771df4bc1641a52a3159994 # v1
         with:
-          java-version: 'latest-ea'
```
jvukicev, line 163: This shouldn't be done in this PR I believe (or not at all)?

.github/workflows/test-native-maven-plugin.yml
```diff
@@ -108,9 +108,9 @@ jobs:
       - name: "🔧 Install GraalVM (dev)"
         uses: graalvm/setup-graalvm@60c26726de13f8b90771df4bc1641a52a3159994 # v1
         with:
-          java-version: 'latest-ea'
```
jvukicev, line 111: Same here?

jormundur00 (approved):
Re-review complete. The duplicate feature build argument is resolved: both sample DSLs now provide a single -H:Features entry. I found no remaining blocking issues; the launcher/runtime changes and their Gradle 9 configuration-cache coverage look good.

PR #875, opened 2026-04-13
Fix order of access filter files by moving default to the start

### LINKED ISSUE #874: default access-filter.json overrides user settings
**Description**

`access-filter.json` is not being applied.

**To Reproduce**

Repository in question: <https://github.com/ArloL/drifty>

The [access-filter.json](https://github.com/ArloL/drifty/blob/116185c9f4062fe9e9f670fbe2e5e1e5db76af50/src/main/resources/META-INF/native-image/access-filter.json):
```
{ "rules": [
    { "excludeClasses": "**" },
    { "includeClasses": "io.github.arlol.githubcheck.**" },
    …
]}
```

The [plugin configuration](https://github.com/ArloL/drifty/blob/116185c9f4062fe9e9f670fbe2e5e1e5db76af50/pom.xml#L88)

```bash
$ ./mvnw -Pgraal clean test
```

See that `./target/native/agent-output/test/reachability-metadata.json` contains e.g. `"type": "void"`

**Expected behavior**
`./target/native/agent-output/test/reachability-metadata.json` only contains types from the explicitly included packages

**Logs**

IMHO The relevant output of `./mvnw -Pgraal --debug clean test`. The important bit is the order of the access-filter-file arguments.
```
[DEBUG] Forking command line: /bin/sh -c cd '/Users/aokeeffe/Developer/drifty' && '/Users/aokeeffe/.local/share/mise/installs/java/graalvm-community-25.0.2/bin/java' '-agentlib:native-image-agent=config-output-dir=/Users/aokeeffe/Developer/drifty/target/native/agent-output/test/session-{pid}-{datetime},access-filter-file=/Users/aokeeffe/Developer/drifty/src/main/resources/META-INF/native-image/access-filter.json,access-filter-file=/var/folders/23/y5fw68rd7bvfmqt06k80mr780000gn/T/agent-config/access-filter.json' '-jar' '/Users/aokeeffe/Developer/drifty/target/surefire/surefirebooter-20260413144952793_3.jar' '/Users/aokeeffe/Developer/drifty/target/surefire' '2026-04-13T14-49-52_748-jvmRun1' 'surefire-20260413144952793_1tmp' 'surefire_0-20260413144952793_2tmp'
```

**System Info**
 - OS: `macOS 15.7.3`
 - GraalVM Version: `25.0.2 CE`
 - Java Version: `25`
 - Plugin version: `native-maven-plugin:1.0.0`

**Additional context**

I believe the issue is that the [default access-filter.json](https://github.com/graalvm/native-build-tools/blob/master/common/utils/src/main/resources/access-filter.json) overrides the user provided files since it gets [added to the end of the list](https://github.com/graalvm/native-build-tools/blob/master/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java#L104). If it were the first entry, my acess-filter.json would override it.

Please let me know if I can provide more details 😊

ArloL:
This fixes #874 by adding the default access-filter.json file as the first argument to the agent.

Please let me know if I messed up anything or should add anything else.

`./gradlew test` ran successfully

`./gradlew :native-maven-plugin:inspections` only showed me pre-existing issues

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
@@ -101,9 +101,9 @@ public class AgentConfiguration implements Serializable {
     }
 
     public List<String> getAgentCommandLine() {
-        addDefaultAccessFilter();
         List<String> cmdLine = new ArrayList<>(agentMode.getAgentCommandLine());
         appendOptionToValues("caller-filter-file=", callerFilterFiles, cmdLine);
+        appendOptionToValues("access-filter-file=", List.of(getDefaultAccessFilter()), cmdLine);
         appendOptionToValues("access-filter-file=", accessFilterFiles, cmdLine);
         addToCmd("builtin-caller-filter=", builtinCallerFilter, cmdLine);
         addToCmd("builtin-heuristic-filter=", builtinHeuristicFilter, cmdLine);
@@ -141,18 +141,12 @@ public class AgentConfiguration implements Serializable {
         }
     }
 
-    private void addDefaultAccessFilter() {
-        if (accessFilterFiles == null) {
-            // this could only happen if we instantiated disabled agent configuration
-            return;
-        }
-
+    private String getDefaultAccessFilter() {
         String tempDir = System.getProperty("java.io.tmpdir");
         Path agentDir = Path.of(tempDir).resolve("agent-config");
         Path accessFilterFile = agentDir.resolve(ACCESS_FILTER_PREFIX + ACCESS_FILTER_SUFFIX);
         if (Files.exists(accessFilterFile)) {
-            accessFilterFiles.add(accessFilterFile.toString());
-            return;
+            return accessFilterFile.toString();
         }
 
         try (InputStream accessFilterData = AgentConfiguration.class.getResourceAsStream(DEFAULT_ACCESS_FILTER_FILE_LOCATION)) {
@@ -176,7 +170,7 @@ public class AgentConfiguration implements Serializable {
                 logger.info(accessFilterFile + " already exists. Delete " + tmpAccessFilter);
             }
 
-            accessFilterFiles.add(accessFilterFile.toString());
+            return accessFilterFile.toString();
         } catch (IOException e) {
             throw new RuntimeException("Cannot add default access-filter.json", e);
         }
--- a/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
@@ -48,15 +48,12 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import java.util.logging.Logger;
 
 public class AgentConfiguration implements Serializable {
 
     private static final String ACCESS_FILTER_PREFIX = "access-filter";
     private static final String ACCESS_FILTER_SUFFIX = ".json";
     private static final String DEFAULT_ACCESS_FILTER_FILE_LOCATION = "/" + ACCESS_FILTER_PREFIX + ACCESS_FILTER_SUFFIX;
-
-    private static final Logger logger = Logger.getGlobal();
 
     private final Collection<String> callerFilterFiles;
     private final Collection<String> accessFilterFiles;
--- a/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
@@ -104,9 +104,9 @@ public class AgentConfiguration implements Serializable {
         if (!isEnabled()) {
             return List.of();
         }
-        addDefaultAccessFilter();
         List<String> cmdLine = new ArrayList<>(agentMode.getAgentCommandLine());
         appendOptionToValues("caller-filter-file=", callerFilterFiles, cmdLine);
+        cmdLine.add("access-filter-file=" + getDefaultAccessFilter());
         appendOptionToValues("access-filter-file=", accessFilterFiles, cmdLine);
         addToCmd("builtin-caller-filter=", builtinCallerFilter, cmdLine);
         addToCmd("builtin-heuristic-filter=", builtinHeuristicFilter, cmdLine);
@@ -144,18 +144,12 @@ public class AgentConfiguration implements Serializable {
         }
     }
 
-    private void addDefaultAccessFilter() {
-        if (accessFilterFiles == null) {
-            // this could only happen if we instantiated disabled agent configuration
-            return;
-        }
-
+    private String getDefaultAccessFilter() {
         String tempDir = System.getProperty("java.io.tmpdir");
         Path agentDir = Path.of(tempDir).resolve("agent-config");
         Path accessFilterFile = agentDir.resolve(ACCESS_FILTER_PREFIX + ACCESS_FILTER_SUFFIX);
         if (Files.exists(accessFilterFile)) {
-            accessFilterFiles.add(accessFilterFile.toString());
-            return;
+            return accessFilterFile.toString();
         }
 
         try (InputStream accessFilterData = AgentConfiguration.class.getResourceAsStream(DEFAULT_ACCESS_FILTER_FILE_LOCATION)) {
@@ -179,7 +173,7 @@ public class AgentConfiguration implements Serializable {
                 logger.info(accessFilterFile + " already exists. Delete " + tmpAccessFilter);
             }
 
-            accessFilterFiles.add(accessFilterFile.toString());
+            return accessFilterFile.toString();
         } catch (IOException e) {
             throw new RuntimeException("Cannot add default access-filter.json", e);
         }
--- a/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
@@ -162,12 +162,7 @@ public class AgentConfiguration implements Serializable {
             if (accessFilterData == null) {
                 throw new IOException("Cannot access data from: " + DEFAULT_ACCESS_FILTER_FILE_LOCATION);
             }
-
-            try {
-                Files.createDirectories(agentConfigDir);
-            } catch (FileAlreadyExistsException e) {
-                logger.info("Skip creation of directory " + agentConfigDir + " (already created).");
-            }
+            Files.createDirectories(agentConfigDir);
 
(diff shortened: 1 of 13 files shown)


jormundur00 (changes requested):
Thanks for the PR and for the clear reproducer in the linked issue. I think the underlying report is valid: Native Image access filters are order-sensitive, and multiple `access-filter-file` entries are applied in command-line order. With the current NBT ordering, a user filter can be followed by the built-in default filter, whose broad `includeClasses: "**"` rule can effectively undo a user's earlier broad exclusion. Moving the NBT default filter before user-provided access filters is the right direction because it keeps the built-in baseline while letting explicit user configuration take final precedence.

One issue with this implementation: `getDefaultAccessFilter()` is now called unconditionally from `getAgentCommandLine()`. Previously `addDefaultAccessFilter()` returned early when `accessFilterFiles == null`, which is the disabled-agent constructor path. After this change, `new AgentConfiguration().getAgentCommandLine()` returns an `access-filter-file=...` option instead of an empty list. A disabled agent configuration should not create or return agent options.

Could you keep the disabled guard while preserving the new order? For example, only add the default access filter when the configuration is enabled, then append user access filters after it.

It would also be good to add a small unit test in `common/utils/src/test/java/org/graalvm/buildtools/agent/AgentConfigurationTest.java` covering:
- disabled configuration returns an empty command line
- standard/managed configuration emits the default access filter before user access filters

The existing agent tests are mostly plugin functional tests, so this shared command-line ordering would benefit from direct common-utils coverage.

ArloL:
Thanks, @jormundur00 for your review 🙏

I pushed some changes that hopefully address your concerns. One decision I made that I'm unsure about: it felt iffy to test the creation of the default file using the global java.io.tmpDir. So I opted for injecting it into the AgentConfiguration so that I can override it in the test. Then I reconsidered even using java.io.tmpDir at all and decided against it. This is somewhat cleaner but also changes all call sites - let me know if I should change it to something else. Looking forward to your feedback.

jormundur00 (commented):
Thanks, the updates address my concerns.

I also agree with moving this away from `java.io.tmpdir`. Keeping the generated default filter under the build-local agent config directory seems like the cleaner solution here, and it avoids relying on shared global temp state.

ArloL:
Thanks 🙏

Two more things I've added:
* The try-catch around [.createDirectory / .createDirectories](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/nio/file/Files.html#createDirectories(java.nio.file.Path,java.nio.file.attribute.FileAttribute...)) was hiding a potential bug. The FileAlreadyExistsException is thrown when the path is an existing _file_ not a directory.
* The copying & atomic move is unnecessary now

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
jvukicev: This logger is now unused? We should remove it if there's no need for it.

jvukicev, line 48: Unused import now?

jvukicev: Unused import

jvukicev: Will become unused if/when the logger field gets removed.

ArloL:
Thanks for the thorough check. I've addressed your comments.

ArloL:
Please do let me know if there are blockers remaining in the PR 🙏

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -43,43 +43,44 @@
...
-    private static final Logger logger = Logger.getGlobal();
-
     private final Collection<String> callerFilterFiles;
     private final Collection<String> accessFilterFiles;
     private final Boolean builtinCallerFilter;
     private final Boolean builtinHeuristicFilter;
     private final Boolean experimentalPredefinedClasses;
     private final Boolean experimentalUnsafeAllocationTracing;
     private final Boolean trackReflectionMetadata;
+    private final Path agentConfigDir;
```
jvukicev: **[P1] Keep `AgentConfiguration` serializable**

`AgentConfiguration` implements `Serializable`, but `java.nio.file.Path` does not (`UnixPath` fails with `NotSerializableException`). This can break the Gradle configuration-cache path, which carries this configuration through serializable provider transformations. Please store a serializable representation such as `String` and call `Path.of(...)` when creating the default filter.

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -98,12 +100,16 @@ public AgentConfiguration(Collection<String> callerFilterFiles,
...
     }
 
     public List<String> getAgentCommandLine() {
-        addDefaultAccessFilter();
+        if (!isEnabled()) {
+            return List.of();
+        }
         List<String> cmdLine = new ArrayList<>(agentMode.getAgentCommandLine());
         appendOptionToValues("caller-filter-file=", callerFilterFiles, cmdLine);
+        cmdLine.add("access-filter-file=" + getDefaultAccessFilter());
```
jvukicev: **[P2] Specify the new access-filter precedence rule**

This repository has recently adopted grund specifications, and this is a shared, user-visible behavior change: the built-in access filter now precedes user filters so user configuration has final precedence; disabled configurations emit no agent options.

Please document it in [`common/docs/functional-spec.md`](https://github.com/graalvm/native-build-tools/blob/master/common/docs/functional-spec.md), directly below `### 3.1 Agent modes`, for example:

```md
#### 3.1.1 Access-filter precedence

When the shared agent model emits access-filter options, it must emit Native Build Tools’
built-in default access filter before all user-provided access-filter files. This preserves the
built-in baseline while allowing user configuration to take final precedence. A disabled agent
configuration must emit no agent command-line options.
```

This creates `§FS-common-libraries.3.1.1`. Please cite it at the implementation and test, for example:

```java
// The default filter precedes user filters so user rules have final precedence.
// §FS-common-libraries.3.1.1
cmdLine.add("access-filter-file=" + getDefaultAccessFilter());
```

If you’d prefer, I can add this documentation update in a follow-up.

jormundur00 (dismissed):
Sorry for the late update, please address the following comments when you get the chance.

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
ArloL: I added it. Hope it's the way you need it 😊

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
ArloL: Thanks so much for this. I've added a test for the serialization bit to reproduce this and addressed it using String.

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -141,42 +149,20 @@ private void addToCmd(String option, Boolean value, List<String> cmdLine) {
...
-
-            try {
-                Files.move(tmpAccessFilter, accessFilterFile, StandardCopyOption.ATOMIC_MOVE);
-            } catch (FileAlreadyExistsException e) {
-                Files.delete(tmpAccessFilter);
-                logger.info(accessFilterFile + " already exists. Delete " + tmpAccessFilter);
-            }
-
-            accessFilterFiles.add(accessFilterFile.toString());
+            Files.createDirectories(agentConfigDirPath);
```
jvukicev: **[P1] Avoid creating the filter while Gradle evaluates agent arguments**

`getAgentCommandLine()` is evaluated while Gradle calculates the task graph. Creating `build/native/agent-config/access-filter.json` here invalidates the next configuration-cache invocation because the filesystem entry appeared after the cache snapshot. This is why `JavaApplicationWithAgentFunctionalTest` currently fails in CI: its second invocation expects “Reusing configuration cache”, but Gradle reports that this file was created.

Please keep argument construction side-effect-free and have the Gradle adapter produce this file through a task with a declared output.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/NativeExtension.java
```diff
@@ -130,7 +131,7 @@ public void afterProjectsRead(MavenSession session) {
                 Xpp3Dom configurationRoot = (Xpp3Dom) nativePlugin.getConfiguration();
                 AgentConfiguration agent;
                 try {
-                    agent = AgentUtils.collectAgentProperties(session, configurationRoot);
+                    agent = AgentUtils.collectAgentProperties(session, configurationRoot, Path.of(target, "native", "agent-config"));
```
jvukicev: **[P2] Keep the eagerly generated Maven filter outside `target`**

`afterProjectsRead` runs before Maven lifecycle goals. With `mvn clean test`, the filter is created here and then deleted by `clean` before Surefire starts with the already-injected `access-filter-file` path.

Please use a per-session directory under `java.io.tmpdir` for Maven rather than `${project.build.directory}`.

common/utils/src/main/java/org/graalvm/buildtools/agent/AgentConfiguration.java
```diff
@@ -98,12 +100,18 @@ public AgentConfiguration(Collection<String> callerFilterFiles,
...
     public List<String> getAgentCommandLine() {
-        addDefaultAccessFilter();
+        if (!isEnabled()) {
+            return List.of();
+        }
         List<String> cmdLine = new ArrayList<>(agentMode.getAgentCommandLine());
         appendOptionToValues("caller-filter-file=", callerFilterFiles, cmdLine);
+        // The default filter precedes user filters so user rules have final precedence.
+        // §FS-common-libraries.3.1.1
+        cmdLine.add("access-filter-file=" + getDefaultAccessFilter());
```
jvukicev: **[P3] Put the default filter before direct-mode options**

This fixes precedence for structured `accessFilterFiles`, but direct mode accepts raw user options that may include `access-filter-file=...`. Those options are copied into `cmdLine` before the built-in default is added, so the default still overrides the user’s direct-mode filter.

Please prepend the default before the mode options while preserving the relative order of the user’s direct options.

vjovanov:
@ArloL we are planning architectural changes so we need to merge the PR.  Would you like to finish it, or us?

ArloL:
you can go ahead an finish it 😊

jormundur00 (approved):
I've fixed the last remaining blockers and will merge the PR. You can expect it to land in the next release. Thank you for the contribution @ArloL

PR #894, opened 2026-05-21
Add grund repository specification

### LINKED ISSUE #893: Add grund specification to repository
## Summary
Add a grund-based specification to the repository so maintainers and automation agents have stable citations for project motivation, goals, module boundaries, functional behavior, and architecture.

## Proposed scope
- Add `.agents/grund.toml` and repository agent instructions.
- Add component-level `functional-spec.md` and `architecture.md` files under `docs/spec/`.
- Validate the specification with `grund check`.

## Notes
This should use a branch on the upstream repository so other maintainers with a clone of `graalvm/native-build-tools` can fetch the branch directly.

jormundur00:
## Summary
- add grund configuration and agent instructions
- add component-level `functional-spec.md` and `architecture.md` specs under `docs/spec/`
- document project motivation, goals, module boundaries, decisions, and roadmap

Closes #893

## Verification
- `grund config validate .`
- `grund check .`


docs/spec/goals.md
```diff
@@ -0,0 +1,21 @@
+# GOAL-001-build-tool-native-image-workflows: Gradle and Maven users can build and test native images through native build-tool workflows
```
vjovanov: Non-goals are as important as goals here. We will want to decline various proposals based on non-goals.

I would add another kind `non-goals.md` and state that: 
1. We don't want to make flags for flags that can be accessed by native-image.
2. We don't want to support features that can be done with existing build tools capabilities unless they are excessively complex due to native image multi-target nature.

docs/spec/goals.md
```diff
@@ -0,0 +1,21 @@
...
+§FS-001-gradle-plugin-native-image-workflow and §FS-002-maven-plugin-native-image-workflow.
+
+# GOAL-002-shared-native-image-behavior-stays-consistent: Shared native-image behavior remains consistent across Gradle and Maven
+
+Behavior that is not inherently tied to a build tool should live in common modules and be reused
+by both plugins. The shared layer covers native-image utility behavior, resource model analysis,
+reachability metadata lookup, and JUnit native support. The relevant specs are
+§REPO-001-module-boundaries and §AR-003-shared-common-libraries.
+
+# GOAL-003-repository-fixtures-protect-real-build-scenarios: Samples and functional tests protect real build scenarios
```
vjovanov: We need to add a goal for backwards compatibility. This is more of a requirement, than a goal so we could experiment with another kind REQ.

common/utils/src/main/java/org/graalvm/buildtools/utils/SharedConstants.java
```diff
@@ -47,6 +47,7 @@
 import java.util.Collections;
 import java.util.List;
 
+// §FS-003-metadata-and-resource-workflows.1.4
```
jvukicev: Can we merge this with the comment right under it? Just keep the citation at the bottom of the comment?

common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/FileSystemRepository.java
```diff
@@ -57,6 +57,7 @@
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+// §FS-003-metadata-and-resource-workflows.5, §AR-003-shared-common-libraries.3.2
```
jvukicev: IF possible, we should add small javadocs regarding what this (and other like this) class does, and then keep the citations as a part of that javadoc.

docs/spec/gradle-plugin/architecture.md
```diff
@@ -0,0 +1,139 @@
+# AR-001-gradle-plugin-boundary: The Gradle plugin module adapts shared behavior to Gradle APIs
```
jvukicev: I'd remove these "numbers" from tags (such as `001` here), as `AR-gradle-plugin-boundary` should already be unique on its own. Numbers just get stuff confusing in my opinion.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/tasks/BuildNativeImageTask.java
```diff
@@ -84,6 +84,7 @@
 import static org.graalvm.buildtools.gradle.internal.NativeImageExecutableLocator.graalvmHomeProvider;
 import static org.graalvm.buildtools.utils.SharedConstants.EXECUTABLE_EXTENSION;
 
+// §FS-001-gradle-plugin-native-image-workflow.2.1, §AR-001-gradle-plugin-boundary.3.2
```
jvukicev: Just like the other comment, move the citation to the javadocs under it.

docs/spec/build-infra/architecture.md
```diff
@@ -0,0 +1,115 @@
+# AR-005-build-documentation-and-release-infrastructure: Repository infrastructure is separate from product runtime behavior
```
jvukicev: 1. `build-infra` is missing a `functional-spec.md`,
2. Maybe we shouldn't have a separate directory for each of these "subprojects" with its own architecture and functional-spec. Perhaps we should just have one `.md` for each of these (such as `gradle-plugin.md`, `maven-plugin.md`, `common.md`, etc.)

jvukicev: Also, when refactoring the markdowns to be `build-infra.md` etc., try not to just "list spam" everything (and in other `.md` files). It feels like we have too many subsections that just have one line of text in them and such, which we can probably aggregate/remove.

On a different note, when combining previous architecture/functional-spec mds, we can create new tags for these specific sections.

jormundur00 (commented):
We should probably have a top-level `functional-spec.md` and `architecture.md` that reflect the top level outlook of the repo and cite all these other markdown files.

docs/spec/functional-spec.md
```diff
@@ -0,0 +1,112 @@
...
+
+## 8. Documentation and grounded maintenance
+
+User-facing guides remain under the AsciiDoc documentation and root README-style files. Maintainer
+specifications live under `docs/spec/` and provide stable citation targets for design, behavior,
+CI, code comments, and tests. Behavior changes should update the most specific component spec
+before code; placement, dependency, or workflow changes should update the relevant architecture,
+component, CI, or E2E section first (§AR-repository-architecture.5).
+
+## 9. Non-goals
```
vjovanov: This should be in non-goals.md

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Project map
+
+- [GRUND](docs/spec/grund.md): Project motivation
+- [GOAL](docs/spec/goals.md): Project direction and outcomes
+- [NGOAL](docs/spec/non-goals.md): Project non-goals and out-of-scope proposals
+- [REQ](docs/spec/requirements.md): Cross-cutting requirements and constraints
+- [FS](docs/spec/functional-spec.md): Repository functional specification
+- [AR](docs/spec/architecture.md): Repository architecture
+- [GRADLE](docs/spec/gradle-plugin.md): Gradle plugin behavior and architecture
```
vjovanov: I would not mix function and architecture. Those are different things.

Also, I would have a common spec of what we want to support in both plugins, and then separate files for each plugin. That way we can track parity easier.

AGENTS.md
```diff
@@ -0,0 +1,47 @@
...
+- `grund refs <ID>` — every site that cites the ID; add `--summary` for one line per file. Run before renaming or moving a declaration.
+- `grund list` / `grund list --kind FS,AR` — discover IDs if you get lost
+
+### Project map
+
+- [GRUND](docs/spec/grund.md): Project motivation
+- [GOAL](docs/spec/goals.md): Project direction and outcomes
+- [NGOAL](docs/spec/non-goals.md): Project non-goals and out-of-scope proposals
+- [REQ](docs/spec/requirements.md): Cross-cutting requirements and constraints
+- [FS](docs/spec/functional-spec.md): Repository functional specification
```
vjovanov: I would keep the top level for what we really do: functional-spec here makes no sense as the repository doesn't do anything. 

The top-level architecture can contain how we deploy etc.

docs/spec/maven-plugin/functional-spec.md
```diff
@@ -0,0 +1,321 @@
...
+configuration, plugin descriptor, and repository model. This functional contract realizes
+§GOAL-build-tool-native-image-workflows for Maven and depends on the shared behavior in
+§COMMON-libraries and §TESTING-native-tests-and-fixtures.
+
+## 1. Plugin goal surface
+
+The Maven plugin must expose Native Image workflows as Maven goals with lifecycle bindings,
+parameter expressions, dependency resolution scopes, and descriptor metadata that fit Maven users.
+
+### 1.1 Build goals
```
vjovanov: What does it do? Just listing it is not enough, this section should exactly explain what it does and link why.

docs/spec/maven-plugin.md
```diff
@@ -0,0 +1,321 @@
...
+goal may remain as a compatibility alias, but it must direct users toward `compile-no-fork`.
+
+### 1.2 Test goal
+
+The plugin must provide `native:test` for compiling and optionally executing native test images.
+The detailed native test contract is §TESTING-native-tests-and-fixtures.
+
+### 1.3 Metadata and support goals
+
+The plugin must provide goals for generated resource configuration, generated test resource
```
vjovanov: Here we have to have names of the commands and the detailed description of what they do. You can also link it to the spec inline in the code.

docs/spec/decisions.md
```diff
@@ -0,0 +1,33 @@
+# DEC-adopt-component-functional-architecture-docs: Adopt component spec docs beside existing user documentation
```
vjovanov: This is superfluous. Make decisions to be a folder with a README.md. It can be empty.

jvukicev: Done in 873af75a: replaced the single decisions file with docs/spec/decisions/README.md and changed the DEC kind to use that folder. I also removed the root roadmap/RM kind and its references since the existing roadmap only contained grund-adoption follow-up notes rather than a real maintained roadmap.

vjovanov (approved):

PR #895, opened 2026-05-21
Allow empty classpath for layer-create builds

### LINKED ISSUE #892: Allow empty class path for shared layer builds
When building a shared layer, the builder now doesn't automatically adds `$CWD` if no class path is specified. However, when building a normal Native Image, `$CWD` is still automatically added on the class path.

This allows to make layered image simpler to use, as a base layer can sometimes only use java base modules.

The native build tools currently don't allow that, and throw an exception when no class path is specified.

Detecting the shared layer build and avoiding to throw, or adding an option to allow the empty class path would be a nice way to make the user experience better with layers.

jormundur00:
Fixes #892

## Summary
- allow Maven layer-create builds to omit `-cp` when the computed image classpath is empty
- detect the hosted layer-create flag emitted by native-build-tools: `-H:LayerCreate...`
- add Maven coverage for empty classpath layer-create builds and regular empty classpath failures
- add Gradle regression coverage showing existing no-`-cp` behavior for empty classpath layer-create builds

## Testing
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-maven-plugin:test --tests org.graalvm.buildtools.maven.AbstractNativeImageMojoTest`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-gradle-plugin:test --tests org.graalvm.buildtools.gradle.tasks.NativeImageCommandLineProviderTest`

Note: `:native-maven-plugin:checkstyleMain` currently fails on pre-existing unused imports in `NativeTestMojo.java`, unrelated to this change.

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
@@ -53,6 +53,7 @@ import org.apache.maven.project.MavenProject;
 import org.apache.maven.toolchain.ToolchainManager;
 import org.codehaus.plexus.logging.Logger;
 import org.graalvm.buildtools.maven.config.ExcludeConfigConfiguration;
+import org.graalvm.buildtools.model.resources.NativeImageFlags;
 import org.graalvm.buildtools.utils.NativeImageConfigurationUtils;
 import org.graalvm.buildtools.utils.NativeImageUtils;
 import org.graalvm.buildtools.utils.SchemaValidationUtils;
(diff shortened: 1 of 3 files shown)


native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -96,6 +97,8 @@ public abstract class AbstractNativeImageMojo extends AbstractNativeMojo {
     protected static final String NATIVE_IMAGE_META_INF = "META-INF/native-image";
     protected static final String NATIVE_IMAGE_PROPERTIES_FILENAME = "native-image.properties";
     protected static final String NATIVE_IMAGE_DRY_RUN = "nativeDryRun";
+    private static final Pattern LAYER_CREATE_ARG = Pattern.compile(
+            Pattern.quote(NativeImageFlags.LAYER_CREATE) + "(@[^=]*)?=.+");
```
Zeavee, line 101: In the future (when layers will stop being experimental), we will also allow `--layer-create`. We don't have to support it now as I don't really know when we will remove the experimental part for layers, but since this is silent, it might be hard to remember.

Zeavee (commented):
Looks good to me, thank you!

vjovanov (approved):

PR #907, opened 2026-06-05
Fix Gradle system property quoting

### LINKED ISSUE #668: System properties in native-gradle-plugin should be wrapped in double quotes
System properties propagated to java from native-gradle-plugin system properties configuration are wrapped in quotes

{code:java}
graalvmNative {
  binaries {
    main {     
      systemProperties = [propertyName: 'value']
{code}

System.getProperty("propertyName") in java then returns a 7 character string starting and ending with a double-quote character - "value"

(note: reported as GR-60693 to GraalVM and copied here)

jormundur00:
Fixes #668

## What changed

This PR fixes Gradle native-image system property argument construction so property values are no longer wrapped in literal double quotes.

Before this change, Gradle native-image system property values could be emitted as `-Dname="value"`, which changed the actual value seen by native-image even for ordinary property values and values containing spaces.

After this change, Native Build Tools emits `-Dname=value` without injecting literal quotes, while still keeping the value as a single command-line argument.

## Why

Configured Gradle system properties are user-visible native-image inputs. The command-line builder should preserve the configured property value, not rewrite it.

## Example

Before:

A configured value containing spaces could be emitted with extra literal quotes, so native-image received a different value than the one the user configured.

After:

The same property value is emitted without injected quotes and remains a single command-line argument.

## Implementation summary

- Stop wrapping Gradle native-image system property values in literal double quotes when building `-Dname=value` command-line arguments.
- Add a focused regression test that verifies both ordinary and space-containing values are emitted without injected quotes and remain single arguments.

## Validation

- `grund check`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew --no-daemon :native-gradle-plugin:test --tests org.graalvm.buildtools.gradle.tasks.NativeImageCommandLineProviderTest --rerun-tasks`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew --no-daemon :native-gradle-plugin:functionalTest --tests org.graalvm.buildtools.gradle.NativeImageOptionsTest`

### DIFF
--- a/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
+++ b/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
@@ -210,7 +210,7 @@ public class NativeImageCommandLineProvider implements CommandLineArgumentProvid
         cliArgs.add(targetOutputPath);
         options.getSystemProperties().get().forEach((n, v) -> {
             if (v != null) {
-                cliArgs.add("-D" + n + "=\"" + v + "\"");
+                cliArgs.add("-D" + n + "=" + v);
             }
         });
 
--- a/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
+++ b/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
@@ -50,6 +50,37 @@ import spock.lang.Issue
 
 // Protects Gradle native-image argument construction for special image modes. §FS-native-invocation.3.
 class NativeImageCommandLineProviderTest extends AbstractPluginTest {
+    @Issue("https://github.com/graalvm/native-build-tools/issues/668")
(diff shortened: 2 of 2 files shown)


native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/NativeImageCommandLineProvider.java
```diff
@@ -210,7 +210,7 @@ public List<String> asArguments() {
         cliArgs.add(targetOutputPath);
         options.getSystemProperties().get().forEach((n, v) -> {
             if (v != null) {
-                cliArgs.add("-D" + n + "=\"" + v + "\"");
+                cliArgs.add("-D" + n + "=" + v);
```
vjovanov, line 213: Why didn't we change anything in the functional spec?

jvukicev: The original issue existed prior to the `grund` spec existence. With the new spec, `grund` requires gradle code to follow gradle conventions so issues like this shouldn't happen (so that is the part of the spec that covers this).

vjovanov (approved):

PR #917, opened 2026-06-09
Refine grund specification

### LINKED ISSUE #916: Refine grund specification
The grund specification should be refined to be easier to maintain and closer to functional product behavior. It should generally be less verbose and more functional, especially in requirements sections.

Scope:

1. Add a goal stating that Native Build Tools should fetch the latest compatible metadata from the GraalVM Reachability Metadata Repository by default, while still allowing users to pin or override the repository for reproducible and offline builds.
2. Shorten long grund IDs/tags across the repo, and update the matching citations. Names should generally stay around three words unless a longer name is needed for clarity.
3. Make the support matrix explicit about supported JDK, GraalVM Native Image, Gradle, Maven, and metadata repository lines, while keeping the wording concise and functional rather than verbose.

jormundur00:
Closes #916

## Summary
- Add a fresh metadata goal for default GraalVM Reachability Metadata Repository consumption.
- Shorten grund IDs across root, common, Gradle, and Maven specs, and update citations.
- Make the support matrix concise and functional while naming the current supported lines.

## Verification
- `grund check`


docs/spec/functional/resources-and-metadata.md
```diff
@@ -1,11 +1,12 @@
-# FS-resources-and-metadata: Both plugins generate resource config and consume reachability metadata
```
jvukicev, line 1: I'd keep it `resources-and-metadata` here as before (and all citings).

native-gradle-plugin/docs/functional/resources-and-metadata.md
```diff
@@ -1,7 +1,7 @@
-# FS-gradle-resources-and-metadata: Gradle tasks generate resources and consume reachability metadata
+# FS-resources-metadata: Gradle tasks generate resources and consume reachability metadata
```
jvukicev: `resources-and-metadata` like before (just without the`gradle`) + update all citings.

vjovanov (approved):

PR #921, opened 2026-06-09
Document skipNativeBuildForPom for Maven aggregators

### LINKED ISSUE #920: Document skipNativeBuildForPom for Maven multi-module aggregator projects
## Summary

The Maven plugin exposes `skipNativeBuildForPom`, which skips native-image generation when the current Maven project has `packaging=pom`. This is useful for multi-module Maven builds where the parent project is an aggregator and the application main class lives in a child module.

This option is present in the plugin implementation and generated Maven plugin help, and it was mentioned in the 0.10.2 changelog, but it is not described in the main Maven plugin documentation next to `skipNativeBuild` / `skipNativeTests`.

## Motivation

Users configuring `native-maven-plugin` in a parent aggregator POM can hit errors such as:

```text
Error: Main entry point class ... neither found on classpath
```

The parent `packaging=pom` project does not own the application classes, so the native build should normally run in the application module, or the parent should provide shared configuration through `pluginManagement`. When the native profile is inherited by aggregator modules, `skipNativeBuildForPom=true` is the relevant escape hatch.

## Requested documentation

Add user-facing Maven plugin documentation that covers:

- what `skipNativeBuildForPom` does;
- that it defaults to `false` for compatibility;
- when to use it in multi-module Maven builds with parent aggregator POMs;
- that it does not automatically select the application module; users still need to configure/run the native build in the module that owns the main class.

A good location is near the existing `<skipNativeBuild>` option in `docs/src/docs/asciidoc/maven-plugin.adoc`.

## Related

Follow-up from #585.

jormundur00:
Fixes #920

## What changed

This PR documents `<skipNativeBuildForPom>` for Maven aggregator projects.

Before this change, the Maven docs did not clearly explain that `skipNativeBuildForPom` skips native image generation only for projects with `pom` packaging while still allowing one native profile to be shared across aggregator and leaf modules.

After this change, the docs explain the option, its default value, and the multi-module aggregator use case.

## Why

This makes it easier to configure multi-module Maven builds where the aggregator project should not attempt native image generation, but application modules still should.

## Example

Before:

Users configuring one native profile across a parent aggregator POM and application modules had to infer how to prevent the aggregator itself from trying to build a native image.

After:

The docs now show that `<skipNativeBuildForPom>` can be enabled for `pom` packaging while still leaving the actual native build to the module that owns the main class.

## Implementation summary

- Document `<skipNativeBuildForPom>` next to `<skipNativeBuild>` in the Maven plugin docs.
- Explain that it skips native image generation only for Maven projects with `pom` packaging.
- Document the default `false` value for compatibility.
- Cover the multi-module parent aggregator use case and clarify that users still need to run or configure the native build in the application-owning module.

## Validation

- `grund check`
- `git diff --check`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :docs:asciidoctor --rerun-tasks`

### DIFF
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -64,7 +64,6 @@ The following configuration options are available:
    application modules to build their native images. It does not select the application module
    automatically; configure or run the native build in the module that owns the main class. To
    enable it, add:
-// §maven/FS-maven-native-image-builds.2
 [source,xml, role="multi-language-sample"]
 ----
 <skipNativeBuildForPom>true</skipNativeBuildForPom>
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -55,6 +55,20 @@ The following configuration options are available:
 ----
 <skipNativeBuild>true</skipNativeBuild>
 ----
+`<skipNativeBuildForPom>`::
+   To skip generation of the native image only for Maven projects with `pom` packaging, use
+   `<skipNativeBuildForPom>`. This option defaults to `false` for compatibility. It is useful in
+   multi-module builds where a parent aggregator POM shares the native profile or plugin
+   configuration with child modules but does not contain the application main class. Enabling this
+   option prevents native image generation in the aggregator project while still allowing
+   application modules to build their native images. It does not select the application module
+   automatically; configure or run the native build in the module that owns the main class. To
+   enable it, add:
+// §maven/FS-maven-native-image-builds.2
+[source,xml, role="multi-language-sample"]
+----
+<skipNativeBuildForPom>true</skipNativeBuildForPom>
+----
 `<skipNativeTests>`::
    To skip generation and execution of the native image compiled tests, add:
 [source,xml, role="multi-language-sample"]


docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -55,6 +55,19 @@ The following configuration options are available:
 ----
 <skipNativeBuild>true</skipNativeBuild>
 ----
+`<skipNativeBuildForPom>`::
+   To skip generation of the native image only for Maven projects with `pom` packaging, use
```
vjovanov, line 59: Can we infer this from the project?

jvukicev: The plugin already infers `pom` packaging when this option is enabled. We keep it opt-in because packaging alone does not tell us whether a project is intended to build a native image: a POM project can still provide an explicit classpath and main class.

Without this option, users need to avoid inheriting the native execution into the aggregator—for example, put it separately in every application module, or keep configuration in `pluginManagement` and add executions in each leaf. That is more verbose and easier to misconfigure in a larger reactor.

`skipNativeBuildForPom` is therefore a convenience for shared native-plugin configuration: it skips the aggregator execution while application modules continue to build normally.

vjovanov (approved):

PR #925, opened 2026-06-15
Fix custom native test classpath wiring

### LINKED ISSUE #702: Incorrect classpath when configuring additional test suites
**Describe the bug**

Let's say that I have a source set named `integTest` and corresponding test task named `integTest`. I register a new native binary for this test suite: 
```
graalvmNative {
    registerTestBinary("integTest") {
        usingSourceSet(sourceSets.integTest)
        forTestTask(integTest)
    }
}
```
The problem is that `nativeImageIntegTestClasspath` auto-created by the plugin extends from `testRuntimeClasspath`, not `integTestRuntimeClasspath`, resulting in warnings such as `Warning: Could not resolve class XX for reflection configuration. Reason: java.lang.NoClassDefFoundError: XX` and compilation/runtime failures. 

The code snippets that seem relevant are:
https://github.com/graalvm/native-build-tools/blob/67213c369c3e7f016be91e1e2dbdaebef58e0cc5/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java#L751-L761
where `TEST_RUNTIME_CONFIGURATION_NAME` is always used for `baseClasspathConfigurationName`.

The workaround is to set the classpath explicitly in a build script:
```
dependencies {
    integTestImplementation 'org.graalvm.buildtools:junit-platform-native:0.10.5'
}

configurations.nativeImageIntegTestClasspath {
    setExtendsFrom([configurations.integTestRuntimeClasspath])
}
```
Note that the JUnit extension has to be added manually because the plugin adds the extension only to the testImplementation configuration:
https://github.com/graalvm/native-build-tools/blob/67213c369c3e7f016be91e1e2dbdaebef58e0cc5/native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java#L875-L882

**To Reproduce**
See above.

**Expected behavior**
The `nativeImageIntegTestClasspath` extends from `integTestRuntimeClasspath`, and the junit extension is added to the classpath by the plugin. 

**Logs**
See above.

 - OS: MacOS
 - GraalVM Version 17.0.10+11.1 (build 17.0.10+11-LTS-jvmci-23.0-b27)
 - Java Version 17.0.10
 - Plugin version `native-gradle-plugin:0.10.5`

**Additional context**
N/A

jormundur00:
Fixes #702

## What changed

This PR fixes custom Gradle native test binaries so they follow the selected custom source set runtime classpath instead of always falling back to the default `testRuntimeClasspath`.

Before this change, a custom native test binary registered with `registerTestBinary` could accidentally inherit the default test classpath and mask missing dependencies or configuration errors in the custom source set.

After this change, custom native test binaries derive their runtime classpath from the selected source set, while still keeping the expected JUnit native support and generated task wiring.

## Why

Custom native test binaries should behave like the source set they were registered for. Without this fix, users could get passing custom native test setups only because the default test classpath leaked in underneath them.

## Example

Before:

A custom test image such as `integTest` could appear to work even if its own source set dependencies were incomplete, because the default `test` classpath masked the problem.

After:

The same custom test image follows its own source set runtime classpath, so project, JUnit, and plugin-provided native test dependencies come from the intended configuration.

## Implementation summary

- Add `FS-gradle-native-tests.1.1` for custom native test binaries.
- Use the selected source set runtime classpath configuration instead of always using `testRuntimeClasspath`.
- Add functional regression coverage that checks custom test image classpaths include the expected custom source set and plugin-provided dependencies.
- Update the custom-test sample so it no longer inherits dependencies from the default `test` configurations.

## Validation

- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-gradle-plugin:functionalTest --tests "org.graalvm.buildtools.gradle.JavaApplicationWithTestsFunctionalTest.custom test image classpath follows custom source set"`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-gradle-plugin:functionalTest --tests "org.graalvm.buildtools.gradle.JavaApplicationWithTestsFunctionalTest.can register a custom test image"`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-gradle-plugin:test :native-gradle-plugin:inspections`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-gradle-plugin:configCacheFunctionalTest --tests "org.graalvm.buildtools.gradle.JavaApplicationWithTestsFunctionalTest.custom test image classpath follows custom source set"`
- `git diff --check`
- `grund gradle/FS-gradle-native-tests.1.1 --brief`

### DIFF
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy
@@ -46,7 +46,7 @@ import org.graalvm.buildtools.gradle.fixtures.TestResults
 import spock.lang.Issue
 import spock.lang.Unroll
 
-// Protects Gradle native-test task wiring and execution. §FS-native-tests.1 §FS-native-tests.2.
+// Protects Gradle native-test task wiring and execution. §FS-native-tests.1 §FS-native-tests.1.1 §FS-native-tests.2.
 class JavaApplicationWithTestsFunctionalTest extends AbstractFunctionalTest {
     @Unroll("can execute tests in a native image with JUnit Platform #junitVersion")
     def "can build a native image and run it"() {
@@ -197,6 +197,25 @@ class JavaApplicationWithTestsFunctionalTest extends AbstractFunctionalTest {
             skipped ':nativeTestCompile', ':nativeTest'
             doesNotContain ':build'
         }
+
+        where:
+        junitVersion = System.getProperty('versions.junit')
+    }
+
+    @Issue("https://github.com/graalvm/native-build-tools/issues/702")
+    @Unroll("custom test image classpath follows custom source set with JUnit Platform #junitVersion")
+    def "custom test image classpath follows custom source set"() {
+        given:
+        withSample("java-application-with-custom-tests")
+
+        when:
+        run 'dependencies', '--configuration', 'nativeImageIntegTestClasspath'
+
+        then:
+        outputContains "project :"
+        outputContains "org.junit.jupiter:junit-jupiter"
+        outputContains "org.junit.platform:junit-platform-launcher"
+        outputContains "org.graalvm.buildtools:junit-platform-native"
 
         where:
         junitVersion = System.getProperty('versions.junit')
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy
@@ -213,7 +213,6 @@ class JavaApplicationWithTestsFunctionalTest extends AbstractFunctionalTest {
         run 'dependencies', '--configuration', 'nativeImageIntegTestClasspath'
 
         then:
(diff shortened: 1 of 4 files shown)


vjovanov (dismissed):

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/JavaApplicationWithTestsFunctionalTest.groovy
```diff
@@ -46,7 +46,7 @@ import org.graalvm.buildtools.gradle.fixtures.TestResults
 import spock.lang.Issue
 import spock.lang.Unroll
 
-// Protects Gradle native-test task wiring and execution. §FS-native-tests.1 §FS-native-tests.2.
+// Protects Gradle native-test task wiring and execution. §FS-native-tests.1 §FS-native-tests.1.1 §FS-native-tests.2.
```
vjovanov: Shouldn't this be above each test individually?

jvukicev: Yes, I've moved the PRs citation directly above the test that exercises the feature.

vjovanov (dismissed):

vjovanov (approved):

PR #928, opened 2026-06-16
Make build infrastructure runnable on newer JDKs

### LINKED ISSUE #927: Make repository build infrastructure runnable on newer JDKs
## Problem

The repository build currently assumes an older Gradle runtime stack, so normal maintainer commands are not reliably runnable when `JAVA_HOME` points at newer JDK lines such as JDK 21 or JDK 25.

This is separate from the plugin runtime floor and from the functional-test matrix for older Gradle versions. CI can continue to run on JDK 17, but local repository build infrastructure should also configure and run on newer JDKs supported by the checked-in Gradle wrapper.

## Scope

- Keep Java artifacts/toolchains compatible with the supported Java floor.
- Update build infrastructure dependencies that block running the repository on newer JDKs.
- Keep the existing CI JDK matrix unchanged.
- Validate normal Gradle/Maven plugin sanity tasks on newer JDKs.

jormundur00:
## Summary

Fixes #927.

This updates repository build infrastructure so normal maintainer Gradle commands can run when `JAVA_HOME` points at newer JDK lines such as JDK 21 or JDK 25, while preserving the Java 17 toolchain target for produced artifacts and leaving the existing CI JDK matrix unchanged.

Changes include:

- bumping all checked-in Gradle wrappers to Gradle 9.6.1, matching the current tested Gradle line
- updating build plugins that were incompatible with Gradle 9
- moving the Gradle Plugin Publish metadata to the current DSL shape
- updating Spock/JUnit Platform test runtime dependencies for Gradle 9
- fixing small Kotlin/Groovy compatibility issues in build logic and tests
- fixing a latent nested-closure no-op in the resource-jar test fixture so requested extra resources are actually written and the native-image resource-config test exercises its intended path
- fixing two existing `common-utils` Checkstyle violations exposed by the Gradle upgrade

The final item belongs in this PR because Checkstyle is not version-pinned: Gradle 8.3 used Checkstyle 9.3, while the updated Gradle wrapper uses a newer Checkstyle version that reports these pre-existing violations. Fixing them keeps the common-utils CI gate green with the upgraded wrapper.

## Testing

- `./gradlew :native-gradle-plugin:test --tests org.graalvm.buildtools.gradle.tasks.GenerateResourcesConfigFileTest --no-daemon` on Gradle 9.6.1 — 9 tests passed
- `./gradlew :docs:asciidoctor --no-daemon` on Gradle 9.6.1 — documentation rendered successfully
- `grund check`
- `./gradlew :utils:checkstyleMain :utils:checkstyleTest` on Gradle 9.5.1 before aligning the wrapper with the tested 9.6.1 line
- Earlier local validation also covered the same sanity path on JDK 17 and JDK 21 before moving the patch onto the direct repo branch.

### DIFF
--- a/build-logic/documentation-plugins/build.gradle.kts
+++ b/build-logic/documentation-plugins/build.gradle.kts
@@ -49,6 +49,6 @@ repositories {
 }
 
 dependencies {
-    implementation("org.asciidoctor:asciidoctor-gradle-jvm:3.3.2")
+    implementation("org.asciidoctor.jvm.convert:org.asciidoctor.jvm.convert.gradle.plugin:4.0.5")
     implementation("org.ajoberstar:gradle-git-publish:3.0.0")
 }
--- a/docs/spec/requirements.md
+++ b/docs/spec/requirements.md
@@ -42,7 +42,9 @@ The plugins require Java 17+ to run. Native Image workflows must support JDK 25+
 against the two most recent GraalVM LTS versions. The Gradle plugin
 supports Gradle 8.3+ and currently tests `current`, `8.4`, `8.14.2`, `9.0.0`, and `9.6.1`. The Maven plugin
 supports Maven 3.9.x and currently tests with `3.9.9`. Metadata defaults must track the latest
-compatible published repository while remaining pinnable by version, URI, or local path.
+compatible published repository while remaining pinnable by version, URI, or local path. Repository
+build infrastructure must run on the supported Java floor and remain runnable on newer JDK lines
+supported by the checked-in Gradle wrapper.
 
 ## 2. Changing support
 
--- a/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/AbstractPluginTest.groovy
+++ b/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/AbstractPluginTest.groovy
@@ -85,29 +85,21 @@ abstract class AbstractPluginTest extends Specification {
 
     protected File newResourcesDirectory() {
         def dir = testDirectory.resolve("exploded${resourceCount++}").toFile()
-        new FileTreeBuilder(dir) {
-            'META-INF' {
-                'INDEX.LIST'('dummy')
-            }
-            'org' {
-                'foo' {
-                    'some' {
-                        'resource.txt'('resource text')
-                    }
-                }
-            }
-        }
+        new File(dir, "META-INF").mkdirs()
+        new File(dir, "META-INF/INDEX.LIST").text = "dummy"
+        new File(dir, "org/foo/some").mkdirs()
+        new File(dir, "org/foo/some/resource.txt").text = "resource text"
         dir
     }
 
     protected File newResourcesJar(Map<String, String> extraResources = [:]) {
         File resourcesDirectory = newResourcesDirectory()
-        extraResources.each { path, contents -> {
+        extraResources.each { path, contents ->
             def resource = new File(resourcesDirectory, path)
             if (resource.getParentFile().directory || resource.getParentFile().mkdirs()) {
                 resource << contents
             }
-        }}
+        }
         File jar = new File(testDirectory.toFile(), "resources-${resourceCount}.jar")
         Path resourcesPath = resourcesDirectory.toPath()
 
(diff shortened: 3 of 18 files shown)


vjovanov (dismissed):

native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/AbstractPluginTest.groovy
```diff
@@ -85,29 +85,21 @@ abstract class AbstractPluginTest extends Specification {
...
+        new File(dir, "META-INF/INDEX.LIST").text = "dummy"
+        new File(dir, "org/foo/some").mkdirs()
+        new File(dir, "org/foo/some/resource.txt").text = "resource text"
         dir
     }
 
     protected File newResourcesJar(Map<String, String> extraResources = [:]) {
         File resourcesDirectory = newResourcesDirectory()
-        extraResources.each { path, contents -> {
+        extraResources.each { path, contents ->
```
vjovanov, line 97: This is more than a Groovy-compat cleanup — it fixes a latent no-op. In the original `extraResources.each { path, contents -> { ... } }`, the inner `{ ... }` was a closure literal that was created and returned but **never invoked**, so `extraResources` were silently never written to disk. Removing the braces means they are now actually materialized.

That is the correct fix, but it is a behavior change: any functional test that passes `extraResources` now gets files that previously never existed. Could you confirm the dependent tests still assert what they intend, and mention this fix in the PR description so it is not mistaken for a pure style change?

build-logic/documentation-plugins/build.gradle.kts
```diff
@@ -49,6 +49,6 @@ repositories {
 }
 
 dependencies {
-    implementation("org.asciidoctor:asciidoctor-gradle-jvm:3.3.2")
+    implementation("org.asciidoctor.jvm.convert:org.asciidoctor.jvm.convert.gradle.plugin:4.0.5")
```
vjovanov, line 52: This is a major-version jump (asciidoctor-gradle 3.3.2 → 4.0.5, switching from the library coordinates to the `org.asciidoctor.jvm.convert` plugin-marker). The 3.x→4.x line had breaking changes in the docs pipeline. The testing checklist covers checkstyle and `grund check` but not a docs render — could you confirm the `:docs` build still succeeds on this wrapper before merge?

docs/spec/requirements.md
```diff
@@ -42,7 +42,9 @@ The plugins require Java 17+ to run. Native Image workflows must support JDK 25+
 against the two most recent GraalVM LTS versions. The Gradle plugin
 supports Gradle 8.3+ and currently tests `current`, `8.4`, `8.14.2`, `9.0.0`, and `9.6.1`. The Maven plugin
```
vjovanov, line 43: Minor consistency nit: this line says the Gradle plugin tests `…, 9.0.0, and 9.6.1`, but every checked-in wrapper in this PR is pinned to **9.5.1**. These are different axes (tested targets vs. the build wrapper), so not a bug — but bumping the wrapper to 9.6.1, or adding a one-line note on why they differ, would avoid confusion for the next maintainer.

native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/AbstractPluginTest.groovy
jvukicev: Confirmed. This helper has one caller that supplies extra resources: the test "jars which have a native-image resource config file do not contribute to detected resources". Its neighboring jar-detection test establishes that a normal resource jar contributes the resource, while this test expects an empty include list when META-INF/native-image/resource-config.json is present. I reran the full GenerateResourcesConfigFileTest under Gradle 9.6.1: all 9 tests passed, including those two cases. I also updated the PR description to call out the latent nested-closure no-op explicitly.

build-logic/documentation-plugins/build.gradle.kts
jvukicev: Confirmed. ./gradlew :docs:asciidoctor --no-daemon succeeds under the updated Gradle 9.6.1 wrapper and generates the expected HTML and asset tree under docs/build/docs/asciidoc. I added this validation to the PR description. The render reports only the existing non-fatal AsciiDoc content warnings (invalid references/heading sequence/unterminated admonition), not an Asciidoctor 4 compatibility failure.

docs/spec/requirements.md
jvukicev: Aligned these versions in dff06a88: all five checked-in wrappers now use Gradle 9.6.1, matching the declared and CI-tested current line. The focused resource test, docs render, and grund check all pass with the 9.6.1 root wrapper.

vjovanov (approved):

PR #931, opened 2026-06-16
Clarify buildArgs option entries

### LINKED ISSUE #930: Clarify that buildArgs entries are native-image argument tokens
The Maven and Gradle plugin docs should clarify how `buildArgs` entries map to Native Image options.

The important rule is that each configured entry should contain one Native Image option expression. The plugin configuration is not a shell command line, and users should not combine several independent Native Image options into one entry.

Some Native Image options include a value in the same expression, such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`; those should stay in the documented Native Image form. The confusing case is a single entry that combines unrelated options, such as `--static --libc=musl -H:+ReportExceptionStackTraces`.

For example, Maven docs should prefer this shape:

```xml
<buildArgs>
    <buildArg>--static</buildArg>
    <buildArg>--libc=musl</buildArg>
    <buildArg>--emit build-report</buildArg>
    <buildArg>-H:+ReportExceptionStackTraces</buildArg>
</buildArgs>
```

and Gradle docs should show the equivalent:

```kotlin
buildArgs.addAll(
    "--static",
    "--libc=musl",
    "--emit build-report",
    "-H:+ReportExceptionStackTraces"
)
```

This would prevent confusion from examples or user configurations that treat `buildArgs` as a shell-style string, such as putting `--static --libc=musl` into one Maven `<buildArg>` entry.

Suggested docs locations:

- `docs/src/docs/asciidoc/maven-plugin.adoc`, near the `<buildArgs>` section.
- `docs/src/docs/asciidoc/gradle-plugin.adoc`, near the `buildArgs.add(...)` description.

jormundur00:
Fixes #930

## What changed

This PR clarifies the docs for Maven and Gradle `buildArgs` entries.

Before this change, users could read the docs and assume a `buildArgs` entry should be written like a shell command line instead of as one native-image option expression per entry.

After this change, the docs make it clear that each Maven `<buildArg>` or Gradle `buildArgs` entry should contain a single native-image option expression, and the examples are updated accordingly.

## Why

This avoids configuration confusion and makes the examples line up with how Native Build Tools actually passes native-image arguments.

## Example

Before:

Users could infer that unrelated options should be combined into one `buildArgs` entry.

After:

The docs show separate entries for unrelated options while preserving valid single-expression forms such as `--emit build-report`, `--enable-monitoring=jfr`, and `-H:Name=value`.

## Implementation summary

- Clarify that each Maven `<buildArg>` entry should contain one native-image option expression.
- Add equivalent Gradle `buildArgs.addAll(...)` guidance.
- Update examples so unrelated options are configured as separate entries while keeping valid same-expression native-image forms.

## Validation

- `git diff --check`
- `env JAVA_HOME=/home/jovan/.mx/jdks/labsjdk-ee-17-jvmci-23.1-b02 PATH=/home/jovan/.mx/jdks/labsjdk-ee-17-jvmci-23.1-b02/bin:$PATH ./gradlew :docs:asciidoctor`
- Confirmed rendered Maven and Gradle HTML contains the new `buildArgs` guidance and examples.

### DIFF
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -173,6 +173,20 @@ You can also pass **build-time** and **run-time** arguments:
 - `buildArgs.add('<buildArg>')`: Configures the build by passing options directly to `native-image`. You can pass any Native Image build option listed https://www.graalvm.org/reference-manual/native-image/overview/Options/[here].
 - `runtimeArgs.add('<runtimeArg>')`: Specifies runtime arguments consumed by your application.
 
+Each `buildArgs` list entry should contain one Native Image option expression, not a shell command line.
+Put unrelated options in separate entries, but keep values that belong to the option expression together,
+such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
+
+[source,kotlin, role="multi-language-sample"]
+----
+buildArgs.addAll(
+    "--static",
+    "--libc=musl",
+    "--emit build-report",
+    "-H:+ReportExceptionStackTraces"
+)
+----
+
 ==== Command Line Options
 
 The preferred way to configure native binaries is through the DSL, so that you can provide reproducible builds which do not depend on obscure CLI invocations.
--- a/docs/src/docs/asciidoc/gradle-plugin.adoc
+++ b/docs/src/docs/asciidoc/gradle-plugin.adoc
@@ -175,15 +175,13 @@ You can also pass **build-time** and **run-time** arguments:
 
 Each `buildArgs` list entry should contain one Native Image option expression, not a shell command line.
 Put unrelated options in separate entries, but keep values that belong to the option expression together,
-such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
+such as `--emit build-report` or `--enable-monitoring=jfr`.
 
 [source,kotlin, role="multi-language-sample"]
 ----
 buildArgs.addAll(
-    "--static",
-    "--libc=musl",
     "--emit build-report",
-    "-H:+ReportExceptionStackTraces"
+    "--enable-monitoring=jfr"
 )
 ----
 
--- a/docs/src/docs/asciidoc/maven-plugin.adoc
+++ b/docs/src/docs/asciidoc/maven-plugin.adoc
@@ -43,10 +43,16 @@ The following configuration options are available:
    image name is not supplied, the artifact ID of the project will be used by default.
 `<buildArgs>`::
    If you want to pass additional build options to `native-image`, use the `<buildArgs>` tag. You can pass any Native Image build option listed https://www.graalvm.org/reference-manual/native-image/overview/Options/[here].
+   Each `<buildArg>` entry should contain one Native Image option expression, not a shell command line.
+   Put unrelated options in separate `<buildArg>` entries, but keep values that belong to the option expression together,
+   such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
 [source,xml, role="multi-language-sample"]
 ----
 <buildArgs>
-    <buildArg>--argument</buildArg>
+    <buildArg>--static</buildArg>
+    <buildArg>--libc=musl</buildArg>
+    <buildArg>--emit build-report</buildArg>
+    <buildArg>-H:+ReportExceptionStackTraces</buildArg>
 </buildArgs>
 ----
 `<skipNativeBuild>`::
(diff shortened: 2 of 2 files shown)


docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -173,6 +173,20 @@ You can also pass **build-time** and **run-time** arguments:
...
+Put unrelated options in separate entries, but keep values that belong to the option expression together,
+such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
+
+[source,kotlin, role="multi-language-sample"]
+----
+buildArgs.addAll(
+    "--static",
+    "--libc=musl",
+    "--emit build-report",
+    "-H:+ReportExceptionStackTraces"
```
vjovanov: I would omit the non-public flags.

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -43,10 +43,16 @@ The following configuration options are available:
...
+   Put unrelated options in separate `<buildArg>` entries, but keep values that belong to the option expression together,
+   such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
 [source,xml, role="multi-language-sample"]
 ----
 <buildArgs>
-    <buildArg>--argument</buildArg>
+    <buildArg>--static</buildArg>
+    <buildArg>--libc=musl</buildArg>
+    <buildArg>--emit build-report</buildArg>
+    <buildArg>-H:+ReportExceptionStackTraces</buildArg>
```
vjovanov: Same here.

docs/src/docs/asciidoc/maven-plugin.adoc
```diff
@@ -43,10 +43,16 @@ The following configuration options are available:
    image name is not supplied, the artifact ID of the project will be used by default.
 `<buildArgs>`::
    If you want to pass additional build options to `native-image`, use the `<buildArgs>` tag. You can pass any Native Image build option listed https://www.graalvm.org/reference-manual/native-image/overview/Options/[here].
+   Each `<buildArg>` entry should contain one Native Image option expression, not a shell command line.
+   Put unrelated options in separate `<buildArg>` entries, but keep values that belong to the option expression together,
+   such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
```
vjovanov: And here.

docs/src/docs/asciidoc/gradle-plugin.adoc
```diff
@@ -173,6 +173,20 @@ You can also pass **build-time** and **run-time** arguments:
 - `buildArgs.add('<buildArg>')`: Configures the build by passing options directly to `native-image`. You can pass any Native Image build option listed https://www.graalvm.org/reference-manual/native-image/overview/Options/[here].
 - `runtimeArgs.add('<runtimeArg>')`: Specifies runtime arguments consumed by your application.
 
+Each `buildArgs` list entry should contain one Native Image option expression, not a shell command line.
+Put unrelated options in separate entries, but keep values that belong to the option expression together,
+such as `--emit build-report`, `--enable-monitoring=jfr`, or `-H:Name=value`.
```
vjovanov: Here as well.

vjovanov (changes requested):

docs/src/docs/asciidoc/gradle-plugin.adoc
jvukicev: Updated in 0e5cf6e7: I removed `--static`, `--libc=musl`, and `-H:+ReportExceptionStackTraces`. I kept `--emit build-report` and added `--enable-monitoring=jfr` because both are public options already used and explained in this project’s end-to-end documentation; they also demonstrate separate option expressions.

docs/src/docs/asciidoc/maven-plugin.adoc
jvukicev: Updated in 0e5cf6e7: I removed the non-public flags. I kept `--emit build-report` and added `--enable-monitoring=jfr` because both are public options already used and explained in this project’s end-to-end documentation; they also demonstrate separate option expressions.

docs/src/docs/asciidoc/maven-plugin.adoc
jvukicev: Updated in 0e5cf6e7: removed the `-H:Name=value` form from the prose. The examples now mention only the public `--emit build-report` and `--enable-monitoring=jfr` forms.

docs/src/docs/asciidoc/gradle-plugin.adoc
jvukicev: Updated in 0e5cf6e7: removed the `-H:Name=value` form from the prose. The examples now mention only the public `--emit build-report` and `--enable-monitoring=jfr` forms.

vjovanov (approved):

PR #936, opened 2026-06-17
Log selected Maven metadata repository

### LINKED ISSUE #444: Log the Maven Repository Metadata version used by the plugin
We've seen a number of reports in Spring Boot where users don't really understand they have to do something to use the latest reachability metadata version. It looks like that one bit that could help is by logging the  version that the plugin is using. I've looked at a regular Maven project and couldn't find that information in the logs.

Making sure users are aware which version is running so that if the issue report they're looking at mention a different version, they know they have to do something.

A log at info level that provides the version of the reachability metadata.

I haven't.

**Additional context**
None.

jormundur00:
Fixes #444

## What changed

This PR makes the Maven plugin log which reachability metadata repository version or source it selected.

Before this change, metadata repository selection happened silently in normal Maven output. Users could see the effects of metadata selection, but not which repository version or source was actually in use.

After this change, the Maven plugin prints one concise info-level line describing the selected official metadata repository version, or the configured source when a semantic version is not available.

## Why

Metadata repository selection is user-visible integration state. Surfacing it in normal Maven output makes builds easier to understand and troubleshoot.

## Example

Before:

A Maven build could resolve metadata from the default repository, a forced version, a local path, or a custom URL without telling the user which repository source was actually chosen.

After:

The build emits output such as `Using GraalVM reachability metadata repository version <version>` or `Using GraalVM reachability metadata repository from <source>`.

## Implementation summary

- Add one concise info-level Maven log line after metadata repository selection and before dependency metadata lookup.
- Report official repository versions when they are known.
- Report local paths, ZIP paths, and arbitrary URLs as repository sources when a semantic version is not known.
- Keep disabled metadata quiet with respect to the selected-repository line.
- Extend Maven functional coverage across default, explicit, implicit-enable, path, ZIP, URL, forced-version, and disabled metadata cases.

## Validation

- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal ./gradlew :native-maven-plugin:test --no-daemon`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal GRAALVM_HOME=/home/jovan/ol/bb/master/graal/sdk/mxbuild/linux-amd64/GRAALVM_JAVA25/graalvm-25.1.0-dev+10.1 ./gradlew :native-maven-plugin:functionalTest --no-daemon --fail-fast --tests org.graalvm.buildtools.maven.MetadataRepositoryFunctionalTest`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal GRAALVM_HOME=/home/jovan/ol/bb/master/graal/sdk/mxbuild/linux-amd64/GRAALVM_JAVA25/graalvm-25.1.0-dev+10.1 ./gradlew :native-maven-plugin:functionalTest --no-daemon --fail-fast --tests org.graalvm.buildtools.maven.OfficialMetadataRepositoryFunctionalTest`
- `git diff --check`
- `grund maven/FS-maven-resources-and-metadata.2`

### DIFF
--- a/native-gradle-plugin/docs/functional/resources-and-metadata.md
+++ b/native-gradle-plugin/docs/functional/resources-and-metadata.md
@@ -27,6 +27,10 @@ metadata repository URI, version, exclusions, and module-to-config-version overr
 directory must be consumable by native compile tasks and must contain only metadata selected for
 the binary's dependency graph. When no URI or version is pinned, the Gradle default should follow
 the repository-wide freshness goal in [§root/GOAL-fresh-metadata](../../../docs/spec/goals.md#goal-fresh-metadata-users-can-fetch-the-latest-graalvm-reachability-metadata).
+Normal Gradle output must identify the selected metadata repository version
+or, when a version is not known for a local path or arbitrary URL, the selected repository source.
+This keeps important integration state visible without flooding build logs, as required by
+[§root/GOAL-concise-actionable-output](../../../docs/spec/goals.md#goal-concise-actionable-output-build-output-is-concise-actionable-and-token-efficient).
 
 ## 4. Missing metadata reports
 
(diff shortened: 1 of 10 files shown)


vjovanov:
Do we have the same for Gradle? We should output the same text.

jvukicev:
@vjovanov Good point — added this for Gradle in db3ef79e. Gradle now emits the same `Using GraalVM reachability metadata repository version <version>` or `... from <source>` message, with coverage for official repositories, custom paths/archives, and disabled metadata.

vjovanov (approved):

PR #942, opened 2026-06-18
Fix Gradle metadata repository service scoping

### LINKED ISSUE #415: Native gradle plugin - Metadata repository URL can be set only in the first project being evaluated
In a multi module project, the metadata repository URL can be set only in the build.gradle file of the first module being evaluated. This happens because `GraalVMReachabilityMetadataService` is a shared service and created just for the first evaluated module. The metadata repository URL from any other module is ignored.

**System Info:**
 - OS: Mac
 - GraalVM Version: 22.3.0 CE
 - Java Version: 17
 - Plugin version: native-gradle-plugin:0.9.20

jormundur00:
Fixes #415

## What changed

This PR fixes multi-module Gradle builds where different subprojects configure different reachability metadata repositories.

Before this change, Native Build Tools could reuse repository settings from the first subproject that initialized the shared metadata service. That meant later subprojects could silently use the wrong metadata repository.

After this change, each subproject uses its own configured metadata repository.

## Why

Multi-project builds can now safely use different metadata repositories per subproject without cross-project leakage.

## Example

Before:

```gradle
project(":app1") {
    graalvmNative {
        metadataRepository {
            uri = uri("file:///tmp/metadata-repo-a")
        }
    }
}

project(":app2") {
    graalvmNative {
        metadataRepository {
            uri = uri("file:///tmp/metadata-repo-b")
        }
    }
}
```

In some builds, `app2` could still end up using the repository configuration from `app1`.

After:

`app1` uses `metadata-repo-a` and `app2` uses `metadata-repo-b`, matching each subproject's own configuration.

## Implementation summary

- Scope the shared Gradle metadata service per subproject so one project's repository settings do not leak into another.
- Keep service reuse within a single project while allowing different subprojects to carry different metadata repository parameters.
- Add a multi-project regression that verifies each subproject resolves metadata from its own configured repository.

## Validation

Run from the issue worktree with `JAVA_HOME` and `GRAALVM_HOME` set to the installed GraalVM 17 runtime:

- `./gradlew :native-gradle-plugin:functionalTest --tests "org.graalvm.buildtools.gradle.NativeConfigRepoFunctionalTest.subprojects can use different metadata repositories"` passed.
- `./gradlew :native-gradle-plugin:test :native-gradle-plugin:inspections` passed.
- `./gradlew :native-gradle-plugin:configCacheFunctionalTest --tests "org.graalvm.buildtools.gradle.NativeConfigRepoFunctionalTest.subprojects can use different metadata repositories"` passed.
- `grund fmt . --marker --check` passed.
- Targeted grounding checks resolved and found the changed `FS-gradle-resources-and-metadata.3` citations.

### DIFF
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeConfigRepoFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeConfigRepoFunctionalTest.groovy
@@ -151,7 +151,7 @@ graalvmNative {
         outputContains "[graalvm reachability metadata repository for org.graalvm.internal:library-with-reflection:1.5]: Configuration is forced to version 2"
     }
 
-    // §FS-resources-and-metadata.3
+    // Multi-project metadata collection must honor each project's configured repository URI. §FS-resources-and-metadata.3.
     def "subprojects can use different metadata repositories"() {
         given:
         settingsFile << """
--- a/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeConfigRepoFunctionalTest.groovy
+++ b/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeConfigRepoFunctionalTest.groovy
@@ -42,8 +42,12 @@
 package org.graalvm.buildtools.gradle
 
 import org.graalvm.buildtools.gradle.fixtures.AbstractFunctionalTest
+import org.graalvm.buildtools.gradle.fixtures.FileUtils
 import org.gradle.api.logging.LogLevel
 import spock.lang.Unroll
+
+import java.nio.file.Files
+import java.nio.file.Path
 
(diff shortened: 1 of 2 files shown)


vjovanov (approved):

native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/NativeConfigRepoFunctionalTest.groovy
```diff
@@ -147,4 +151,80 @@ graalvmNative {
         outputContains "[graalvm reachability metadata repository for org.graalvm.internal:library-with-reflection:1.5]: Configuration is forced to version 2"
     }
 
+    def "subprojects can use different metadata repositories"() {
```
vjovanov, line 155: Tests should always cite the spec they are testing.

jvukicev: It did cite it, just inside of the test. I've moved the citation to cover the whole test.

PR #958, opened 2026-07-06
Fix layered image configuration cache test

### LINKED ISSUE #957: Layered image functional test is not configuration-cache compatible
## Summary

`LayeredApplicationFunctionalTest` currently fails when run through the Gradle configuration-cache functional-test suite. It is temporarily skipped there while the Gradle layer model is corrected.

## Current behavior

The test builds `libdependencies` as a dependency layer, then invokes the application layer and expects `:nativeLibdependenciesCompile` to remain `UP-TO-DATE`. With configuration cache enabled, the TestKit fixture creates its exploded repository after the first cache snapshot, preventing reuse on the next nested build. The layer binary is also wired through the normal image classpath, which includes the application artifact; an application-source change therefore invalidates the dependency layer.

## Expected behavior

Layered images should support Gradle configuration cache. The dependency-layer task should use only its declared layer inputs, remain up-to-date for application-only changes, and preserve the main/use-layer binary classpath.

## Proposed direction

- Make the TestKit repository available before the first cached nested build.
- Model create-layer inputs explicitly rather than inheriting the application image classpath.
- Add focused task-input/configuration-cache coverage, then remove the temporary `@IgnoreIf`.

## Evidence

- https://github.com/graalvm/native-build-tools/actions/runs/28664929208/job/85014613567
- https://github.com/graalvm/native-build-tools/actions/runs/28666165975/job/85018703603

jormundur00:
## What changed

Layered native-image builds in the Gradle plugin now work in the configuration-cache functional-test path. A dependency layer declared with `createLayer { jars.from(...) }` uses its declared layer JARs as the Native Image invocation classpath, keeping the dependency-layer command line independent of application artifacts.

The binary classpath is no longer cleared when configuring a layer. This preserves package-based layer creation: `createLayer { packages = [...] }` retains the binary classpath that supplies the selected package classes.

The layered application functional test now runs with configuration cache enabled on Linux and verifies cache storage and reuse. After an application-only source change, the dependency layer stays up to date while the application layer rebuilds.

## Why

The earlier implementation made the dependency layer inherit an application classpath, invalidating it for application-only changes. Clearing that classpath fixed the JAR-layer scenario but unintentionally broke package-based layer declarations and made the DSL order-dependent.

The command-line provider now chooses the declared layer JARs only for JAR-based layers, while package-based layers retain their configured binary classpath.

## Example

```groovy
graalvmNative {
    binaries {
        libdependencies {
            createLayer {
                jars.from(configurations.runtimeClasspath)
            }
        }
        libpackages {
            classpath.from(layout.projectDirectory.file("packages.jar"))
            createLayer {
                packages = ["com.example.shared"]
            }
        }
    }
}
```

`nativeLibdependenciesCompile` uses the declared layer JARs for both `-H:LayerCreate` and `-cp`; `nativeLibpackagesCompile` retains `packages.jar` as `-cp`.

## Implementation summary

- Select the declared `CreateLayerOptions.jars` as `-cp` in `NativeImageCommandLineProvider`, without mutating `NativeImageOptions.classpath`.
- Add unit coverage for both JAR-based and package-based layer creation.
- Specify the layer classpath rules in `FS-native-invocation.3`.
- Keep the configuration-cache fixture changes that separately assert cache-store output and cache reuse.

Fixes #957.

## Validation

Passed locally with Gradle running on JDK 17 and Native Image from GraalVM 25:

```text
grund check
git diff --check
JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal \
  GRAALVM_HOME=/home/jovan/.sdkman/candidates/java/current \
  ./gradlew :native-gradle-plugin:test \
  --tests org.graalvm.buildtools.gradle.tasks.NativeImageCommandLineProviderTest

JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal \
  GRAALVM_HOME=/home/jovan/.sdkman/candidates/java/current \
  ./gradlew :native-gradle-plugin:configCacheFunctionalTest \
  --tests org.graalvm.buildtools.gradle.LayeredApplicationFunctionalTest
```

The focused configuration-cache test passed with zero failures and verified cache reuse, `LayerCreate`, `LayerUse`, and that the dependency layer remains up to date after an application-only source change.

### DIFF
--- a/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
+++ b/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
@@ -112,6 +112,7 @@ class NativeImageCommandLineProviderTest extends AbstractPluginTest {
         args.any { it.startsWith("${NativeImageFlags.LAYER_CREATE}=libbase.nil") && it.contains("module=java.base") }
     }
 
+    // Uses declared layer JARs instead of the inherited binary classpath. §FS-native-invocation.3.
     @Issue("https://github.com/graalvm/native-build-tools/issues/957")
     def "uses layer jars instead of inherited binary classpath for layer-create build"() {
         given:
--- a/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
+++ b/native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
@@ -144,10 +144,45 @@ class NativeImageCommandLineProviderTest extends AbstractPluginTest {
         ).asArguments()
 
         then:
-        !options.classpath.files.contains(appJar)
+        options.classpath.files.contains(appJar)
         !args.contains(appJar.absolutePath)
         args.any { it.startsWith("${NativeImageFlags.LAYER_CREATE}=libdependencies.nil") && it.contains("path=${dependencyJar}") }
         args.contains("-cp")
         args.contains(dependencyJar.absolutePath)
+    }
+
+    def "retains binary classpath for package-based layer-create build"() {
+        given:
+        def project = newProject()
+        project.plugins.apply(ApplicationPlugin)
+        project.plugins.apply(NativeImagePlugin)
+        def appJar = testDirectory.resolve("app.jar").toFile()
+        appJar.text = "application"
+        def options = project.extensions.getByType(GraalVMExtension).binaries.create("libpackages")
+        options.classpath.from(appJar)
+        options.createLayer {
+            it.modules.add("java.base")
+            it.packages.add("org.graalvm.demo")
+        }
+        options.excludeConfigArgs.set([])
+        options.configurationFileDirectories.setFrom([])
+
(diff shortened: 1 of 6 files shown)


native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
```diff
@@ -111,4 +111,78 @@ class NativeImageCommandLineProviderTest extends AbstractPluginTest {
         args.contains(NativeImageFlags.UNLOCK_EXPERIMENTAL_VMOPTIONS)
         args.any { it.startsWith("${NativeImageFlags.LAYER_CREATE}=libbase.nil") && it.contains("module=java.base") }
     }
+
+    @Issue("https://github.com/graalvm/native-build-tools/issues/957")
```
vjovanov, line 116: Cite the spec?

vjovanov (approved):

native-gradle-plugin/src/test/groovy/org/graalvm/buildtools/gradle/tasks/NativeImageCommandLineProviderTest.groovy
jvukicev: Added citation.

PR #926, opened 2026-06-16
Fix generateTestResourceConfig autodetection paths

jormundur00:
Fixes #566

## What changed

This PR fixes Maven `native:generateTestResourceConfig` autodetection so current-project test resources are discovered from processed test output rather than from the main output artifact or raw test resource source directories.

Before this change, current-project autodetection could scan the wrong paths for test resources.

After this change, current-project test resource autodetection uses `target/test-classes` while preserving the existing shared classpath scanning behavior for dependencies.

## Why

Users expect test resource config generation to reflect the processed test output Maven actually builds for the current project. Scanning the wrong paths can either miss resources or pick up the wrong inputs.

## Example

Before:

`native:generateTestResourceConfig` could treat raw `src/test/resources` directories or main output artifacts as the current-project source for autodetection.

After:

The same goal now scans the processed current-project test output under `target/test-classes`, which matches Maven's test resource pipeline.

## Implementation summary

- Add an overridable current-project artifact hook in `AbstractResourceConfigMojo`.
- Update `NativeBuildTestResourceConfigMojo` to use `mavenProject.getBuild().getTestOutputDirectory()` for current-project autodetection.
- Remove raw test resource source directories from the test resource config autodetection path.
- Clarify the Maven resource metadata functional spec.
- Update functional expectations so autodetection-enabled test resource generation follows processed test output resources.

## Validation

- `grund fmt . --marker --check`
- `git diff --check`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal PATH=/home/jovan/.sdkman/candidates/java/17.0.12-graal/bin:$PATH ./gradlew :native-maven-plugin:inspections --no-daemon`
- `JAVA_HOME=/home/jovan/.sdkman/candidates/java/17.0.12-graal PATH=/home/jovan/.sdkman/candidates/java/17.0.12-graal/bin:$PATH ./gradlew :native-maven-plugin:functionalTest --no-daemon --fail-fast --tests org.graalvm.buildtools.maven.JavaApplicationWithResourcesFunctionalTest`

PR #966, opened 2026-07-08
Fix Maven native test resource handling

### LINKED ISSUE #566: With autodetection enabled, `generateTestResourceConfig` should look at processed test resources only
What happens currently, given that `resources.autodetection.enabled` is set, is the following:

## `generateResourceConfig`

* ✅  Detects resources in `target/classes`.

```
[INFO] [INFO] --- native:0.9.28:generateResourceConfig (default-generateResourceConfig) @ package-native-image ---
[INFO] [INFO] Detected resources for /Users/alvaro/Dev/micronaut-projects/micronaut-maven-plugin/target/it/package-native-image/target/classes are [logback.xml, application.yml]
[INFO] [INFO] Resources configuration written into /Users/alvaro/Dev/micronaut-projects/micronaut-maven-plugin/target/it/package-native-image/target/native/generated/generateResourceConfig/resource-config.json
```

## `generateTestResourceConfig`

* :x: Detects resources in `target/classes`. It should be in `target/test-classes`.
* :x: Detects resources in `src/test/resources`. This shouldn't be attempted since a) those are unprocessed; and b) processed test resources will already be in `target/test-classes`.

```
[INFO] [INFO] --- native:0.9.28:generateTestResourceConfig (default-generateTestResourceConfig) @ package-native-image ---
[INFO] [INFO] Detected resources for /Users/alvaro/Dev/micronaut-projects/micronaut-maven-plugin/target/it/package-native-image/target/classes are [logback.xml, application.yml]
[INFO] [INFO] Detected resources for /Users/alvaro/Dev/micronaut-projects/micronaut-maven-plugin/target/it/package-native-image/src/test/resources are [application-test.yml]
[INFO] [INFO] Resources configuration written into /Users/alvaro/Dev/micronaut-projects/micronaut-maven-plugin/target/it/package-native-image/target/native/generated/generateTestResourceConfig/resource-config.json
```

alvarosanchez:
## Summary

Use Maven's processed build outputs consistently for native tests:

- stop adding raw `src/test/resources` directories to the `native:test` image classpath
- make `native:generateTestResourceConfig` scan both `project.build.outputDirectory` and `project.build.testOutputDirectory` for the current project
- preserve dependency artifact scanning
- document the Maven lifecycle semantics in the functional specs

This supersedes #926 and closes #566. Unlike #926, it also fixes the native-test runtime classpath, where both `target/test-classes` and `src/test/resources` were previously passed to `native-image`.

## Why

`target/classes` and `target/test-classes` are Maven's processed main and test outputs. They contain compiled classes plus resources after filtering, exclusions, custom output paths, and resource-generating executions. Adding raw test resource source directories as well:

- duplicates resources already copied to `target/test-classes`
- bypasses Maven filtering and exclusions
- omits resources generated directly into the test output
- can make duplicate-resource detection fail during native test image generation

If a build intentionally skips test-resource processing, the native plugin now respects that lifecycle choice rather than reintroducing raw resources behind Maven's back.

Main resources do not have the same current duplication path: the native test classpath already uses the processed main output or packaged artifact and does not add `src/main/resources`. Test-resource autodetection nevertheless scans both processed main and test outputs because both are part of the native-test runtime classpath.

## TDD / verification

The functional tests were changed before the follow-up source fix and failed against the previous implementation:

- dynamic main-resource lookup failed in the native test because `message.txt` was absent from generated resource configuration
- generated test resource configuration contained only processed test resources, omitting processed main resources

The implementation then made those tests pass. Integration coverage now verifies from generated build artifacts and the actual native-image argument file that:

- `target/classes` and `target/test-classes` are on the native-test classpath
- neither `src/main/resources` nor `src/test/resources` is on that classpath
- processed main resources required through a runtime-selected name are included and available in the native test image
- a raw test resource excluded by Maven is not detected
- a resource generated directly into `target/test-classes` is detected
- test resource autodetection scans both processed outputs
- resource-config checks are independent of JSON entry ordering

Local verification:

- `grund check`
- `./gradlew :native-maven-plugin:test :native-maven-plugin:checkstyleMain --no-daemon`
- `./gradlew :native-maven-plugin:functionalTest --no-daemon --tests 'org.graalvm.buildtools.maven.JavaApplicationWithResourcesFunctionalTest'`

### DIFF
--- a/native-maven-plugin/docs/functional/native-tests.md
+++ b/native-maven-plugin/docs/functional/native-tests.md
@@ -6,9 +6,11 @@ run those tests as a native image through Maven's test lifecycle.
 ## 1. Test classpath
 
 `native:test` must include compile, runtime, test, compile-plus-runtime, and provided-scope
-dependencies needed by the test image. It must add compiled test classes, test resources, plugin
-artifacts relevant to Native Build Tools and JUnit, and inferred `junit-platform-native`
-dependencies.
+dependencies needed by the test image. Current-project content must come from Maven's build
+outputs, including the processed test output directory; raw test resource source directories must
+not be added because doing so bypasses Maven resource filtering and exclusions and duplicates
+resources already present in the processed output. The goal must also add plugin artifacts relevant
+to Native Build Tools and JUnit and inferred `junit-platform-native` dependencies.
 
 ## 2. Test discovery preconditions
 
--- a/native-maven-plugin/docs/functional/resources-and-metadata.md
+++ b/native-maven-plugin/docs/functional/resources-and-metadata.md
@@ -11,9 +11,9 @@ respect existing Native Image resource configuration and the shared classpath sc
 [§common/FS-common-libraries.2](../../../common/docs/functional-spec.md#2-resource-configuration).
 
 For `native:generateTestResourceConfig`, autodetection for the current project must use Maven's
-processed test output directory instead of the main output artifact or raw test resource source
-directories. This preserves resource filtering, exclusions, and resources generated directly into
-the test output.
+processed main and test output directories instead of raw resource source directories. This scans
+the native-test runtime classpath while preserving resource filtering, exclusions, and resources
+generated directly into either output.
 
 ## 2. Reachability metadata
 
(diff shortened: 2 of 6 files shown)


jormundur00 (commented):
P2 — Direct goal invocation no longer detects test resources

`getProjectArtifacts()` now returns only `target/test-classes`. Invoking `mvn native:generateTestResourceConfig` directly does not execute `process-test-resources`, so that directory may not exist and autodetection produces an empty configuration. Previously, this goal scanned the declared raw test-resource directories. Please retain an appropriate fallback when the processed output is absent, or otherwise ensure the prerequisite lifecycle phase runs. The new functional test binds the goal to `test`, so it does not cover direct invocation.

jormundur00 (approved):
Looks good to me. The follow-up correctly scans the processed main and test outputs, and the updated tests cover filtered/generated resources plus the native-test classpath.

PR #981, opened 2026-07-20
Normalize platform-dependent resource and test paths

### LINKED ISSUE #973: Resource autodetection generates Windows-only path separators

### LINKED ISSUE #974: Functional tests fail on macOS path aliases

jormundur00:
## Problem

This PR addresses two cross-platform problems:

1. **Windows resource autodetection produced invalid resource names.** Classpath-directory scanning passed `Path.toString()` directly to resource-pattern generation, producing backslash-separated names such as `org\\graalvm\\demo\\expected.txt`. Native Image resource names use portable `/` separators, so nested resources could be missing from Windows native images.
2. **Functional tests assumed one textual representation of paths and arguments.** Equivalent temporary locations are reported differently across platforms—for example `/var/...` versus `/private/var/...` on macOS and `RUNNER~1` versus `runneradmin` on Windows. Windows also defaults to Native Image argument files, and checked-out Maven samples use CRLF. The builds succeeded, but exact string assertions failed.

### Portable resource discovery

- route directory resource names through the existing separator-normalization helper before generating resource patterns
- add regression coverage proving nested directory resources use names such as `org/graalvm/demo/expected.txt`

### Portable functional tests

- accept equivalent absolute and canonical temporary-path representations in Maven agent-output assertions
- construct metadata-repository expectations from equivalent file URI representations
- use normalized separators in path-bearing assertions
- explicitly use and inspect a Native Image `@args` file for the Gradle console-color scenario on every operating system
- inspect Gradle argument-file contents when validating custom binary metadata exclusions
- normalize CRLF before the Maven resource fixture matches and modifies its sample POM

## Testing

Validation performed during initial development:

- `grund check`
- `./gradlew :utils:test :utils:checkstyleMain :utils:checkstyleTest :native-gradle-plugin:compileFunctionalTestGroovy :native-maven-plugin:compileFunctionalTestGroovy`
- affected Gradle functional scenarios with GraalVM 25.0.3 (6 passed)
- affected Maven functional scenarios with GraalVM 25.0.3 (5 passed)

Cross-platform validation:

- exercised the generated Gradle and Maven functional-test matrices on `ubuntu-22.04`, `windows-latest`, and `macos-latest` using a temporary test-only commit
- fixed the platform-specific path, line-ending, and argument-file assertions exposed by that matrix
- all CI tests passed before removing the temporary test-only commit from this PR
- `grund check`
- `git diff --check`

Closes #973
Closes #974

### DIFF
--- a/common/utils/src/main/java/org/graalvm/buildtools/model/resources/ClassPathDirectoryAnalyzer.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/model/resources/ClassPathDirectoryAnalyzer.java
@@ -106,7 +106,8 @@ class ClassPathDirectoryAnalyzer extends ClassPathEntryAnalyzer {
                 hasNativeImageResourceFile = true;
                 return FileVisitResult.TERMINATE;
             }
-            maybeAddResource(root.relativize(file).toString(), resources);
+            // Native Image resource names use portable separators. §FS-common-libraries.2.
+            maybeAddResource(relativePathOf(file), resources);
             return FileVisitResult.CONTINUE;
         }
     }
--- a/common/utils/src/test/java/org/graalvm/buildtools/utils/ClassPathEntryAnalyzerTest.java
+++ b/common/utils/src/test/java/org/graalvm/buildtools/utils/ClassPathEntryAnalyzerTest.java
@@ -43,14 +43,20 @@ package org.graalvm.buildtools.utils;
 import org.graalvm.buildtools.model.resources.ClassPathEntryAnalyzer;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Collections;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class ClassPathEntryAnalyzerTest {
+    @TempDir
+    Path temporaryDirectory;
+
     @Test
     @DisplayName("A non existent jar shouldn't cause an exception")
     public void testShouldAllowNonExistentJar() throws IOException {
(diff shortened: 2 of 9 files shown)


build-logic/common-plugins/src/main/groovy/org.graalvm.build.github-actions-helper.gradle
```diff
@@ -7,7 +7,7 @@ def matrixDefault = [
 
 def latestMatrixDefault = [
         "java-version": [ "17.0.12" ],
-        "os": [ "ubuntu-22.04" ]
+        "os": [ "ubuntu-22.04", "windows-latest", "macos-latest" ]
```
jvukicev: This is here only for test, will be removed in merging commit.

vjovanov (approved):

PR #997, opened 2026-07-22
Deprecate fallback plugin options

### LINKED ISSUE #991: Deprecate the `fallback` option — Native Image fallback feature removed in GraalVM 25.1
## Summary

The `fallback` build option in both plugins maps to Native Image's `--no-fallback` flag. As of **GraalVM 25.1**, the fallback feature has been **removed**: `--no-fallback` is deprecated and no longer has any effect ([25.1 release notes](https://www.graalvm.org/release-notes/25.1/), "Deprecated Functionality"). We should deprecate the plugin option accordingly.

## Current behavior

Both plugins default `fallback = false`, which means `--no-fallback` is emitted on **every** build by default:

- **Gradle:** `getFallback().convention(false)` (`BaseNativeImageOptions.java:266`), translated via `.map(NEGATE)` → `--no-fallback` (`NativeImageCommandLineProvider.java:191`)
- **Maven:** `@Parameter(property = "fallback", defaultValue = "false")` (`AbstractNativeImageMojo.java:142`) → `if (!fallback) cliArgs.add("--no-fallback")` (`AbstractNativeImageMojo.java:229`)

On GraalVM 25.1+ this passes a no-op flag (and may produce a deprecation warning on every build).

## Proposed change

1. Add `@Deprecated` + javadoc to the Gradle `getFallback()` DSL option and the Maven `<fallback>` parameter, explaining that the Native Image fallback feature was removed in GraalVM 25.1 and the flag is now a no-op. Keep it functional for a release cycle for users on older GraalVM/JDK versions.
2. (Optional) Suppress emission of `--no-fallback` on 25.1+ to avoid the per-build deprecation warning. Note: the removal landed specifically in 25.1, but both 25.0 and 25.1 are JDK-25-based, so the existing `majorJDKVersion` check used for the color flags cannot distinguish them — gating this cleanly requires the actual GraalVM version, not the JDK major.

## Notes

- `--no-fallback` remains meaningful (and important, given the default-on behavior) for GraalVM versions **before** 25.1, so the option should be deprecated rather than removed outright.

jormundur00:
## What changed

The Gradle `fallback` DSL option and Maven `<fallback>` parameter are now marked deprecated. Both remain available for compatibility.

The plugins now distinguish the GraalVM release from the JDK version reported by `native-image --version`:

- Before GraalVM 25.1, disabling fallback continues to generate `--no-fallback`.
- On positively identified GraalVM 25.1+ releases, the plugins omit the generated `--no-fallback` flag.
- Runtime-environment lines are parsed independently of vendor labels, covering Oracle GraalVM, Mandrel, and BellSoft Liberica NIK formats.
- If the GraalVM release cannot be identified, the plugins conservatively retain the flag.
- Explicit user-supplied `--no-fallback` build arguments remain unchanged.

## Why

Native Image removed the fallback feature in GraalVM 25.1, so `--no-fallback` is deprecated and has no effect there. Suppressing the plugin-generated flag avoids a deprecation warning on every build while preserving behavior for older GraalVM releases.

Both GraalVM 25.0 and 25.1 use JDK 25, so the existing JDK-major check cannot distinguish them. The implementation reads the first dotted release from a runtime-environment line in the existing `native-image --version` output, regardless of the distribution's vendor label, and does not add another version-check process.

## Example

Existing configurations continue to work while showing the deprecation in API documentation:

```groovy
nativeImage {
    fallback = false // deprecated on GraalVM 25.1+
}
```

```xml
<configuration>
  <fallback>false</fallback> <!-- deprecated on GraalVM 25.1+ -->
</configuration>
```

## Implementation summary

- Added deprecation annotations and GraalVM 25.1 guidance to the Gradle and Maven option surfaces.
- Added shared, vendor-neutral GraalVM release detection independent of the JDK version.
- Suppressed plugin-generated `--no-fallback` on confirmed GraalVM 25.1+ releases in both plugins.
- Preserved compatibility behavior for older and unrecognized releases and preserved explicit build arguments.
- Updated the common, Gradle, and Maven functional specs.
- Added exact Oracle GraalVM, Mandrel, and BellSoft Liberica NIK parser fixtures, including same-format 25.1 cases.
- Added Gradle/Maven command-line regression tests.

## Validation

- `grund check`
- `grund fmt . --marker --cross-refs --check`
- `./gradlew :utils:check`
- Focused Gradle and Maven argument regression tests
- Maven invocation functional test against real Oracle GraalVM 25.0.3, Mandrel 25.0.3.0-Final, and BellSoft Liberica NIK 25.0.3-2 installations
- Verified real current outputs are identified as pre-25.1 and same-format 25.1 transformations are identified as 25.1+ for all three distributions
- `./gradlew :native-gradle-plugin:inspections :native-maven-plugin:inspections :utils:check`
- `git diff --check`
- A complete Maven unit run was attempted for the original deprecation change; it has one unrelated reproducible `NativeExtensionTest` failure because this environment lacks a Spock class-mocking backend. The changed Maven tests and module inspections passed.

Fixes #991

### DIFF
--- a/common/docs/functional-spec.md
+++ b/common/docs/functional-spec.md
@@ -27,9 +27,10 @@ required by [§REQ-no-buildtool-apis](requirements.md#req-no-buildtool-apis-comm
 Common utilities keep Gradle and Maven command-line handling consistent. They must preserve
 whitespace, quotes, backslashes, and platform paths when escaping arguments; write Native Image
 argument files and return the corresponding `@...` argument ([§root/GLOSS-argument-file](../../docs/spec/glossary.md#gloss-argument-file-native-image-argument-file)); distinguish the GraalVM release from the JDK version in Native Image version output well enough to
-choose version-specific behavior, retaining compatibility behavior when the GraalVM release cannot
-be identified; and centralize Native Image configuration file names and metadata directory names
-used by plugins and tests.
+choose version-specific behavior, using the first dotted release on a runtime-environment line
+regardless of its vendor label and retaining compatibility behavior when that release cannot be
+identified; and centralize Native Image configuration file names and metadata directory names used
+by plugins and tests.
 
 ## 2. Resource configuration
 
--- a/common/docs/functional-spec.md
+++ b/common/docs/functional-spec.md
@@ -26,9 +26,10 @@ required by [§REQ-no-buildtool-apis](requirements.md#req-no-buildtool-apis-comm
 
 Common utilities keep Gradle and Maven command-line handling consistent. They must preserve
 whitespace, quotes, backslashes, and platform paths when escaping arguments; write Native Image
-argument files and return the corresponding `@...` argument ([§root/GLOSS-argument-file](../../docs/spec/glossary.md#gloss-argument-file-native-image-argument-file)); parse Native
-Image and JDK versions well enough to choose version-specific behavior; and centralize Native
-Image configuration file names and metadata directory names used by plugins and tests.
+argument files and return the corresponding `@...` argument ([§root/GLOSS-argument-file](../../docs/spec/glossary.md#gloss-argument-file-native-image-argument-file)); distinguish the GraalVM release from the JDK version in Native Image version output well enough to
+choose version-specific behavior, retaining compatibility behavior when the GraalVM release cannot
+be identified; and centralize Native Image configuration file names and metadata directory names
+used by plugins and tests.
 
 ## 2. Resource configuration
 
--- a/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
+++ b/common/utils/src/main/java/org/graalvm/buildtools/utils/NativeImageUtils.java
@@ -65,8 +65,8 @@ public class NativeImageUtils {
     private static final Pattern javaVersionPattern = Pattern.compile("^native-image ([0-9]+).*", Pattern.DOTALL);
     private static final Pattern javaVersionLegacyPattern = Pattern.compile(".* \\(Java Version ([0-9]+)\\.([0-9]+)\\.([0-9]+).*");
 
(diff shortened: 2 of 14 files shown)


vjovanov (approved):

vjovanov:
I would test this with Mandrel and Bellsoft distributions.

jvukicev:
@vjovanov Thanks for the suggestion. Adding Mandrel and Liberica NIK as permanent CI dimensions would significantly increase the number and cost of CI jobs, so I tested them locally instead.

I tested the parser and Maven command invocation against actual installations of:

- Oracle GraalVM 25.0.3
- Mandrel 25.0.3.0-Final
- BellSoft Liberica NIK 25.0.3-2

Mandrel uses `OpenJDK Runtime Environment ...`, whereas Oracle GraalVM and Liberica NIK use `GraalVM Runtime Environment ...`. I loosened the parser to recognize the first dotted release on any `Runtime Environment` line, independently of the vendor label, and added the exact outputs as regression-test fixtures.

I also tested same-format 25.1 variants for all three distributions. Assuming they retain their current output structure, they are recognized as 25.1+ and the plugin-generated `--no-fallback` is omitted.

An unknown distribution using a different version-output structure remains conservative: it will retain `--no-fallback`, even on 25.1. The build should still work because the option remains accepted as deprecated and without effect, but it may print the deprecation warning.

We could avoid the version-specific omission entirely if this approach is considered too heuristic, but then builds using the default disabled-fallback configuration would print the `--no-fallback` deprecation warning on every GraalVM 25.1 build. I think the current conservative approach gives us the better tradeoff.

vjovanov:
No need to add to the CI, just test. Excellent!

wirthi (approved):

fniephaus:
Pinging @zakkak for Mandrel/Quarkus and @AlexanderScherbatiy for NIK/Bellsoft.

fniephaus (approved):
LGTM

PR #1011, opened 2026-07-30
Uplift the Native Image layer configuration model

### LINKED ISSUE #977: Proposal: Uplift the layers configuration model (typed model, shared selection, Gradle+Maven parity)
> [!NOTE]
> **This is a design proposal for discussion — not an accepted plan or committed work.** It sketches a direction and its trade-offs so the approach can be reviewed before anyone implements it. Scope, DSL/XML surface, and naming are all open to change.

## Summary

The experimental layered-image support introduced in #719 models a layer as a *binary* and assembles the `-H:LayerCreate` string inline in the plugin. This has produced a leaky DSL, a config-cache bug (#957), and zero Maven parity. This issue proposes uplifting the model: a build-tool-neutral **`ArtifactSelection`** type in `common/` with the flag rendering owned there, a Gradle DSL that treats layers as first-class (not binaries), and a symmetric Maven surface built on a resolvable layer artifact.

## Motivation — where the current model hurts

Grounded in the current code:

- **A layer is modeled as a binary.** You declare `libdependencies { createLayer { … } }` inside `binaries { }`. This forces (a) the `lib`-prefix rule enforced by a thrown exception (`BaseNativeImageOptions.java:444-445`), (b) an output-path override hack (`NativeImageCommandLineProvider.java:203-205`), and (c) a stringly-typed `useLayer("libdependencies")` cross-reference that fails late.
- **It causes a real bug.** #957: because the layer is wired through the application image classpath, an application-source change invalidates the dependency layer, and the model isn't configuration-cache compatible. #957's own proposed direction — *"model create-layer inputs explicitly rather than inheriting the application image classpath"* — is exactly this uplift, in miniature.
- **The flag string bypasses `common/`.** `-H:LayerCreate=…` is assembled inline in `NativeImageCommandLineProvider`; `common/` holds only the constant. This is at odds with **FS-native-builds.2 / FS-common-libraries.1**, which require command-line construction to go through shared utilities.
- **No Maven support at all.** Maven users hand-write `-H:LayerCreate` in `<buildArgs>` and, for cross-module reuse, resort to the profile + `install-file` + re-declared-dependency contortion documented in #788. This violates **GOAL-plugin-parity**.
- **`packages` is unreliable.** #719 itself notes package-level selection breaks on dynamically-loaded optional modules (logback), so `jars` (i.e. resolved dependencies) is the only dependable input. This is the evidence for anchoring the model on *resolved dependencies*, not raw `package=`/`module=`.

### 1. Shared `ArtifactSelection` in `common/`

A build-tool-neutral value type (no Gradle/Maven types — **REQ-no-buildtool-apis**):

```java
// common/
class ArtifactSelection {
    boolean all;
    List<String> modules;     // module=<name>
    List<String> packages;    // package=<name>
    List<Path>   paths;       // path=<jar> — coordinates already resolved by the plugin layer
}
// The only place the module=/package=/path= grammar lives:
String renderLayerCreate(String layerName, ArtifactSelection s);
```

The plugin layer resolves dependency coordinates → `Path`s (Gradle configurations / Maven Aether) and hands the neutral selection to the common renderer. This fixes the FS-common-libraries.1 deviation and gives argument-file conversion for free.

### 2. Gradle DSL restructure

Lift layers out of `binaries`; reference them by typed handle:

```groovy
graalvmNative {
    layers {
        base {
            contents {
                modules("java.base")
                fromConfiguration(configurations.runtimeClasspath)   // all external deps → path=
                dependencies("com.acme:extension") { transitive = true }  // individual dep + closure
            }
        }
    }
    binaries {
        main { layer = layers.base }        // typed; fails at config time on typo
    }
}
```

Retires the `lib`-prefix rule, the output-override hack, and the string linkage. Create-vs-use is modeled by position, not `instanceof` on a mixed `DomainObjectSet`.

### 3. Maven parity

A dedicated `layer-create` goal that produces `libbase.nil` and attaches it as a **`nil`-typed artifact** (registered in the existing `NativeExtension` build extension), consumed as a normal dependency:

```xml
<!-- producer module -->
<layer>
  <name>base</name>
  <contents>
    <modules><module>java.base</module></modules>
    <includeDependencies>all</includeDependencies>
  </contents>
</layer>

<!-- consumer module -->
<dependency>
  <groupId>com.acme</groupId><artifactId>base-layer</artifactId>
  <version>1.0</version><type>nil</type>
</dependency>
<useLayer><artifact>com.acme:base-layer</artifact></useLayer>
```

The `nil` dependency makes the reactor order the base before the app, and works whether the layer comes from the reactor, `~/.m2`, or a remote repo — replacing the #788 workaround.

## Spec alignment

Spec links pinned to commit [`e067164`](https://github.com/graalvm/native-build-tools/tree/e06716403d86bb595d0657e6e5e516bdbc06ffde).

- **[NGOAL-no-flag-mirroring](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/non-goals.md#ngoal-no-flag-mirroring-the-plugins-do-not-add-build-tool-flags-that-only-forward-to-native-image-flags):** the first-class surface is justified by *dependency resolution + artifact wiring*, which the `buildArgs`/`<buildArgs>` pass-through cannot express. Raw-flag knobs (`all`, bare `module=`/`package=`) stay ergonomic conveniences inside the already-integrated block, not the reason it exists.
- **[FS-native-builds.1](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/functional/native-image-builds.md#1-required-inputs)** already lists "layer options" as a first-class optional input — this makes good on that.
- **[FS-common-libraries.1](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/common/docs/functional-spec.md#1-shared-native-image-utilities) / [FS-native-builds.2](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/functional/native-image-builds.md#2-command-line-construction):** moving rendering into `common/` fixes an existing deviation.
- **[GOAL-plugin-parity](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/goals.md#goal-plugin-parity-shared-native-image-behavior-remains-consistent-across-gradle-and-maven) / [FS-plugin-common](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/functional/plugin-common.md#fs-plugin-common-gradle-and-maven-expose-aligned-native-image-plugin-behavior):** closes the Gradle/Maven gap.
- **[NGOAL-no-buildtool-duplicates](https://github.com/graalvm/native-build-tools/blob/e06716403d86bb595d0657e6e5e516bdbc06ffde/docs/spec/non-goals.md#ngoal-no-buildtool-duplicates-the-plugins-do-not-reimplement-capabilities-that-gradle-or-maven-already-provide):** per-binary (main/test/custom) layer sets are the "multi-target boilerplate" case the non-goal explicitly permits.

## Relationship to existing issues

- **#957 / #958 (open, in flight):** same modeling flaw, narrower scope. **Coordinate:** this issue is the strategic model #957 is a symptom of. Since #958 hasn't merged, prefer landing the config-cache fix *on* this model to avoid doing it twice; if #958 lands first, this issue builds on it.
- **#719:** origin of the current DSL (supersedes its `createLayer`/`useLayer` surface).
- **#610, #778:** the `incremental-build` automation both depend on a clean layer model; this unblocks them.
- **#788:** the Maven artifact model here removes that workaround.
- **#558:** future consumer of `ArtifactSelection` (see the companion `-H:Preserve` issue).

## Deliverables

- [ ] `ArtifactSelection` + `renderLayerCreate` in `common/` with unit tests (FS-common-libraries.8).
- [ ] Gradle `layers { }` container + typed `layer =` binary property; migrate samples.
- [ ] Maven `layer-create` goal + `nil` artifact type + `<useLayer>` resolver.
- [ ] Config-cache-correct create-layer inputs (closes #957).
- [ ] Functional tests: individual-dependency selection, cross-module reuse, config cache.

## Out of scope

- The `incremental-build` / `-incremental` automation (#610, #778) — layered *on top* of this model.
- `-H:Preserve` — companion issue, depends on the `ArtifactSelection` deliverable here.

## Open question

Umbrella (absorb #957, retarget #958 onto this model) vs. tactical-then-strategic (#958 ships the narrow fix, this builds on it)? Depends on how close #958 is to merge — needs a call with its author.

jvukicev:
## What changed

This PR turns Native Image layers from an experimental, binary-scoped Gradle feature into a shared model with first-class Gradle and Maven support.

A layer produces a build-time `.nil` artifact plus runtime shared libraries. Consumers pass the `.nil` to Native Image and must ship the runtime libraries with the final executable.

> Layer consumption is supported on GraalVM 25.1 and later. GraalVM 25.0.x is allowed for compatibility but can fail internally after loading a valid layer.

## Gradle DSL: complete surface

A named layer is an entry in `graalvmNative.layers`. Its task is named `native<LayerName>Layer` and writes `build/native/layers/<name>/<name>.nil`.

### Declare layer contents

Use any combination of these selectors in a layer's `contents` block:

```groovy
graalvmNative {
    layers {
        selected {
            contents {
                // JDK modules; include java.base in a module-selecting base layer.
                modules('java.base', 'java.logging')

                // Packages found on the layer classpath.
                packages('org.slf4j')

                // Explicit files, FileCollections, directories, or providers.
                from(files('libs/local.jar'))

                // Every resolved artifact from a Configuration or Provider<Configuration>.
                fromConfiguration(configurations.runtimeClasspath)

                // An individually resolved module; transitive is true by default.
                dependencies('org.slf4j:slf4j-api:2.0.17')

                // An individually resolved module without its transitives.
                dependencies('com.acme:standalone:1.0') {
                    transitive = false
                }

                // A version-catalog accessor is also accepted.
                // dependencies(libs.slf4j.api)
            }

            // Producer-specific Native Image settings.
            buildArgs('-Ob')
            verbose = true
        }

        everything {
            contents {
                // The full runtime dependency graph, including project artifacts.
                all = true

                // all may also be combined with module/package selectors.
                modules('java.base')
            }
        }
    }
}
```

`modules(...)`, `packages(...)`, `from(...)`, `fromConfiguration(...)`, and `dependencies(...)` append selections. `all = true` selects the runtime graph. Empty, invalid, and duplicate selections fail before Native Image runs.

### Consume named layers

The preferred consumer form is lazy, name-based `usesLayer('<name>')`; it works regardless of declaration order. Repeat it to select multiple layers.

```groovy
graalvmNative {
    binaries {
        main {
            usesLayer('selected')
            usesLayer('everything')
        }

        test {
            usesLayer('selected')
        }

        create('api') {
            sharedLibrary = true
            imageName = 'api'
            usesLayer('selected')
        }
    }
}
```

Typed alternatives are available when passing the layer object or a provider:

```groovy
graalvmNative {
    binaries.main {
        useLayer(graalvmNative.layers.selected)
        useLayer(graalvmNative.layers.named('everything'))
    }

    // The singular assignment selects one layer and replaces a prior assignment.
    binaries.test {
        layer = graalvmNative.layers.selected
    }
}
```

Inside a Groovy binary closure, use the qualified `graalvmNative.layers.<name>` form for typed references: the deprecated binary-level `layers` property shadows the extension container.

A layer can itself consume a named layer to construct a stack:

```groovy
graalvmNative {
    layers {
        base {
            contents.modules('java.base')
        }
        framework {
            usesLayer('base')
            contents.packages('org.example.framework')
        }
    }
    binaries.main {
        usesLayer('framework')
    }
}
```

The older binary-scoped forms remain deprecated compatibility adapters:

```groovy
graalvmNative {
    binaries {
        libdependencies {
            createLayer {
                modules = ['java.base']
                classpath.from(externalDependenciesOf(configurations.runtimeClasspath))
            }
        }
        main {
            useLayer('libdependencies')
        }
    }
}
```

`nativeRun` supplies selected layer directories to the platform loader. With the Gradle `application` plugin, `installDist`, `distZip`, and `distTar` include the executable, staged runtime libraries, and a loader-aware launcher. Named-layer reuse is first-class within a Gradle project; cross-project use currently requires manual file/runtime wiring.

## Maven DSL: complete surface

Maven creates a layer with the `layer-create` goal. Configure the plugin as an extension so Maven consistently registers the `nil` artifact type in reactor and repository builds:

```xml
<plugin>
  <version>${native.maven.plugin.version}</version>
  <extensions>true</extensions>
</plugin>
```

### Declare layer contents

Bind `layer-create` in a producer execution. A `<layer>` supports module, package, path, dependency-coordinate, and full-runtime selectors:

```xml
<execution>
  <id>create-selected-layer</id>
  <phase>package</phase>
  <goals><goal>layer-create</goal></goals>
  <configuration>
    <layer>
      <name>selected</name>

      <modules>
        <module>java.base</module>
        <module>java.logging</module>
      </modules>

      <packages>
        <package>org.slf4j</package>
      </packages>

      <paths>
        <path>${project.basedir}/libs/local.jar</path>
      </paths>

      <dependencies>
        <dependency>
          <artifact>org.slf4j:slf4j-api:2.0.17</artifact>
          <transitive>true</transitive>
        </dependency>
        <dependency>
          <artifact>com.acme:standalone:1.0</artifact>
          <transitive>false</transitive>
        </dependency>
      </dependencies>
    </layer>
  </configuration>
</execution>

<execution>
  <id>create-everything-layer</id>
  <phase>package</phase>
  <goals><goal>layer-create</goal></goals>
  <configuration>
    <layer>
      <name>everything</name>
      <all>true</all>
      <!-- Legacy equivalent: <includeDependencies>all</includeDependencies> -->
    </layer>
  </configuration>
</execution>
```

Dependency coordinates use `groupId:artifactId` or `groupId:artifactId:version`; `<transitive>` defaults to `true`. The goal attaches both `<name>.nil` and a platform-classified runtime ZIP. A POM-packaging module may be a layer-only producer.

### Consume one or more layers

Declare each producer as a normal dependency of type `nil`, then select it by coordinate. `<useLayers>` can be configured at plugin level or goal-execution level.

```xml
<dependencies>
  <dependency>
    <groupId>org.example</groupId>
    <artifactId>selected-layer</artifactId>
    <version>${project.version}</version>
    <type>nil</type>
    <scope>runtime</scope>
  </dependency>
  <dependency>
    <groupId>org.example</groupId>
    <artifactId>everything-layer</artifactId>
    <version>${project.version}</version>
    <type>nil</type>
    <scope>runtime</scope>
  </dependency>
</dependencies>

<plugin>
  <extensions>true</extensions>
  <configuration>
    <useLayers>
      <useLayer>
        <artifact>org.example:selected-layer</artifact>
      </useLayer>
      <useLayer>
        <artifact>org.example:everything-layer:${project.version}</artifact>
      </useLayer>
    </useLayers>
  </configuration>
</plugin>
```

`<useLayers>` applies to every goal receiving that configuration: application compilation, `native:test`, shared-library builds, and `layer-create`. Use an execution-specific configuration when those goals need different stacks. Maven consumption is coordinate-only; a local `.nil` file path is not a supported `<useLayers>` form.

Repository consumers resolve the matching runtime ZIP and stage it under `target/native/layer-runtime`. Package that directory with the final executable and expose it through `LD_LIBRARY_PATH` (Linux), `DYLD_LIBRARY_PATH` (macOS), or `PATH` (Windows).

## Shared behavior and deployment

- Shared `common/utils` renders `-H:LayerCreate` and `-H:LayerUse` consistently for both plugins.
- Layer producers keep their selected classpath/module-path inputs; consumers receive the producer layer automatically in the supported Gradle/Maven flows.
- `.nil` files are build-time artifacts. Runtime shared libraries are deployment artifacts and must accompany the final executable.
- `.nil` files can be large; plan Maven repository retention and cache capacity before broadly publishing layers.

## Testing

Validated locally with grounding and formatting checks; shared utility tests; Gradle inspections, unit tests, and layered functional tests; and Maven inspections, unit tests, and layered functional tests.

Functional coverage includes module, package, path, dependency, and all-dependency selection where supported by the Native Image version; main executable, native-test, shared-library, and additional native-test-suite consumers; Maven reactor/`nil` resolution; and layered executable launch with the producer directory on `LD_LIBRARY_PATH`.

Closes #977

## AI workflow

Generated by [Rhei](https://github.com/vjovanov/rhei)’s `github-issue-fix` workflow using 23 completed AI-assisted steps, 3 resolved models, and 3 focused review cycles.

<details>
<summary>View AI workflow details</summary>

Approved proposal ID: `193a8259f600cc11`

The proposal was approved by an authorized repository member with `/rhei approve 193a8259f600cc11` before implementation.

State path:

`issue-intake → approval-check → approval-apply → implement-fix → validate/review cycle 1 → address-review → validate/review cycle 2 → address-review → validate/review cycle 3 → record-blocked-publication → completed`

Deterministic approval, dispatch, review-routing, and completion transitions were performed by Rhei and did not consume model tokens.

1. **Issue intake and repository grounding** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Inspected the issue, repository rules, worktree, and specification fit. Tokens: 1,247,127 total; 1,231,844 input including 1,133,568 cached-read; 15,283 output.
2. **Implement the approved proposal** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Implemented the shared, Gradle, Maven, specification, sample, and test changes. Tokens: 23,727,061 total; 23,664,080 input including 22,914,688 cached-read; 62,981 output.
3. **Validation cycle 1** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Ran grounding, formatting, focused unit, inspection, functional, and Native Image checks. Tokens: 2,849,937 total; 2,836,313 input including 2,691,712 cached-read; 13,624 output.
4. **Requirements review cycle 1** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Reviewed issue scope and accepted behavior. Tokens: 523,722 total; 518,684 input including 450,048 cached-read; 5,038 output.
5. **Specification review cycle 1** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Reviewed grund alignment, repository rules, and specification coverage. Tokens: 559,568 total; 553,019 input including 482,432 cached-read; 6,549 output.
6. **Implementation review cycle 1** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Reviewed code quality, API boundaries, maintainability, and coverage. Tokens: 635,412 total; 627,370 input including 540,160 cached-read; 8,042 output.
7. **Validation review cycle 1** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Assessed test evidence, failures, and remaining CI risk. Tokens: 202,333 total; 198,525 input including 161,792 cached-read; 3,808 output.
8. **Aggregate review cycle 1** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Reconciled focused findings into the first repair list. Tokens: 137,439 total; 134,751 input including 108,800 cached-read; 2,688 output.
9. **Review repair cycle 1** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Addressed the first aggregate review findings. Tokens: 10,152,091 total; 10,136,921 input including 9,917,312 cached-read; 15,170 output.
10. **Validation cycle 2** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Re-ran focused checks after the first repair. Tokens: 3,897,354 total; 3,884,800 input including 3,730,176 cached-read; 12,554 output.
11. **Requirements review cycle 2** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Rechecked issue requirements after repair. Tokens: 239,926 total; 236,685 input including 191,104 cached-read; 3,241 output.
12. **Specification review cycle 2** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Rechecked specifications and grounding after repair. Tokens: 558,130 total; 549,894 input including 473,600 cached-read; 8,236 output.
13. **Implementation review cycle 2** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Rechecked implementation quality and edge cases after repair. Tokens: 951,066 total; 943,220 input including 846,848 cached-read; 7,846 output.
14. **Validation review cycle 2** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Reassessed validation evidence and disclosure. Tokens: 201,903 total; 197,665 input including 158,336 cached-read; 4,238 output.
15. **Aggregate review cycle 2** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Reconciled the second focused review cycle. Tokens: 160,616 total; 158,379 input including 112,256 cached-read; 2,237 output.
16. **Review repair cycle 2** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Addressed the second aggregate review findings. Tokens: 604,631 total; 598,878 input including 542,336 cached-read; 5,753 output.
17. **Validation cycle 3** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Re-ran focused validation after the second repair. Tokens: 2,177,923 total; 2,165,805 input including 2,049,536 cached-read; 12,118 output.
18. **Requirements review cycle 3** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Completed the final requirements-focused review. Tokens: 488,847 total; 483,569 input including 425,088 cached-read; 5,278 output.
19. **Specification review cycle 3** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Completed the final specification and grounding review. Tokens: 490,430 total; 485,942 input including 428,544 cached-read; 4,488 output.
20. **Implementation review cycle 3** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Identified the two remaining legacy compatibility blockers. Tokens: 890,896 total; 881,455 input including 800,896 cached-read; 9,441 output.
21. **Validation review cycle 3** — `openai:gpt-5.6-terra` via Codex, reasoning effort `medium`. Confirmed focused validation passed and recorded broader gaps. Tokens: 239,300 total; 234,312 input including 201,088 cached-read; 4,988 output.
22. **Aggregate review cycle 3** — `openai:gpt-5.6-sol` via Codex, reasoning effort `medium`. Marked publication blocked because the bounded repair visits were exhausted with two fixable compatibility findings. Tokens: 228,588 total; 225,929 input including 199,424 cached-read; 2,659 output.
23. **Blocked-publication record** — `openai:gpt-5.6-luna` via Codex, reasoning effort `medium`. Recorded the local-only result without pushing or opening a PR. Tokens: 135,286 total; 133,502 input including 110,208 cached-read; 1,784 output.

Rhei accounting covers all 23 completed model invocations: 51,299,586 total tokens; 51,081,542 input tokens including 48,669,952 cached-read tokens; and 218,044 output tokens. Cached-read tokens are included in input and total counts. Cache-write dimensions were unsupported, and the recorded models were unpriced.

After the bounded Rhei review-repair loop ended, a human explicitly directed a manual continuation through Codex. That continuation added API-level `@Deprecated` migration documentation, added focused legacy producer/consumer compatibility coverage, reran the relevant grounding, inspection, unit, and functional checks, committed the result, and published this PR. This post-Rhei session is disclosed separately because its token usage was not captured by Rhei accounting.

</details>

### DIFF
--- a/docs/src/docs/asciidoc/layered-images.adoc
+++ b/docs/src/docs/asciidoc/layered-images.adoc
@@ -13,10 +13,7 @@ application. For the complete configuration reference, see
 
 [IMPORTANT]
 ====
-Layered images are an experimental Native Image feature. Native Build Tools can express
-multi-level layer stacks, but current Native Image releases support one shared base layer and a
-final application image. Keep layer chaining disabled until the Native Image version used by the
-build supports it.
+Layered images are an experimental Native Image feature.
 
 Native Build Tools supports layer consumption on GraalVM 25.1 and later. Consumption on GraalVM
 25.0.x is permitted for compatibility, but is unsupported and proceeds at your own risk because
@@ -53,7 +50,7 @@ Both plugins support module, package, path, dependency, and complete-classpath s
 and packages may be combined in the same layer. Keep frequently changing application classes out
 of a broadly reused layer so that application changes do not force the layer to be rebuilt.
 
-For example, a Gradle layer can select JDK modules and packages from resolved dependencies:
+For example, a Gradle layer can select JDK modules and packages:
 
 [source,groovy]
 ----
@@ -63,7 +60,6 @@ graalvmNative {
             contents {
                 modules('java.base', 'java.logging')
                 packages('org.slf4j')
-                fromConfiguration(configurations.runtimeClasspath)
             }
         }
     }
@@ -83,13 +79,12 @@ The equivalent Maven layer definition is:
   <packages>
     <package>org.slf4j</package>
   </packages>
-  <dependencies>
-    <dependency>
-      <artifact>org.slf4j:slf4j-api</artifact>
-    </dependency>
-  </dependencies>
 </layer>
 ----
+
+Package selectors add packages to the layer; they do not filter files selected through
+`fromConfiguration(...)` (Gradle) or `<dependencies>` (Maven). Use a separate configuration or
+dependency selector when the intent is to include the resolved artifact files themselves.
 
 [[gradle-layer-reuse]]
 == Reuse a Layer with Gradle
@@ -140,35 +135,6 @@ When the `application` plugin is present, `installDist`, `distZip`, and `distTar
 main native executable, selected runtime files under `lib/native-layers/<layer>`, and a
 `bin/<imageName>-native` launcher that configures the platform loader. Projects using another
 packaging model can consume the cacheable `nativeLayerRuntimeFiles` task output.
-
-[[gradle-additional-tests]]
-=== Share a Layer with Additional Test Suites
-
-Register an additional native test binary and configure it like any other layer consumer. The
-following fragment assumes an `integTest` source set and `Test` task already exist:
-
-[source,groovy]
-----
-graalvmNative {
-    registerTestBinary('integTest') {
-        usingSourceSet(sourceSets.integTest)
-        forTestTask(tasks.named('integTest'))
-    }
-
-    binaries {
-        test {
-            usesLayer('sharedRuntime')
-        }
-        integTest {
-            usesLayer('sharedRuntime')
-        }
-    }
-}
-----
-
-`nativeTest` and `nativeIntegTest` now build separate native test images while sharing the same
-layer producer. Different suites can instead select different layers when their runtime
-requirements differ.
 
 [[gradle-multiple-binaries]]
 === Share a Layer Between Executables and Libraries
@@ -314,38 +280,6 @@ The equivalent Maven `<modules>` list uses one `<module>` element for each entry
 omits a module, adding it only to the final consumer is too late: Native Image can report a newly
 seen boot package or another layer-compatibility failure. Add the missing module to the producer
 layer and rebuild it.
-
-[[layer-chaining]]
-== Prepare Multi-Level Layer Stacks
-
-Native Build Tools can describe a layer that consumes another layer. In Gradle, call
-`usesLayer` from the consuming layer:
-
-[source,groovy]
-----
-graalvmNative {
-    layers {
-        base {
-            contents.modules('java.base')
-        }
-        framework {
-            usesLayer('base')
-            contents.packages('org.example.framework')
-        }
-    }
-    binaries.main {
-        usesLayer('framework')
-    }
-}
-----
-
-In Maven, declare the base layer as a `nil` dependency of the next producer module and add
-`<useLayers>` to that module's `layer-create` execution. Each produced layer remains an
-independently ordered build artifact.
-
-Do not enable this topology until the Native Image version used by the build supports chained
-shared layers. When deploying a multi-level stack, every layer's runtime shared library must be
-included in the library search path.
 
 [[shipping-layered-images]]
 == Ship a Layered Application
--- a/docs/src/docs/asciidoc/layered-images.adoc
+++ b/docs/src/docs/asciidoc/layered-images.adoc
@@ -17,6 +17,13 @@ Layered images are an experimental Native Image feature. Native Build Tools can 
 multi-level layer stacks, but current Native Image releases support one shared base layer and a
 final application image. Keep layer chaining disabled until the Native Image version used by the
 build supports it.
+
+Native Build Tools supports layer consumption on GraalVM 25.1 and later. Consumption on GraalVM
+25.0.x is permitted for compatibility, but is unsupported and proceeds at your own risk because
+Native Image may fail internally while loading a valid layer. The plugins warn for 25.0.x
+consumption; creating a layer without consuming another layer does not warn. The runtime release
+reported by `native-image --version`, rather than its first component-version line, determines
+this policy.
 ====
 
 [[layered-image-artifacts]]
@@ -26,12 +33,18 @@ Creating a layer produces two kinds of output:
 
 * A `.nil` file that another Native Image build consumes through `-H:LayerUse`. This is a
   build-time artifact.
-* A platform shared library that the final executable loads at runtime, such as `libbase.so` on
-  Linux. This is a deployment artifact.
+* Platform runtime libraries that the final executable loads, such as `libbase.so` and supporting
+  native libraries on Linux. These are deployment artifacts.
 
-The final executable does not contain the layer's shared library. A deployment must therefore
-include the executable and the shared library from every layer it uses. Publishing or attaching
-the `.nil` file does not automatically publish, copy, or embed the runtime libraries.
+The final executable does not contain the layer's runtime libraries. A deployment must therefore
+include the executable and every discovered runtime library from each layer it uses. Headers and
+diagnostic files emitted beside a layer are not runtime inputs.
+
+Maven attaches a deterministic ZIP containing the runtime files alongside the `.nil`. Its
+classifier identifies the target operating system and architecture, for example
+`layer-runtime-linux-amd64`; non-default Linux C libraries such as musl add that compatibility
+dimension. Gradle models the producer directories as task inputs and stages runtime files without
+flattening different layers into one directory.
 
 [[layer-content-selection]]
 == Select Layer Contents
@@ -123,6 +136,11 @@ Run the application and native tests normally:
 Both task graphs reuse `nativeSharedRuntimeLayer`. The Gradle run tasks add the directories of
 their selected layers to the runtime library search path.
 
+When the `application` plugin is present, `installDist`, `distZip`, and `distTar` also include the
+main native executable, selected runtime files under `lib/native-layers/<layer>`, and a
+`bin/<imageName>-native` launcher that configures the platform loader. Projects using another
+packaging model can consume the cacheable `nativeLayerRuntimeFiles` task output.
+
 [[gradle-additional-tests]]
 === Share a Layer with Additional Test Suites
 
@@ -159,6 +177,14 @@ Repeated `usesLayer` calls can select multiple layers, and the same layer can be
 multiple binaries. For example, an executable and a shared-library binary can both select
 `sharedRuntime`. Each consumer keeps its own Native Image options; layer-sensitive options must
 remain compatible between the producer and every consumer.
+
+Shared-library binaries do not generally require a `mainClass`. For multiple custom binaries,
+configure a distinct `imageName` for each one so packaging and human-facing artifact names remain
+unambiguous.
+
+Same-project named-layer reuse is Gradle's first-class sharing model. Cross-project reuse does not
+currently publish a consumable layer variant; it requires manual `getLayerFiles()` or equivalent
+file wiring, including the runtime files needed for deployment.
 
 [[maven-layer-reuse]]
 == Reuse a Layer with Maven
@@ -199,9 +225,10 @@ Bind `layer-create` in the producer module:
 </plugin>
 ----
 
-The goal creates
-_target/native/layers/shared-runtime/shared-runtime.nil_ and attaches it to the project as a
-`nil` artifact.
+The goal creates _target/native/layers/shared-runtime/shared-runtime.nil_ and attaches it to the
+project as a `nil` artifact. It also attaches the platform-classified runtime ZIP. A layer producer
+may use `<packaging>pom</packaging>`: `layer-create` still runs when bound to the lifecycle.
+`skipNativeBuildForPom` applies to `compile-no-fork`, not to `layer-create`.
 
 [[maven-main-and-tests]]
 === Share One Layer Between Build Executions
@@ -260,10 +287,33 @@ different layers. Multiple consumer modules can declare and select the same `nil
 [NOTE]
 ====
 When a native test is executed, the plugin adds the directories containing its selected layer
-artifacts to the platform library search path. For a layer resolved from a repository, the
-runtime shared library must also be staged next to the `.nil` file or added to the search path
-separately.
+runtime files to the platform library search path. For repository dependencies, the plugin
+automatically resolves the matching runtime ZIP and extracts it into a consumer-owned directory
+under `target/native/layer-runtime`. It never relies on a producer `target` directory or extracts
+into the local Maven repository. If an older publication contains only a `.nil`, the build fails
+before Native Image runs and reports the missing companion coordinate.
 ====
+
+Maven `useLayers` is coordinate-based: declare a dependency with `<type>nil</type>` and select it
+by `groupId:artifactId[:version]`. Local layer-file paths are not a supported Maven consumption
+form. Reactor and external-repository consumers use the same declaration.
+
+[[native-test-layer-modules]]
+== Include Modules Required by Every Consumer
+
+A layer shared by applications and native tests must contain every JDK module required by every
+consumer. Native JUnit fixtures commonly need at least `java.management`, and tests using JDBC
+need `java.sql`; add `java.xml` when XML APIs are used. For example:
+
+[source,groovy]
+----
+contents.modules('java.base', 'java.management', 'java.sql')
+----
+
+The equivalent Maven `<modules>` list uses one `<module>` element for each entry. If the producer
+omits a module, adding it only to the final consumer is too late: Native Image can report a newly
+seen boot package or another layer-compatibility failure. Add the missing module to the producer
+layer and rebuild it.
 
 [[layer-chaining]]
 == Prepare Multi-Level Layer Stacks
@@ -307,10 +357,15 @@ example, a Linux distribution can use this layout:
 ----
 application/
 |-- bin/
-|   `-- application
+|   `-- application-native
 `-- lib/
-    |-- libbase.so
-    `-- libframework.so
+    |-- native/
+    |   `-- application
+    `-- native-layers/
+        |-- base/
+        |   `-- libbase.so
+        `-- framework/
+            `-- libframework.so
 ----
 
 The `.nil` files are needed to build downstream images, but they are not required merely to run
@@ -323,8 +378,8 @@ need to consume the layers.
 [source,bash]
 ----
 APP_HOME=/opt/application
-LD_LIBRARY_PATH="$APP_HOME/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" \
-    "$APP_HOME/bin/application"
+LD_LIBRARY_PATH="$APP_HOME/lib/native-layers/base${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" \
+    "$APP_HOME/lib/native/application"
 ----
 
 For a container image, copy both directories and set `LD_LIBRARY_PATH` in the runtime image:
@@ -332,8 +387,8 @@ For a container image, copy both directories and set `LD_LIBRARY_PATH` in the ru
 [source,dockerfile]
 ----
 COPY application/ /opt/application/
-ENV LD_LIBRARY_PATH=/opt/application/lib
-ENTRYPOINT ["/opt/application/bin/application"]
+ENV LD_LIBRARY_PATH=/opt/application/lib/native-layers/base
+ENTRYPOINT ["/opt/application/lib/native/application"]
 ----
 
 [[shipping-macos]]
@@ -342,8 +397,8 @@ ENTRYPOINT ["/opt/application/bin/application"]
 [source,bash]
 ----
 APP_HOME=/opt/application
-DYLD_LIBRARY_PATH="$APP_HOME/lib${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}" \
-    "$APP_HOME/bin/application"
+DYLD_LIBRARY_PATH="$APP_HOME/lib/native-layers/base${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}" \
+    "$APP_HOME/lib/native/application"
 ----
 
 [[shipping-windows]]
@@ -351,11 +406,19 @@ DYLD_LIBRARY_PATH="$APP_HOME/lib${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}" \
 
 [source,powershell]
 ----
-$env:PATH = "$PWD\lib;$env:PATH"
-& "$PWD\bin\application.exe"
+$env:PATH = "$PWD\lib\native-layers\base;$env:PATH"
+& "$PWD\lib\native\application.exe"
 ----
 
-Native Build Tools builds the final executable and each layer's shared library, but does not
-package them together for deployment. The project's packaging process must include the executable
-and all layer shared libraries, and configure the platform loader to find those libraries at
-runtime.
+Gradle application distributions generate this packaging and a loader-aware launcher. Maven stages
+repository runtime bundles for build and native-test execution, but a Maven deployment plugin must
+still copy the executable and staged runtime directories into the final application or container.
+
+[[layer-storage]]
+== Plan Repository Storage
+
+`.nil` files can be hundreds of megabytes even for representative JDK-module or dependency layers.
+Normal Maven transport compression may reduce transfer size, but Native Image must ultimately
+receive the original `.nil`. Plan repository retention, quotas, and remote-cache capacity, and
+publish broadly shared layers deliberately. Content-addressing or cross-artifact deduplication is
+outside the current layer contract.
--- a/docs/src/docs/asciidoc/layered-images.adoc
+++ b/docs/src/docs/asciidoc/layered-images.adoc
@@ -1,0 +1,361 @@
+= Layered Native Images: Reuse, Testing, and Deployment
+The GraalVM team
+
+Native Image layers let several native binaries reuse a separately built part of an image. A
+typical project puts stable modules and dependencies in a base layer, then consumes that layer
+from the application image, native tests, or other native binaries.
+
+This guide focuses on how layers fit into a real build: selecting layer contents, sharing one
+layer between build targets, using layers from native tests, and shipping the resulting
+application. For the complete configuration reference, see
+<<gradle-plugin.adoc#layered-images,the Gradle plugin documentation>> and
+<<maven-plugin.adoc#layered-images,the Maven plugin documentation>>.
+
+[IMPORTANT]
+====
+Layered images are an experimental Native Image feature. Native Build Tools can express
+multi-level layer stacks, but current Native Image releases support one shared base layer and a
+final application image. Keep layer chaining disabled until the Native Image version used by the
+build supports it.
+====
+
+[[layered-image-artifacts]]
+== Understand the Generated Artifacts
+
+Creating a layer produces two kinds of output:
+
+* A `.nil` file that another Native Image build consumes through `-H:LayerUse`. This is a
+  build-time artifact.
+* A platform shared library that the final executable loads at runtime, such as `libbase.so` on
+  Linux. This is a deployment artifact.
+
+The final executable does not contain the layer's shared library. A deployment must therefore
+include the executable and the shared library from every layer it uses. Publishing or attaching
+the `.nil` file does not automatically publish, copy, or embed the runtime libraries.
+
+[[layer-content-selection]]
+== Select Layer Contents
+
+Both plugins support module, package, path, dependency, and complete-classpath selection. Modules
+and packages may be combined in the same layer. Keep frequently changing application classes out
+of a broadly reused layer so that application changes do not force the layer to be rebuilt.
+
+For example, a Gradle layer can select JDK modules and packages from resolved dependencies:
+
+[source,groovy]
+----
+graalvmNative {
+    layers {
+        sharedRuntime {
+            contents {
+                modules('java.base', 'java.logging')
+                packages('org.slf4j')
+                fromConfiguration(configurations.runtimeClasspath)
+            }
+        }
+    }
+}
+----
+
+The equivalent Maven layer definition is:
+
+[source,xml]
+----
+<layer>
+  <name>shared-runtime</name>
+  <modules>
+    <module>java.base</module>
+    <module>java.logging</module>
+  </modules>
+  <packages>
+    <package>org.slf4j</package>
+  </packages>
+  <dependencies>
+    <dependency>
+      <artifact>org.slf4j:slf4j-api</artifact>
+    </dependency>
+  </dependencies>
+</layer>
+----
+
+[[gradle-layer-reuse]]
+== Reuse a Layer with Gradle
+
+Gradle layers are named objects outside the binary container. Calling `usesLayer` adds the
+producer task as a dependency of the consuming binary, so requesting `nativeCompile` or
+`nativeTest` also builds the layer when necessary.
+
+[[gradle-main-and-tests]]
+=== Share One Layer Between the Application and Native Tests
+
+The `main` and `test` binaries can consume the same named layer:
+
+[source,groovy]
+----
+graalvmNative {
+    layers {
+        sharedRuntime {
+            contents {
+                modules('java.base', 'java.logging')
+                fromConfiguration(configurations.runtimeClasspath)
+            }
+        }
+    }
+    binaries {
+        main {
+            usesLayer('sharedRuntime')
+        }
+        test {
+            usesLayer('sharedRuntime')
+        }
+    }
+}
+----
+
+Run the application and native tests normally:
+
+[source,bash]
+----
+./gradlew nativeRun
+./gradlew nativeTest
+----
+
+Both task graphs reuse `nativeSharedRuntimeLayer`. The Gradle run tasks add the directories of
+their selected layers to the runtime library search path.
+
+[[gradle-additional-tests]]
+=== Share a Layer with Additional Test Suites
+
+Register an additional native test binary and configure it like any other layer consumer. The
+following fragment assumes an `integTest` source set and `Test` task already exist:
+
+[source,groovy]
+----
+graalvmNative {
+    registerTestBinary('integTest') {
+        usingSourceSet(sourceSets.integTest)
+        forTestTask(tasks.named('integTest'))
+    }
+
+    binaries {
+        test {
+            usesLayer('sharedRuntime')
+        }
+        integTest {
+            usesLayer('sharedRuntime')
+        }
+    }
+}
+----
+
+`nativeTest` and `nativeIntegTest` now build separate native test images while sharing the same
+layer producer. Different suites can instead select different layers when their runtime
+requirements differ.
+
+[[gradle-multiple-binaries]]
+=== Share a Layer Between Executables and Libraries
+
+Repeated `usesLayer` calls can select multiple layers, and the same layer can be selected by
+multiple binaries. For example, an executable and a shared-library binary can both select
+`sharedRuntime`. Each consumer keeps its own Native Image options; layer-sensitive options must
+remain compatible between the producer and every consumer.
+
+[[maven-layer-reuse]]
+== Reuse a Layer with Maven
+
+Maven represents a layer as an attached artifact of type `nil`. A consumer declares the producer
+as a normal dependency with `<type>nil</type>`, which gives Maven enough information to order
+reactor modules and resolve layers from a repository without adding them to the Java classpath.
+
+[[maven-producer-module]]
+=== Create a Layer in a Producer Module
+
+Bind `layer-create` in the producer module:
+
+[source,xml]
+----
+<plugin>
+  <groupId>org.graalvm.buildtools</groupId>
+  <artifactId>native-maven-plugin</artifactId>
+  <extensions>true</extensions>
+  <executions>
+    <execution>
+      <id>create-shared-runtime</id>
+      <phase>package</phase>
+      <goals>
+        <goal>layer-create</goal>
+      </goals>
+      <configuration>
+        <layer>
+          <name>shared-runtime</name>
+          <modules>
+            <module>java.base</module>
+            <module>java.logging</module>
+          </modules>
+        </layer>
+      </configuration>
+    </execution>
+  </executions>
+</plugin>
+----
+
+The goal creates
+_target/native/layers/shared-runtime/shared-runtime.nil_ and attaches it to the project as a
+`nil` artifact.
+
+[[maven-main-and-tests]]
+=== Share One Layer Between Build Executions
+
+Declare the producer in the consumer module:
+
+[source,xml]
+----
+<dependency>
+  <groupId>org.example</groupId>
+  <artifactId>shared-runtime-layer</artifactId>
+  <version>${project.version}</version>
+  <type>nil</type>
+  <scope>runtime</scope>
+</dependency>
+----
+
+Put `<useLayers>` in the plugin-level configuration when every configured goal should use the
+same layer. In this example the application image and native tests share it:
+
+[source,xml]
+----
+<plugin>
+  <groupId>org.graalvm.buildtools</groupId>
+  <artifactId>native-maven-plugin</artifactId>
+  <extensions>true</extensions>
+  <configuration>
+    <useLayers>
+      <useLayer>
+        <artifact>org.example:shared-runtime-layer</artifact>
+      </useLayer>
+    </useLayers>
+  </configuration>
+  <executions>
+    <execution>
+      <id>build-application</id>
+      <phase>package</phase>
+      <goals>
+        <goal>compile-no-fork</goal>
+      </goals>
+    </execution>
+    <execution>
+      <id>run-native-tests</id>
+      <phase>test</phase>
+      <goals>
+        <goal>test</goal>
+      </goals>
+    </execution>
+  </executions>
+</plugin>
+----
+
+Use execution-specific `<configuration>` blocks instead when application and test images need
+different layers. Multiple consumer modules can declare and select the same `nil` dependency.
+
+[NOTE]
+====
+When a native test is executed, the plugin adds the directories containing its selected layer
+artifacts to the platform library search path. For a layer resolved from a repository, the
+runtime shared library must also be staged next to the `.nil` file or added to the search path
+separately.
+====
+
+[[layer-chaining]]
+== Prepare Multi-Level Layer Stacks
+
+Native Build Tools can describe a layer that consumes another layer. In Gradle, call
+`usesLayer` from the consuming layer:
+
+[source,groovy]
+----
+graalvmNative {
+    layers {
+        base {
+            contents.modules('java.base')
+        }
+        framework {
+            usesLayer('base')
+            contents.packages('org.example.framework')
+        }
+    }
+    binaries.main {
+        usesLayer('framework')
+    }
+}
+----
+
+In Maven, declare the base layer as a `nil` dependency of the next producer module and add
+`<useLayers>` to that module's `layer-create` execution. Each produced layer remains an
+independently ordered build artifact.
+
+Do not enable this topology until the Native Image version used by the build supports chained
+shared layers. When deploying a multi-level stack, every layer's runtime shared library must be
+included in the library search path.
+
+[[shipping-layered-images]]
+== Ship a Layered Application
+
+A portable application bundle should keep the executable and its layer libraries together. For
+example, a Linux distribution can use this layout:
+
+[source,text]
+----
+application/
+|-- bin/
+|   `-- application
+`-- lib/
+    |-- libbase.so
+    `-- libframework.so
+----
+
+The `.nil` files are needed to build downstream images, but they are not required merely to run
+an already-built final executable. Keep them in a build artifact repository when other builds
+need to consume the layers.
+
+[[shipping-linux]]
+=== Linux
+
+[source,bash]
+----
+APP_HOME=/opt/application
+LD_LIBRARY_PATH="$APP_HOME/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}" \
+    "$APP_HOME/bin/application"
+----
+
+For a container image, copy both directories and set `LD_LIBRARY_PATH` in the runtime image:
+
+[source,dockerfile]
+----
+COPY application/ /opt/application/
+ENV LD_LIBRARY_PATH=/opt/application/lib
+ENTRYPOINT ["/opt/application/bin/application"]
+----
+
+[[shipping-macos]]
+=== macOS
+
+[source,bash]
+----
+APP_HOME=/opt/application
+DYLD_LIBRARY_PATH="$APP_HOME/lib${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}" \
+    "$APP_HOME/bin/application"
+----
+
+[[shipping-windows]]
+=== Windows
+
+[source,powershell]
+----
+$env:PATH = "$PWD\lib;$env:PATH"
+& "$PWD\bin\application.exe"
+----
+
+Native Build Tools builds the final executable and each layer's shared library, but does not
+package them together for deployment. The project's packaging process must include the executable
+and all layer shared libraries, and configure the platform loader to find those libraries at
+runtime.
--- a/native-gradle-plugin/docs/functional/plugin-model.md
+++ b/native-gradle-plugin/docs/functional/plugin-model.md
@@ -75,9 +75,10 @@ The same duplicate and name validation applies when a named layer consumes anoth
 Every selection form (`layer =`, `useLayer(NativeImageLayer)`,
 `useLayer(Provider<NativeImageLayer>)`, and `usesLayer(String)`) contributes its logical layer
 name to the same selection set. A consuming binary may select each logical name only once. The
-plugin resolves provider-backed names at a cache-safe validation boundary after configuration and
-fails that validation before any selected producer task executes; it retains lazy name resolution
-and configuration-cache compatibility.
+plugin resolves provider-backed names lazily and fails duplicate-name validation before any
+selected producer task executes. This layer model must not introduce avoidable project or task
+container serialization; general configuration-cache compatibility remains governed by the
+plugin's existing configuration-cache baseline.
 
 `fromConfiguration(...)` and `all` include resolved external and project dependencies. Dependency
 selectors also accept Gradle providers, including version-catalog accessors.
@@ -96,12 +97,17 @@ original naming and external-module-only semantics. Only legacy layer declaratio
 | Native test | `binaries.test.usesLayer` | `LayeredApplicationFunctionalTest` |
 | Shared library | `binaries.main.sharedLibrary` with `usesLayer` | `LayeredApplicationFunctionalTest` |
 | Chained layers | `NativeImageLayer.usesLayer` | `LayeredApplicationFunctionalTest` |
-| Cross-project publication | Gradle task output and file dependency wiring | Gradle-specific; Maven uses reactor/repository `nil` artifacts instead |
+| Application distribution | Main layered binary is added to `installDist`, `distZip`, and `distTar` with staged runtime files and a native launcher | Installed and archived distributions run without producer build directories |
+| Cross-project publication | Not first-class; manual `getLayerFiles()` or file wiring only | No consumable publication variant is promised |
 
 Layer shared libraries remain external runtime dependencies. On Linux set `LD_LIBRARY_PATH`; on
 macOS set `DYLD_LIBRARY_PATH`; and on Windows add the layer directory to `PATH` before running a
-layered executable. `nativeRun` supplies this for its selected layers, but deployed executables
-need the same platform loader configuration. [§root/FS-native-builds.6](../../../docs/spec/functional/native-image-builds.md#6-layered-images).
+layered executable. `nativeRun` supplies this for its selected layers. When the `application`
+plugin is present, the main distribution includes the native executable under `lib/native`, each
+selected layer's runtime files under `lib/native-layers/<layer>`, and a `<imageName>-native`
+launcher under `bin` that configures the loader before execution. Other packaging plugins can
+consume the cacheable `nativeLayerRuntimeFiles` task output.
+[§root/FS-native-builds.6](../../../docs/spec/functional/native-image-builds.md#6-layered-images).
 
 ## 3. Default binaries
 
--- a/native-gradle-plugin/docs/functional/plugin-model.md
+++ b/native-gradle-plugin/docs/functional/plugin-model.md
@@ -34,6 +34,32 @@ The extension feeds the option objects consumed by compile, run, resource-genera
 and native-test tasks. Users should not repeat `imageName`, `mainClass`, `buildArgs`, metadata
 directories, or resource settings separately on each task.
 
+Named Native Image layers are first-class entries in a `layers` container beside `binaries`.
+Each layer owns its contents and deterministic `.nil` output. Binaries consume one or more layer
+objects through typed, lazy references:
+
+```groovy
+graalvmNative {
+    layers {
+        base {
+            contents {
+                modules.add('java.base')
+                fromConfiguration(configurations.runtimeClasspath)
+            }
+        }
+    }
+    binaries {
+        main {
+            useLayer(layers.base)
+        }
+    }
+}
+```
+
+The previous binary-scoped `createLayer` and string-based `useLayer` surface remains as a
+deprecated compatibility adapter. Only legacy layer declarations retain the `lib` binary-name
+rule. [§root/REQ-backwards-compatibility.1](../../../docs/spec/requirements.md#1-deprecation-over-removal).
+
 ## 3. Default binaries
 
 For Java application projects, the plugin must create a `main` binary whose `mainClass` convention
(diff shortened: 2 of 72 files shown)


native-gradle-plugin/docs/functional/plugin-model.md
```diff
@@ -34,6 +34,32 @@ The extension feeds the option objects consumed by compile, run, resource-genera
...
+        base {
+            contents {
+                modules.add('java.base')
+                fromConfiguration(configurations.runtimeClasspath)
+            }
+        }
+    }
+    binaries {
+        main {
+            useLayer(layers.base)
```
vjovanov: This example doesn't work as written. Inside a binary's configuration closure, `layers` resolves delegate-first to the binary's own (deprecated) `NativeImageCompileOptions.getLayers()` `DomainObjectSet`, not to the extension's new `layers` container, so `layers.base` fails with a missing-property error. Both updated samples avoid the short form and write the full `graalvmNative.layers.dependencies` path instead — and the PR description uses the same broken form (`useLayer(layers.dependencies)`).

Beyond fixing the example, the name shadowing between the extension-level `layers` container and the binary-level legacy `layers` property is a standing usability trap for as long as the deprecated surface remains; it deserves at least a documented workaround, and ideally an unambiguous access path in the DSL.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java
```diff
@@ -430,22 +431,48 @@ public void layers(Action<? super DomainObjectSet<LayerOptions>> spec) {
...
+    }
+
+    @Override
+    public void useLayer(Provider<? extends NativeImageLayer> layer) {
+        // Provider-backed layer consumption preserves the typed, lazy extension surface. §FS-plugin-model.2.
+        getLayerFiles().from(layer.flatMap(NativeImageLayer::getOutputFile));
+    }
+
+    @Override
+    public void setLayer(NativeImageLayer layer) {
```
vjovanov, line 489: `setLayer` appends instead of replacing: `layer = layers.a; layer = layers.b` silently configures **two** `-H:LayerUse` arguments, while assigning the *same* layer twice throws "already assigned". Assignment semantics that append-or-throw will surprise anyone combining `binaries.all { }` conventions with per-binary overrides — a setter should replace the previous value.

Related: `useLayer(Provider<...>)` has no duplicate detection at all, and there is no `getLayer()` counterpart to read back the typed assignment (a write-only `layer` property).

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/dsl/LayerContents.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public void packages(String... packages) {
+        getPackages().addAll(Arrays.asList(packages));
+    }
+
+    public void from(Object... paths) {
+        getFiles().from(paths);
+    }
+
+    public void fromConfiguration(Configuration configuration) {
```
vjovanov, line 105: `fromConfiguration` silently filters to `ModuleComponentIdentifier`, so project/subproject dependencies are dropped from the layer with no warning — and nothing in the method name hints at "external modules only". In a multi-module build, `fromConfiguration(configurations.runtimeClasspath)` produces a layer missing the user's own libraries, discovered only through image behavior. Consider logging what was excluded, or at minimum documenting the external-only semantics in the javadoc and user docs.

Also: `dependencies(String)` has no `Provider` overload, so version-catalog accessors (`libs.foo`) can't be used.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
```diff
@@ -508,6 +507,55 @@ private void configureAutomaticTaskCreation(Project project,
...
+            layerOptions.getLayerCreate().set(create);
+            layerOptions.getBuildArgs().convention(layer.getBuildArgs());
+            layerOptions.getVerbose().convention(layer.getVerbose());
+            layer.getVerbose().convention(false);
+
+            Configuration runtimeClasspath = project.getConfigurations()
+                .getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME);
+            create.getClasspath().from(layer.getContents().getAll()
+                .flatMap(all -> Boolean.TRUE.equals(all)
+                    ? layerOptions.externalDependenciesOf(runtimeClasspath)
```
vjovanov: Gradle's `all` selection resolves to *external module dependencies only*, while Maven's `all` (`runtimeArtifactPaths()` in `LayerCreateMojo`) includes reactor project artifacts. The same selector therefore selects different contents in multi-module Gradle vs. Maven builds, contradicting the "equivalent selector behavior" requirement in §FS-native-builds.6. Note also that `all` is hard-wired to `runtimeClasspath` here regardless of what the layer's `contents` block declared.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -609,6 +671,15 @@ protected void buildImage() throws MojoExecutionException {
         }
     }
 
+    // Native Image 25.0.3 can crash after loading a layer, so fail before invocation. §FS-native-builds.3.
+    static void checkLayerConsumptionCompatibility(String versionInformation) throws MojoExecutionException {
```
vjovanov: This guard exists only in Maven — the Gradle plugin has no equivalent check, so Gradle users on Native Image 25.0.3 still hit the raw hosted-layer-finalization crash instead of this actionable message. The PR description claims the guard generically. Either add the same check to the Gradle consumption path (e.g. in `BuildNativeImageTask`) or state explicitly in the spec/description that it is Maven-only.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -493,6 +509,49 @@ protected void addDependenciesToClasspath() throws MojoExecutionException {
         }
     }
 
+    private List<String> resolveLayerUseArguments() throws MojoExecutionException {
```
vjovanov, line 531: Two usability traps in the consumption flow:

1. A declared `<type>nil</type>` dependency that no `<useLayers>` entry references is **silently ignored** — the image builds without the layer and nothing warns the user. Consider warning (or failing) when unreferenced `nil` dependencies exist on the project.
2. The "Layer dependency 'x' was not found" error doesn't hint at the most likely cause: forgetting to declare the corresponding `<type>nil</type>` dependency in `<dependencies>`. Mentioning that in the message would save users a round-trip.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
```diff
@@ -493,6 +509,49 @@ protected void addDependenciesToClasspath() throws MojoExecutionException {
...
+            File layerFile = matches.get(0).getFile();
+            if (layerFile == null) {
+                throw new MojoExecutionException("Layer dependency '" + selector + "' has no resolved file");
+            }
+            arguments.add(NativeImageLayerArguments.renderLayerUse(layerFile.toPath()));
+        }
+        return arguments;
+    }
+
+    private static boolean matchesLayerCoordinate(Artifact artifact, String selector) {
```
vjovanov: Malformed coordinates (e.g. `foo` with no `:`) throw a raw `IllegalArgumentException` from inside a stream filter, surfacing as a stack trace instead of a clean `MojoExecutionException` like the surrounding validation. Same pattern in `LayerCreateMojo.matches`.

vjovanov (commented):
I reviewed this with a focus on usability gaps. The architecture is right — shared selection/rendering in `common`, typed lazy Gradle wiring, Maven-native `nil` artifact resolution — and the error paths that do exist fail before Native Image is invoked. But there are several gaps worth addressing before merge. File-specific issues are inline; the cross-cutting ones are below.

**1. No end-user documentation.** The PR updates the internal spec docs and samples, but `docs/src/docs/asciidoc/gradle-plugin.adoc`, `maven-plugin.adoc`, and the end-to-end guides contain no mention of layers at all, and `changelog.adoc` gets no entry. Users cannot discover the `layers { }` DSL, the `layer-create` goal, `<useLayers>`, or the `nil` dependency type from the published documentation. For a PR whose point is uplifting usability, this is the biggest gap.

**2. The happy path was never verified on a supported Native Image release.** Per the PR's own testing notes, the consumer image was only run on 25.0.3 — the exact release this PR blocks for consumption. The full produce → consume → run flow should be validated on a newer Native Image before merge.

**3. Maven setup silently depends on `<extensions>true</extensions>`** for the `nil` artifact handler registration, but that is only visible in the sample. Forgetting it degrades confusingly (the `nil` dependency falls back to the default artifact handler and can leak onto compile classpaths). It should be documented as a requirement with a recognizable failure mode.

**4. Empty layers aren't validated.** A named Gradle layer with no contents renders an unqualified `-H:LayerCreate=<name>.nil` with an empty classpath and defers to an obscure Native Image error. A configuration-time "layer 'x' has no contents" would be friendlier.

**5. The new model still warns** `Experimental support for layered images enabled. DSL may change at any time.` on every build. If the first-class DSL is the stable uplift, consider scoping or softening that message — it reads oddly against the PR's framing.

Summary of the inline comments: the documented `useLayer(layers.base)` example doesn't resolve (`layers` is shadowed inside binary closures), `layer =` assignment accumulates rather than replaces, `fromConfiguration`/`all` silently drop project dependencies (and Gradle/Maven `all` semantics diverge in multi-module builds), the 25.0.3 guard exists only in Maven, unreferenced `nil` dependencies are silently ignored, and malformed coordinates surface as raw `IllegalArgumentException`s.

native-gradle-plugin/docs/functional/plugin-model.md
jvukicev: Fixed in commit 3d3948392. The spec and user docs now use the unambiguous qualified graalvmNative.layers path inside binary closures, and explicitly document the legacy layers name-shadowing workaround.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/dsl/LayerContents.java
jvukicev: Fixed in commit 3d3948392. fromConfiguration now includes all resolved artifacts, including project and subproject dependencies. I also added Provider overloads for MinimalExternalModuleDependency so version-catalog accessors can be used directly, with tests and user documentation.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
jvukicev: Fixed in commit 3d3948392. Layer coordinates are parsed and validated once through a shared helper. Malformed or blank components now produce a clean MojoExecutionException in both consumption and layer-create. Added parameterized tests for invalid forms.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/NativeImagePlugin.java
jvukicev: Fixed in commit 3d3948392. Gradle all now resolves the complete runtime artifact set, including project dependencies, matching Maven reactor behavior. The shared spec and multi-project tests were updated accordingly.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
jvukicev: Fixed in commit 3d3948392. The Native Image 25.0.3 compatibility check is now shared and runs in Gradle BuildNativeImageTask before layer consumption, producing the same actionable upgrade or remove-layers guidance as Maven.

native-gradle-plugin/src/main/java/org/graalvm/buildtools/gradle/internal/BaseNativeImageOptions.java
jvukicev: Fixed in commit 3d3948392. setLayer now replaces the previous assignment, getLayer exposes the typed assignment, and both direct and provider-backed selections participate in duplicate-name validation. Added unit coverage for replacement, readback, and provider duplicates.

native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/AbstractNativeImageMojo.java
jvukicev: Fixed in commit 3d3948392. Maven now warns for every declared nil dependency not selected by useLayers, including when useLayers is absent. The missing-layer error now tells users to declare the corresponding dependency with type nil. Added tests for both cases.

olpaw:
In the example from the description:
```
    layers {
        dependencies {
            contents {
                modules("java.base")
                fromConfiguration(configurations.runtimeClasspath)
            }
        }
    }
```
is `dependencies` a user chosen name or part of the DSL used for layers?

jvukicev:
> In the example from the description:
> 
> ```
>     layers {
>         dependencies {
>             contents {
>                 modules("java.base")
>                 fromConfiguration(configurations.runtimeClasspath)
>             }
>         }
>     }
> ```
> 
> is `dependencies` a user chosen name or part of the DSL used for layers?

`dependencies` is a user-chosen layer name in the example.

vjovanov:
# PM test report

I reviewed the full diff, published the PR's plugins (`1.1.8-SNAPSHOT`) to Maven local, and drove them with scratch Gradle/Maven projects on Oracle GraalVM 25.0.1, deliberately trying to break the new surface. Overall: **the new model is a real improvement and the happy paths work end-to-end** — but I found **3 hard bugs (one breaks the advertised backwards compatibility), 1 latent CI failure, and a set of first-contact UX traps** that will hit almost every adopter.

## What works (verified by running it)

- **Gradle named layer end-to-end**: layer builds to `build/native/layers/deps/deps.nil`, `nativeCompile` consumes via `-H:LayerUse`, `nativeRun` executes correctly (`LD_LIBRARY_PATH` wiring works).
- **The #957 fix is real**: changing app source leaves the layer task UP-TO-DATE; only the app image rebuilds. Configuration cache stores and reuses with zero problems.
- **Maven reactor flow**: `mvn package` orders base-layer before app via the `nil` dependency, builds `base.nil`, consumes it, produces a working executable. Out-of-reactor consumption from `~/.m2` also works (`mvn install`, then building the consumer alone).
- **Validation messages are mostly good**: empty layer, typo'd `useLayers` coordinate ("Declare the corresponding `<dependency>` with `<type>nil</type>`"), duplicate selection (`[deps, deps]`), `all`+modules conflict, and the unreferenced-nil warning all fire with actionable text.
- Module-only and `all = true` Gradle layers both build successfully.

## Bugs (should block merge)

**1. The legacy Gradle compatibility path is broken end-to-end.** Using exactly the DSL from master's shipped sample (`libdependencies { createLayer { … } }` + `useLayer("libdependencies")`): the producer now emits **`<project-name>.so`** instead of `libdependencies.so`, because `NativeImageCommandLineProvider` removed the legacy `targetOutputPath = layerCreateName` override and only the *new* path sets `imageName = "lib" + name`. The producer "succeeds", then the consumer dies with:

```
VMError$HostedError: Expecting that shared layer library file starts with lib and ends with .so. Found: layer-lab.so
```

This matches the PR's own disclosure that review cycle 3 flagged "two remaining legacy compatibility blockers" and marked publication blocked. The retained unit tests check argument rendering, not the produced artifact name. Fix: have the legacy `createLayer` path set the image name to the binary name (it's already forced to start with `lib`).

**2. `externalDependenciesOf` was renamed without a deprecated alias.** Any build script written against 1.1.7 — including anyone who copied the *current shipped sample* — fails on upgrade with `Could not find method externalDependenciesOf()`. That directly contradicts "keeps the previous APIs operational" and §REQ-backwards-compatibility. Keep a deprecated delegating `externalDependenciesOf` for at least one release. (Note the rename also silently changed semantics: the `ModuleComponentIdentifier` filter was dropped, so it now includes project artifacts.)

**3. Maven modules-only layer leaks the whole project classpath into the layer build.** `LayerCreateMojo` sets `classpath = selectedPaths`; when the selection is modules-only that list is **empty**, and `populateClasspath()` treats empty as "use defaults", so the layer build got `-cp base-layer.jar:commons-lang3.jar` alongside `-H:LayerCreate=base.nil,module=java.base`. Gradle passes *no* `-cp` for the same declaration — a parity break and exactly the "leaky inputs" this uplift was meant to kill. The mojo should distinguish "explicitly empty" from "unset".

**4. Latent CI failure in the Maven functional test.** Its success branch (GraalVM ≥ 25.0.5) executes the layered app **without setting `LD_LIBRARY_PATH`**, and the executable has **no rpath** — `readelf` shows `NEEDED libbase.so` with no RUNPATH; running it fails with `cannot open shared object file`. On 25.0.3/25.0.4 the guard makes the test take the failure branch, so the success branch has never actually run. It will start failing the moment CI gets a newer GraalVM.

## UX traps (the first hour of every adopter)

- **The `layers` shadowing footgun is worse than the docs admit — and the PR description's own example is broken.** `layer = layers.deps` (and the PR body's `useLayer(layers.dependencies)`) fails with `Could not get unknown property 'deps' for LayerOptions collection of type DefaultDomainObjectSet` — the deprecated binary-level `layers` swallows the reference, and nothing in the error hints at `graalvmNative.layers`. Suggestions: fix the PR body; add a typed name-based lookup that resolves from the extension container lazily (e.g. `usesLayer('deps')`), or catch this case and error with "did you mean `graalvmNative.layers.deps`?".
- **Declaration order is load-bearing with no lazy escape.** If `binaries {}` precedes `layers {}` in the build file, both `graalvmNative.layers.deps` *and* `graalvmNative.layers.named('deps')` fail (`NativeImageLayer with name 'deps' not found`). There is currently no order-independent way to reference a layer. Either document "layers first" prominently or accept a name-based provider that defers lookup to task-graph time.
- **No layer-name validation.** `create('my,base layer')` happily registers task `nativeMy,base layerLayer`, then fails minutes later inside native-image with `Unknown option 'my'`. The shared `validateLayerName` only checks blank — both plugins should reject names not matching something like `[A-Za-z0-9._-]+` at declaration time (commas, spaces, `=`, path separators all corrupt the flag).
- **Layer builds don't inherit `binaries.all { buildArgs }`** — the PR's own test fixture had to duplicate `-Ob` into `layers.all` "because layer-sensitive options must match between producers and consumers", but the user docs never mention this. A user with `-O3`/GC flags on binaries will produce mismatched layers and get a confusing native-image failure. At minimum add a docs callout; consider warning on known-sensitive flag mismatches.
- **Metadata-repository asymmetry**: Gradle layer tasks get no reachability-metadata wiring at all (a layer built fine while `nativeCompile` was failing the metadata schema check; Maven's `layer-create` *does* attach metadata dirs). Decide which is right — currently the two producers behave differently.
- **Maven runtime/deployment story is undocumented.** The built app hard-depends on `libbase.so` from the *producer's* target directory; there's no rpath and no doc telling users to set `LD_LIBRARY_PATH` or ship the `.so`. Consider rpath support via `-H:AdditionalLinkerOptions` or at least a docs section; the e2e guide ends at "passes it to the consumer image".
- **The `<extensions>true</extensions>` claim is overstated**: I built the whole layered reactor *and* consumed from the local repo without it. Keep the recommendation, but the IMPORTANT box asserting failure modes should be verified/softened.
- Smaller items:
  - `all = true` can't be combined with `modules(…)` — the restriction feels artificial since the docs' own recommended pattern (`modules` + `fromConfiguration`) is equivalent.
  - Undocumented `includeDependencies` accepts only `"all"` and silently ignores anything else.
  - The 25.0.3/25.0.4 guard has no override property and won't match vendor version formats with suffixes (e.g. Mandrel-style `x.y.z.n`).
  - Duplicate detection with provider forms only fires at execution, after the layer has already spent minutes building.
  - No Kotlin DSL sample (the Groovy container syntax doesn't translate 1:1).
  - Gradle has no cross-project layer sharing story (Maven does) — worth a "future work" note for parity.

## Suggested priority

Items 1–4 are merge blockers (1 and 2 break the PR's explicit compatibility promise; 3 breaks the model's core claim; 4 is a time bomb in CI). The shadowing/ordering/name-validation trio is what I'd fix next — they're cheap, and they're the first three errors any real adopter will hit.

---
*Report produced by a hands-on review: static review of all 55 files plus scripted break-it scenarios against locally published `1.1.8-SNAPSHOT` plugins (Gradle 9.6.1, Maven, Oracle GraalVM 25.0.1).*

vjovanov:
# Follow-up: layer test-coverage audit

Two questions worth answering explicitly for this PR: (1) are all selector types — `module`, `package`, `path`, `all` — tested, and (2) are layers consumed from all binary kinds — executables, test binaries, shared libraries? The answer to both is **no** — coverage is thinner than it looks.

## Selector coverage

| Level | module | package | path | all |
|---|---|---|---|---|
| Shared renderer unit test (`NativeImageLayerArgumentsTest`) | ✔ | ✔ | ✔ | ✔ |
| Gradle unit (`NativeImageCommandLineProviderTest`) | ✔ | ✔ | ✔ (jars) | – |
| Gradle functional (real native-image build) | ✔ | ✗ | ✔ (`fromConfiguration`) | ✗ |
| Maven unit (`LayerConfigurationTest`) | – | ✗ | ✔ (dependency matching only) | ✔ (mapping only) |
| Maven functional (real build) | ✔ | ✗ | ✔ (`<dependencies>`) | ✗ |

Every selector is covered at the *string-rendering* level, but against a real native-image only `module=` + `path=` ever run:

- **`package=` is never exercised in any real build**, in either plugin. Given #719 itself flagged package selection as the unreliable one (the logback case), it is the selector most in need of a functional test — and the Gradle unit coverage for it goes through the *legacy* `CreateLayerOptions` path, not the new `layers { contents { packages(...) } }` DSL.
- **`all` is never built for real** anywhere. (I built one manually against 25.0.1 — it works — but the repo has no test proving the bare `-H:LayerCreate=name.nil` + full-classpath form produces a usable layer.)
- **Maven's explicit `<paths>` selector** (`LayerConfiguration.getPaths()` → `resolveSelectedPaths`) has zero tests at any level.
- **A modules-only Maven layer is untested** — the sample and functional test always add a dependency. That is exactly the untested combination where the classpath-leak bug from my earlier report hides (empty `classpath` list falls back to the full project classpath).

## Consumption contexts

Only **main executable binaries** consume layers in any test, in both plugins:

- **Gradle test binary**: nothing links a layer into `binaries.test` / `nativeTest`. The plumbing exists (`DelegatingCompileOptions` explicitly forwards `useLayer` / `getLayerCreate` / `getLayerFiles`, which only matters for delegated test-style options), so the PR added the forwarding but never tests the scenario it enables.
- **Shared library consumer**: no test or sample builds a `sharedLibrary = true` binary that consumes a layer. Since layer linkage is literally dynamic-library linkage (`NEEDED libbase.so`), a shared-lib-consuming-a-layer case has real failure potential and is unverified.
- **Layer chaining (layer consumes layer)** — a gap beyond testing: the new Gradle `NativeImageLayer` DSL has **no consumption API at all** (only `contents`, `buildArgs`, `verbose`), so a multi-level layer stack — which Native Image supports (`LayerUse` + `LayerCreate` in one invocation) and the *legacy* binary-scoped DSL could express — is now **inexpressible** in the new model. Maven can express it (`layer-create` inherits `useLayers` from `AbstractNativeImageMojo`) but never tests it.
- **Maven `native:test` with layers**: `NativeTestMojo` extends `AbstractNativeImageMojo`, so a plugin-level `<useLayers>` configuration silently applies to the test image too — untested, and arguably surprising for users who set `useLayers` in the shared plugin `<configuration>`.

## Suggested additions

1. One functional test per plugin using `packages` selection.
2. A modules-only Maven layer functional test (it will catch the classpath-leak bug).
3. A `nativeTest`-consumes-layer test (Gradle) and a `native:test`-with-`useLayers` test (Maven) — or an explicit decision that test images do not consume layers.
4. A shared-library-consumes-layer test.
5. A decision on whether the new Gradle DSL should support layer chaining before the legacy path (which could express it) is removed.

vjovanov:
## Product review: driving the layer feature end-to-end

I published the plugins from this branch and used them from real Gradle and Maven projects (not the functional test harness) to answer a few adoption questions: does it work with non-Maven dependencies, can it be used for tests / shared libraries / executables, can several consumers share one layer, and can the result actually be shipped.

Short version: the DSL and goal surface are in good shape, and the common cases work. The blocker is deployment — a layer consumed from a Maven repository builds successfully and produces a binary that cannot start.

### Environment

Layer *consumption* is broken on GraalVM 25.0.x, so everything below was run on GraalVM CE 25.1.3 unless stated otherwise.

```bash
# publish the branch
git fetch origin pull/1011/head:pr-1011 && git checkout pr-1011
./gradlew publishAllPublicationsToCommonRepository --no-parallel   # -> build/common-repo

# GraalVM CE 25.1.3
curl -sLO https://github.com/graalvm/graalvm-ce-builds/releases/download/graal-25.1.3/graalvm-community-jdk-25i1-25.0.3_linux-x64_bin.tar.gz
tar xzf graalvm-community-jdk-25i1-25.0.3_linux-x64_bin.tar.gz
export GRAALVM_HOME=$PWD/graalvm-community-25.1.3+9.1
export JAVA_HOME=$GRAALVM_HOME
```

---

## 1. Blocker: a repository-resolved layer produces an unrunnable binary

`layer-create` attaches only the `.nil`. The `.so` that the executable actually links against is never attached, so it never reaches the local or remote repository.

**Repro**

```bash
cd samples/layered-maven-application
mvn -Dnative.maven.plugin.version=1.1.8-SNAPSHOT install     # reactor build, succeeds

ls ~/.m2/repository/org/graalvm/buildtools/samples/base-layer/1.0-SNAPSHOT/
# base-layer-1.0-SNAPSHOT.jar
# base-layer-1.0-SNAPSHOT.nil   (916 MB)
# base-layer-1.0-SNAPSHOT.pom
find ~/.m2/repository -name '*.so'            # nothing

# now build a consumer OUTSIDE the reactor: copy application/ out, drop the <parent>,
# give it its own groupId/version, and keep the <type>nil</type> dependency + <useLayers>
cd /tmp/standalone-consumer && mvn package    # succeeds
./target/standalone-consumer
```

```
./target/standalone-consumer: error while loading shared libraries:
libbase.so: cannot open shared object file: No such file or directory
```

The same happens for `native:test`. The test image builds, `getUsedLayerLibraryDirectories()` points `LD_LIBRARY_PATH` at the `.m2` directory holding the `.nil`, and the run dies:

```
target/native-tests: error while loading shared libraries: libbase.so: cannot open shared object file
[ERROR] Failed to execute goal ...:test ... returned non-zero result
```

It only works inside a reactor because `base.nil` and `libbase.so` happen to sit together in `target/native/layers/base/`.

So the PR's claim that "attached `nil` artifacts work in reactor, local-repository, and remote-repository builds" holds for the *build* but not for the *result*. The `.nil` is a build-time input and the `.so` is the runtime artifact; they need to travel together — attaching the `.so` as a second artifact (classifier or `so`/`dylib`/`dll` type) would fix both the standalone build and `native:test`.

Related sizing concern: the `java.base` layer installs a **916 MB** `.nil` into `~/.m2`, and Gradle produces 249–321 MB `.nil` files. That is not something teams will push to a shared repository, which makes the `.nil`-as-Maven-artifact model worth a second look.

On the Gradle side the same gap exists but is hidden: `nativeRun` sets `LD_LIBRARY_PATH` for you, so it looks fine in development, but nothing wires the `.so` into `installDist`/`distZip`, there is no `-rpath $ORIGIN` support, and `gradle-plugin.adoc` has no deployment section at all.

```bash
build/native/nativeCompile/app
# error while loading shared libraries: liblegacy.so: cannot open shared object file
```

---

## 2. No version gate on layer consumption

On GraalVM 25.0.1 — the LTS most users have installed — consuming a layer fails with a raw Native Image internal error:

```
Error: Trying to create a new ImageHeapConstant for
com.oracle.svm.core.reflect.serialize.SerializationSupport@... after the ImageHeapScanner is sealed.
```

The test suite already knows about this (`hasLayerConsumptionBug()` = `!isGraalVMVersionAtLeast(version, 25, 1)`), but there is no equivalent check in the plugins, so users get the crash with no hint. A pre-flight error like *"Native Image layers require GraalVM 25.1 or later"* would save a lot of confusion.

Related: the PR description says validation ran on Oracle GraalVM 25.0.2, where `hasLayerConsumptionBug()` skips the layer-consumption tests. Worth confirming which layer tests CI actually executed before merging.

---

## 3. Native tests fail unless the layer has the right JDK modules

The sample ships `modules('java.base')`, which cannot work for a test image. With it:

```
Error: com.oracle.svm.shared.util.VMError$HostedError: Newly seen platform package package java.sql
```

Adding `java.sql` and `java.management` fixes it — which is exactly what the functional test in this PR had to do:

```groovy
layers.dependencies.contents.modules('java.sql', 'java.management')
binaries.test { usesLayer('dependencies') }
```
```
[         1 tests successful      ]
```

This requirement is not documented anywhere, and `Newly seen boot package package ...` is not a message anyone can act on. Application code that touches XML additionally needs `java.xml`; adding too many modules then trips a different failure (`The core type AnalysisType<...> was not seen as reachable in the initial layer`), so the working window is narrow and needs guidance.

---

## 4. Custom shared-library binary + layer crashes without `mainClass`

```groovy
binaries {
  create('api') { usesLayer('dependencies'); sharedLibrary = true }   // fails
}
```
```
java.lang.Error: ImageSingletons do not contain key com.oracle.svm.core.JavaMainWrapper$JavaMainSupport
```

Isolated it to the combination:

| configuration | result |
| --- | --- |
| `create('api') { sharedLibrary = true }` (no layer) | builds |
| `binaries.main.sharedLibrary = true` + layer | builds |
| `create('api') { usesLayer(...); sharedLibrary = true }` | **crashes** |
| `create('api') { usesLayer(...); sharedLibrary = true; mainClass = '...' }` | builds |

Worth validating in the plugin rather than letting Native Image abort.

---

## 5. Non-Maven dependencies

**Gradle: works well.** I built a jar by hand with `javac`/`jar` (no build tool, no coordinates) plus a `project(':liblocal')` dependency:

```groovy
dependencies {
    implementation files('libs/legacy.jar')
    implementation project(':liblocal')
    implementation 'org.apache.commons:commons-lang3:3.17.0'
}
graalvmNative {
    layers {
        legacy     { contents { from(files('libs/legacy.jar')) } }
        everything { contents { fromConfiguration(configurations.runtimeClasspath) } }
    }
    binaries { main { usesLayer('legacy') } }
}
```

Both layers build, `fromConfiguration` correctly wires `:liblocal:jar` as a producer task dependency, and `./gradlew :app:nativeRun` prints `legacy-jar | project-dependency | Maven-dependency`. No complaints here.

**Maven: producing is fine, consuming is coordinate-only.** `UseLayerConfiguration` has a single `artifact` field parsed as `groupId:artifactId[:version]`, so there is no way to point `<useLayers>` at a `.nil` file. A layer produced by Gradle, downloaded, or vendored cannot be consumed without `mvn install:install-file`. A `<useLayer><file>...</file></useLayer>` form would close this.

---

## 6. Sharing one layer across consumers and across projects

**Within one Gradle project: works.** Four consumers on one layer configure cleanly, and `main` + a second executable + `test` all build against a single `nativeDependenciesLayer`:

```
> Task :nativeDependenciesLayer UP-TO-DATE
> Task :nativeCompile
> Task :nativeOtherCompile
> Task :nativeTestCompile
```

**Across Gradle projects: only via an undocumented escape hatch.** `NativeImageCompileOptions.getLayerFiles()` is public, and this does work:

```groovy
binaries.create('external') {
  mainClass = 'org.graalvm.demo.Application'
  layerFiles.from(someProject.layout.buildDirectory.file('native/layers/deps/deps.nil'))
}
```
```
- '-H:LayerUse' (origin(s): command line)
Finished generating 'layered-java-application' in 29.1s.
```

But there is no task-dependency wiring (a clean build fails), no name validation, and no compatibility-classpath propagation. A supported `usesLayer(file)` / `usesLayerFrom(configuration)` — ideally an outgoing variant so `project(':base-layer')` just works — would make this the answer to "how do several projects share one layer" instead of a trick.

---

## 7. `pom`-packaging producer works, and is probably the pattern to recommend

A module whose only job is declaring dependencies:

```xml
<packaging>pom</packaging>
<dependencies><dependency>...commons-lang3...</dependency></dependencies>
...
<execution>
  <phase>package</phase>
  <goals><goal>layer-create</goal></goals>
  <configuration><layer><name>deps</name><all>true</all></layer></configuration>
</execution>
```

`mvn install` produces a clean `pom-only-layer-1.0.0.nil` + `.pom` with no jar. This is the tidiest producer shape and is not mentioned anywhere in the docs.

---

## 8. Smaller items

- **Configuration cache fails**: `15 problems were found storing the configuration cache` (`DefaultProject`, `TaskContainer`, `SourceSetContainer` not serializable). Pre-existing — `BaseNativeImageOptions` on master already holds a `TaskContainer` — but `NativeImageLayer` and `LayerContents` add a `Project` field, so "configuration-cache-safe" in the description overstates the current state.
- Custom binaries all default to the same image name (`layered-java-application` appears in both `nativeCompile/` and `nativeOtherCompile/`). Pre-existing, but layers make multi-binary builds much more common.
- `changelog.adoc` gains a "Release 1.1.6" section holding an unrelated Surefire entry, inserted above "Release 1.0.1".

---

## Documentation gaps

The new adoc sections cover syntax well. What is missing is everything needed to actually ship:

1. **Requirements** — "layers require GraalVM 25.1 or later", at the top of both pages.
2. **Deployment, with a worked example** — currently one paragraph in `maven-plugin.adoc` and nothing in `gradle-plugin.adoc`. Needs: which files ship (`.so`, not `.nil`), how to place them next to the application, a container/tarball layout, and the `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` / `PATH` setup. As written, the docs describe a build whose output will not start.
3. **`.nil` vs `.so`** — build-time input vs runtime artifact, and their real sizes, so nobody is surprised by a 900 MB repository artifact.
4. **Module selection for native tests** — that `java.base` alone is not enough, that `java.sql` / `java.management` / `java.xml` are commonly required, and what `Newly seen boot package ...` means. This is the most likely first failure a user hits.
5. **Sharing a layer across modules** — an end-to-end Maven example (producer, consumers, reactor vs repository) and, for Gradle, an honest statement of the current limitation.
6. **`pom`-packaging producer module** as the recommended producer shape.
7. **Shared libraries** — that a layered shared-library binary needs a `mainClass`.

Happy to open follow-up issues for any of these, or send a docs patch for items 1–4.

vjovanov:
## Correction: the "core type ... not seen as reachable the initial layer" failure is a missing `java.base`

In my earlier review I described this failure:

```
com.oracle.svm.shared.util.VMError$HostedError: The core type
AnalysisType<RuntimeMetadataDecoderImpl$$Lambda/0x... -> LambdaSubstitutionType@...,
instantiated: true, reachable: true> was not seen as reachable the initial layer.
```

and attributed it to "adding too many modules" to the layer. **That was wrong.** I've since bisected it, and the actual cause is much simpler and much more actionable:

> **A module-selecting layer that omits `java.base` produces a broken layer.**

The reason I misread it is that the passing case in my notes was the sample, whose `build.gradle` already contains `modules("java.base")` — my appended `modules('java.sql', 'java.management', 'java.xml')` *added* to it rather than replacing it. My own projects specified only the extra modules, so they silently lost `java.base`.

### Evidence

One variable changed, everything else identical (GraalVM CE 25.1.3, `nativeTest`):

| layer modules | app | result |
| --- | --- | --- |
| `java.sql, java.management, java.xml` | trivial | ❌ core type ... not seen as reachable the initial layer |
| `java.base, java.sql, java.management, java.xml` | trivial | ✅ `1 tests successful` |
| `java.sql, java.management, java.xml` | commons-lang3 + file jar + project dep | ❌ core type ... not seen as reachable the initial layer |
| `java.base, java.sql, java.management, java.xml` | commons-lang3 + file jar + project dep | ✅ `1 tests successful` |

### Repro

```groovy
graalvmNative {
    metadataRepository.enabled = false
    layers { dependencies { contents { modules('java.sql', 'java.management', 'java.xml') } } }
    binaries { main { usesLayer('dependencies') }; test { usesLayer('dependencies') } }
}
```
```bash
./gradlew :nativeTest      # -> core type ... was not seen as reachable the initial layer
```
Add `'java.base'` to the `modules(...)` list, change nothing else:
```bash
./gradlew :nativeTest      # -> [ 1 tests successful ]
```

### Corrected model of layer module selection

Two distinct failure modes, and they're easy to confuse:

- **Missing `java.base`** → `The core type AnalysisType<...> was not seen as reachable the initial layer`. Reads like an internal Native Image bug; gives no hint that a module is missing.
- **Missing some other needed module** → `Newly seen boot package <pkg>` / `Newly seen platform package <pkg>`. Nameable: add the module that owns the package.

So `java.base` is **necessary but not sufficient**. My original point stands that the sample's `modules('java.base')` alone is not enough for a native test — that still fails with `Newly seen platform package package java.sql`, and needs `java.sql` and `java.management` on top. The correct guidance is `java.base` **plus** whatever the final image touches.

### Suggested fix

This is cheap to catch in the plugin. When a layer has a non-empty module selection that doesn't include `java.base`, fail (or at minimum warn) before invoking Native Image, with a message naming `java.base`. That converts the least actionable error in the whole feature into a one-line fix for the user. It fits naturally next to the existing empty/duplicate/malformed validation in `ArtifactSelection` / `NativeImageLayerArguments`.

Failing that, the docs must state plainly that `java.base` is effectively mandatory in every module-selecting layer, and that `modules(...)` appends rather than replaces — the sample's `modules("java.base")` masking my override is exactly how a user will trip over this.

### What this does and doesn't change

It resolves the one failure in my review I couldn't characterize, and it's good news: **layered native tests do work for non-trivial applications on GraalVM 25.1**, once `java.base` is present. The other findings are unaffected — in particular the `.so` publishing gap remains the blocker, since a repository-resolved layer still yields a binary that cannot start.

vjovanov:
The PR description still doesn't show me all the possibilities of the DSL.

docs/src/docs/asciidoc/layered-images.adoc
```diff
@@ -0,0 +1,424 @@
...
+
+[source,groovy]
+----
+graalvmNative {
+    layers {
+        sharedRuntime {
+            contents {
+                modules('java.base', 'java.logging')
+                packages('org.slf4j')
+                fromConfiguration(configurations.runtimeClasspath)
```
vjovanov: Won't this capture the `org.slf4j`?

docs/src/docs/asciidoc/layered-images.adoc
```diff
@@ -0,0 +1,424 @@
...
+----
+graalvmNative {
+    registerTestBinary('integTest') {
+        usingSourceSet(sourceSets.integTest)
+        forTestTask(tasks.named('integTest'))
+    }
+
+    binaries {
+        test {
+            usesLayer('sharedRuntime')
```
vjovanov: Where does this come from?

docs/src/docs/asciidoc/layered-images.adoc
```diff
@@ -0,0 +1,424 @@
...
+
+[[gradle-additional-tests]]
+=== Share a Layer with Additional Test Suites
+
+Register an additional native test binary and configure it like any other layer consumer. The
+following fragment assumes an `integTest` source set and `Test` task already exist:
+
+[source,groovy]
+----
+graalvmNative {
```
vjovanov: Can these live in our tests?

docs/src/docs/asciidoc/layered-images.adoc
```diff
@@ -0,0 +1,424 @@
...
+contents.modules('java.base', 'java.management', 'java.sql')
+----
+
+The equivalent Maven `<modules>` list uses one `<module>` element for each entry. If the producer
+omits a module, adding it only to the final consumer is too late: Native Image can report a newly
+seen boot package or another layer-compatibility failure. Add the missing module to the producer
+layer and rebuild it.
+
+[[layer-chaining]]
+== Prepare Multi-Level Layer Stacks
```
vjovanov: Is this possible?

docs/src/docs/asciidoc/layered-images.adoc
jvukicev: Good catch. Yes, the configuration selector captures the resolved artifact, so combining it with the package selector was misleading. I removed it from this example. Our package-selector functional test already uses the intended form, `packages("org.slf4j")`, without `fromConfiguration`.

docs/src/docs/asciidoc/layered-images.adoc
jvukicev: It was assuming a project-defined `integTest` source set and task, which is too project-specific for this guide. I removed that additional-test-suite example.

docs/src/docs/asciidoc/layered-images.adoc
jvukicev: Could you clarify what you mean here? We already test the documented layer scenarios in our functional tests. I had intended this as a separate user-facing guide for layered-image workflows, rather than moving the documentation into test fixtures.

docs/src/docs/asciidoc/layered-images.adoc
jvukicev: The DSL can express it, but current Native Image releases do not support the resulting chained shared-layer topology. I removed the public documentation and support claims, while leaving the API in place for a future Native Image release that supports it. We also retain an ignored functional test for this capability.

vjovanov: Well, I would not mention it as it doesn't work.

docs/src/docs/asciidoc/layered-images.adoc
vjovanov: Like the doc snippets. I would like to test all of these verbatim.

vjovanov (approved):

vjovanov:
We can do followup fixes for this later if needed. It has passed enough reviews that we can merge it and use it in the next release.

PR #845, opened 2026-03-08
Improve native-image executable discovery with explicit/convention launcher distinction

vpelikh:
## Summary

This PR makes native-image executable discovery safer and more predictable by distinguishing an **explicitly configured** launcher from one selected **by convention**, so the plugin always builds against a launcher that actually provides `native-image`, and fails with a clear message when it cannot.

### Before this PR

- Toolchain detection (`toolchainDetection = true`) used the toolchain-resolved JDK as the launcher without checking whether that JDK contained `native-image`, leading to confusing failures.
- When the explicit `javaLauncher` lacked `native-image`, the build failed silently or with an opaque downstream error.
- There was no clear, documented rule for how an explicitly chosen launcher relates to the `GRAALVM_HOME` / `JAVA_HOME` fallback chain.

### After this PR

**Explicit `javaLauncher` — treated as authoritative.**

- If the user explicitly sets `graalvmNative.binaries.<name>.javaLauncher` and that launcher does not contain `bin/native-image`, the build fails immediately with a message naming the launcher and its installation path. No fallback, no `gu install`.

**Convention launcher (toolchain detection) — used only when it can supply Native Image.**

- When `toolchainDetection = true`, toolchain detection is used as the `javaLauncher` convention, so `graalvmNative.binaries.main.javaLauncher` stays queryable.
- At build time the plugin checks whether the resolved launcher actually contains `bin/native-image`. If it does, that launcher is used.
- If it does not, the plugin transparently falls back to the environment chain (`GRAALVM_HOME` → `JAVA_HOME` → the Gradle JVM `java-home`), optionally installing `native-image` via `gu` there. A convention-selected launcher never blocks a working GraalVM in the environment. §FS-native-invocation.1.5
- `gu install native-image` runs only on this environment-fallback path, never against a user-configured or toolchain installation (a convention launcher is never mutated in place).
- If no source yields `native-image`, the build fails with a diagnostic that lists the probed paths and the selected location source.

Net effect: the plugin finds a usable `native-image` whenever one exists on the machine, fails fast with actionable messages when it does not, and always respects an explicit choice over automation.

### Spec documentation

- `FS-native-invocation.1` (Executable discovery) documented in `native-gradle-plugin/docs/functional/native-image-invocation.md`:
  - §1.1: Explicit java launcher
  - §1.2: Convention-selected launcher
  - §1.3: Environment-variable fallback (GRAALVM_HOME → JAVA_HOME → Gradle JVM)
  - §1.4: gu-based installation
[...]

PR #1024, opened 2026-08-12
Make NativeImageLayerRuntimeTest platform-independent

vpelikh:
## Summary

This PR makes `NativeImageLayerRuntimeTest` pass on all platforms by selecting the shared-library extension from the host OS instead of hard-coding `.so`.

### Before this PR

- The test hard-coded `.so` filenames (e.g. `libbase.so`, `nested/libjava.so.1`), so it only passed on Linux.
- On macOS the runtime libraries are `.dylib`, and on Windows `.dll`, so the test's assertions on discovered files, archive entries, and extraction all failed.

### After this PR

- The test resolves the OS-specific shared-library extension once via a new `platformLibraryExtension()` helper (`.dylib` on macOS, `.dll` on Windows, `.so` elsewhere).
- All runtime-discovery, archive, and extraction assertions use that extension, so the test verifies the same behavior on every platform.
- Test renamed from `discoversOnlyRuntimeLibrariesAndArchivesThemDeterministically` to `discoversSharedLibraryAndArchivesThemDeterministically` to match the shared-library wording.

### Verification

- `./gradlew :utils:test --tests '*NativeImageLayerRuntimeTest' --rerun-tasks` passes on macOS: `2 tests, 0 failures, 0 errors` including the updated `discoversSharedLibraryAndArchivesThemDeterministically()`.

### DIFF
--- a/common/utils/src/test/java/org/graalvm/buildtools/utils/NativeImageLayerRuntimeTest.java
+++ b/common/utils/src/test/java/org/graalvm/buildtools/utils/NativeImageLayerRuntimeTest.java
@@ -46,6 +46,7 @@ import org.junit.jupiter.api.io.TempDir;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Locale;
 import java.util.zip.ZipFile;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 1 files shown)


vpelikh:
Found while working on #845.

jvukicev:
Thanks for the fix! LGTM, I'll merge it if/after all CI checks pass.

jvukicev (approved):

PR #1026, opened 2026-08-13
Move grund.toml to the project roots

### LINKED ISSUE #1025: Move grund.toml to the project roots
`grund` discovers its config in a fixed order: a bare `grund.toml` first, then `.agents/grund.toml`. This repo still keeps all four configs under `.agents/` — the workspace root plus the `common`, `native-gradle-plugin`, and `native-maven-plugin` members.

Move each one up to the project root it configures, so the config sits with the other top-level project files rather than hidden a directory down, and matches the layout `grund` uses in its own repo.

Two things ride along, both required rather than optional:

- `.github/workflows/check-grund-spec.yml` filters on the path `.agents/grund.toml`. Left alone, the spec gate would stop triggering on config edits.
- `GRUND_VERSION` is pinned to 0.5.0. Root discovery landed in grund 0.9.0, so 0.5.0 would not find the moved configs — and a grund that finds no config exits 0 without checking anything, so the gate would go green while validating nothing.

Bumping the pin also surfaces something already broken: under 0.9.0 all four `AGENTS.md` files fail `grund check` with `outdated grund init block v3`. `grund init` refreshes them to v7.

No config keys change.

vjovanov:
Closes #1025.

`grund` probes a bare `grund.toml` before `.agents/grund.toml`, so all four configs move up to the project root each one configures — the workspace root plus `common`, `native-gradle-plugin`, and `native-maven-plugin`.

### What's in here

1. Four `git mv`s, pure renames. No config keys change.
2. `.github/workflows/check-grund-spec.yml`: path filter `.agents/grund.toml` → `grund.toml`, and `GRUND_VERSION` 0.5.0 → 0.9.0. The member configs stay covered by the existing `common/**` and `native-*-plugin/**` filters.
3. `grund init` in each of the four projects, refreshing the managed block in every `AGENTS.md` from v3 to v7.

### Why the version bump is load-bearing

Root discovery landed in grund 0.9.0. Pinned at 0.5.0 the gate would look only under `.agents/`, find nothing, and **exit 0 without checking anything** — green CI validating an empty set. I confirmed that zero-config behavior directly: `grund check` on a tree holding a declaration and a dangling citation still prints `success` and exits 0.

### Pre-existing failure this surfaces

The bump is not free. Under 0.9.0 all four `AGENTS.md` files already fail `grund check` on `outdated grund init block v3`, independently of this move — which is why the block refresh is in the same PR rather than a follow-up. Beyond the config-path sentence, v7 adds the `BEGIN`/`END` block markers, a "Marker = live citation" rule, and a clickable-citations paragraph.

### Verification

With grund 0.9.0 locally, both gate commands pass: `grund check` and `grund fmt . --marker --cross-refs --check` exit 0. Before the change, `grund check` exited 1 with the four v3 errors.


jvukicev (approved):
LGTM

PR #1032, opened 2026-08-24
Optional LayerConfiguration

### LINKED ISSUE #1031: native-maven-plugin requires layer configuration
**Describe the bug**

With the recent versions of the native-maven-plugin the layer configuration is required even if it shouldn't for native image builds with GraalVM < 25.1

"'layer' child tag should be defined"

In the plugin.xml the tag is defined and `<required>true</required>`
Image

**To Reproduce**

Just define a native-image-plugin configuration
```xml
            <plugin>
                <version>1.1.10</version>
                <executions>
                    <execution>
                        <id>add-reachability-metadata</id>
                        <goals>
                            <goal>add-reachability-metadata</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <classesDirectory>${project.build.outputDirectory}</classesDirectory>
                    <metadataRepository>
                        <enabled>true</enabled>
                        <version>1.0.12</version>
                    </metadataRepository>
                    <requiredVersion>22.3</requiredVersion>
                </configuration>
            </plugin>
```

**Expected behavior**
The definition above should not result into a validation error.

**Logs**

Image

 - OS: Linux / Windows
 - GraalVM Version 25.0.1
 - Java Version 25
 - Plugin version 1.1.10

**Additional context**

My assumption is, that the schema is not correct in the plugin.xml. It should be `<required>false</requred>` for the layer tag.

When I ignore the error and do not insert a `<layer>` tag, the plugin is still working without errors in a maven build.

klopfdreh:
Fixes: https://github.com/graalvm/native-build-tools/issues/1031

(This is only a suggested change, please cherry-pick and push / close this PR if it is ok)

I saw that there is only an exception when the layer goal is executed, so maybe revert the changes in `executeInternal`

### DIFF
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/LayerCreateMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/LayerCreateMojo.java
@@ -70,7 +70,7 @@ import java.util.Set;
         requiresDependencyResolution = ResolutionScope.RUNTIME,
         requiresDependencyCollection = ResolutionScope.RUNTIME)
 public class LayerCreateMojo extends AbstractNativeImageMojo {
-    @Parameter(required = true)
+    @Parameter(required = false)
     private LayerConfiguration layer;
 
     @Component
--- a/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/LayerCreateMojo.java
+++ b/native-maven-plugin/src/main/java/org/graalvm/buildtools/maven/LayerCreateMojo.java
@@ -92,7 +92,12 @@ public class LayerCreateMojo extends AbstractNativeImageMojo {
 
     @Override
     protected void executeInternal() throws MojoExecutionException {
-        if (layer == null || layer.getName() == null || layer.getName().isBlank()) {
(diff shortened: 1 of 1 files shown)


jvukicev (changes requested):
Thank you for the contribution. Changing `layer` to `@Parameter(required = false)` is the right fix. IntelliJ considers parameters from all mojos when validating plugin-level configuration, so `required = true` incorrectly makes `<layer>` appear mandatory to users invoking unrelated goals.

Please remove the new early return from `executeInternal()`. An explicit `layer-create` invocation without `<layer>` should continue to fail:

```java
if (layer == null || layer.getName() == null || layer.getName().isBlank()) {
    throw new MojoExecutionException(
        "The layer-create goal requires a non-blank layer name"
    );
}
```

Otherwise, a misconfigured `layer-create` execution silently succeeds without producing an artifact. Please retain the first commit annotation change and revert the behavioral change from the second commit.

Please also sign the Oracle Contributor Agreement. The OCA check must pass before the PR can be merged.

klopfdreh:
@jormundur00 - can you apply that change? I am going to close this PR - thanks! 👍

